PR #61, opened 2015-01-13
Address alias

missinglink:
This PR starts to expose the new `address` types via the API, initially just `openaddresses` but later also `TIGER`, `openvenues` etc.

I think it's best to introduce a new `alias` for addresses, so we can turn them on/off and have query-time control over what is returned (just like the existing `poi/admin` aliases).

In this PR I've also slightly changed the way we assign `default` layers, in the past we defaulted to *all layers*, now if you fail to specify a layers param in your `GET` request it'll default to `poi,admin`.

Maybe later we can change that to include `address` but need to do some testing first.

related: https://github.com/pelias/schema/pull/36

cc/ @hkrishna

### DIFF
--- a/sanitiser/_layers.js
+++ b/sanitiser/_layers.js
@@ -11,46 +11,52 @@ function sanitize( req ){
     params = {};
   }
 
-  // which layers to query
-  if('string' === typeof params.layers && params.layers.length){
-    
-    var alias_layers  = ['poi', 'admin'];
-    var alias_indeces = indeces.concat(alias_layers);
+  // default case (no layers specified in GET params)
+  if('string' !== typeof params.layers || !params.layers.length){
+    // @note: 'address' alias disabled by default feature testing completed
+    params.layers = 'poi,admin'; // default layers
+  }
 
-    var layers = params.layers.split(',').map( function( layer ){
-      return layer.toLowerCase(); // lowercase inputs
-    });
-    for( var x=0; x<layers.length; x++ ){
-      if( -1 === alias_indeces.indexOf( layers[x] ) ){
-        return {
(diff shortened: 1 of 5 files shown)


test/unit/sanitiser/suggest.js
```diff
@@ -160,8 +160,9 @@ module.exports.tests.sanitize_layers = function(test, common) {
   });
   test('invalid layer', function(t) {
     sanitize({ layers: 'test_layer', input: 'test', lat: 0, lon: 0 }, function( err, clean ){
-      var msg = 'invalid param \'layer\': must be one or more of geoname,osmnode,osmway,admin0,admin1,admin2,neighborhood,poi,admin';
```
missinglink: I changed the way this test runs in an attempt to make it less brittle when new layers are added.

sanitiser/_layers.js
```diff
@@ -7,50 +9,56 @@ function sanitize( req ){
...
-    // de-dup
-    layers = layers.filter(function(item, pos) {
-      return layers.indexOf(item) === pos;
-    });
-
-    clean.layers = layers;
+  // default case (no layers specified in GET params)
+  if('string' !== typeof params.layers || !params.layers.length){
+    // @note: 'address' alias disabled by default feature testing completed
+    params.layers = 'poi,admin'; // default layers
```
missinglink: These are the new default layers (both aliases) which in effect expand to:

```javascript
[
  'geoname',
  'osmnode',
  'osmway',
  'admin0',
  'admin1',
  'admin2',
  'neighborhood'
]
```

In effect this is everything sans-openaddresses

PR #62, opened 2015-01-14
add osm address type

missinglink:
expose `osmaddress` type in API.

related: https://github.com/pelias/schema/pull/39
related: https://github.com/pelias/openstreetmap/issues/29

cc/ @hkrishna @sevko

### DIFF
--- a/query/indeces.js
+++ b/query/indeces.js
@@ -9,5 +9,6 @@ module.exports = [
   'admin1',
   'admin2',
   'neighborhood',
+  'osmaddress',
   'openaddresses'
 ];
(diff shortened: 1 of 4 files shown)


missinglink:
I'm going to merge this now, it will likely cause an error when requesting the `address` or `osmaddress` layer via the `layer=` GET param until this `type` mapping is inserted in your instance of ES: https://github.com/pelias/schema/pull/39

It's safe to merge because it won't try to load this layer by default (ie. when the `layer=` GET param is blank or set to any other index name or alias name).

PR #67, opened 2015-02-11
Score function

hkrishna:
This is the PR that deals with adding a scoring function using groovy scripts to sort []. Here's how the sorting works (in the given order)

1. _score:  all results are primarily sorted by _score computed by ES using its TF/IDF
2. _geodistance: if two or more have the same score and if the query provides geobias (lat/lon) - its sorted by the distance from the lat/lon provided.
3. population: if two or more records have the same score (and same no geodistance), population (if available) acts as a tie breaker
4. weights: if no lat/lon is provided and population is 0 for two or more records - then its weights based on the dataset the record originated from (https://github.com/pelias/suggester-pipeline/blob/master/lib/weights.js)

Added new tests, updated existing tests to work with the new queries. Package.json has been updated to use the latest elasticsearch-backend (for base queries) and suggester_pipeline (for weights)

### DIFF
--- a/controller/search.js
+++ b/controller/search.js
@@ -13,6 +13,7 @@ function setup( backend, query ){
     // backend command
     var cmd = {
       index: 'pelias',
+      searchType: 'dfs_query_then_fetch',
       body: query( req.clean )
     };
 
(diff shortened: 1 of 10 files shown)


missinglink:
What's the status of this? Can we discuss on our next video call?

hkrishna:
This is ready to be merged into master. it got delayed due to package dependencies versioning. Lets discuss in the next call anyway

PR #89, opened 2015-04-02
Null island bug Fix

### LINKED ISSUE #86: null island bug
There seems to be an issue getting a response for reverse@0,0

http://pelias.mapzen.com/reverse?&input=a&lat=0&lon=0

hkrishna:
Fixes #86


missinglink:
code review :+1:

PR #90, opened 2015-04-02
Allow consumers to filter suggestions by layer on '/suggest'

### LINKED ISSUE #45: /suggest 'layers' params
as per https://github.com/pelias/api/issues/22 we should allow consumers to filter suggestions by `layer` on `'/suggest'` as they do on `'/search'`.

We have the 'dataset' category context suggester already set up to allow this.

hkrishna:
Fixes #45

### DIFF
--- a/query/suggest.js
+++ b/query/suggest.js
@@ -34,7 +34,7 @@ function generate( params, query_mixer, fuzziness ){
         'size' : this.params.size,
         'field' : 'suggest',
         'context': {
-          'dataset': layers || this.params.layers,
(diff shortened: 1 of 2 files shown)


missinglink:
The logic for the query mixer is complex and could do with some documentation in order for other developers to follow your logic, I am assuming you tested this locally so :+1: to merge.

hkrishna:
yeah, maybe its time we ripped out all the query mixing logic out of the API and into its own module? But I'll start with writing up some [documentation](https://github.com/pelias/api/issues/99) on how it works.

PR #93, opened 2015-04-06
Output generator tweaks

### LINKED ISSUE #46: outputGenerator tweaks
In the following example, the 'text' should be **'Arbil, Iraq'** but due to the way the `outputGenerator` is written it matches nothing for `local` and 'Arbil' for `regional` (the `name` is also 'Arbil').

If I simply add `admin1` to the end of the [local array](https://github.com/pelias/api/blob/suggest_plus_mget/helper/outputSchema.json#L11) then both `local` and `regional` will match 'Arbil' and when the de-duper removes one entry, we are back with the text being 'Arbil'.

Not sure if I described this well, but basically we need to continue checking values in the `local` and `regional` arrays until we find a value which is not a duplicate of the name or another admin value, otherwise a POI with ```{ "name": "a", "local_admin": "a", "admin1_abbr": "a", "admin0": "b" }``` will always just be `a` when it would be nicer if it were `a, b`.

Hope that makes sense ;)

```javascript
{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "geometry": {
        "type": "Point",
        "coordinates": [
          44.180874,
          36.369878
        ]
      },
      "properties": {
        "id": "1110:adm1:iq:irq:arbil",
        "type": "admin1",
        "layer": "admin1",
        "name": "Arbil",
        "alpha3": "IRQ",
        "admin0": "Iraq",
        "admin1": "Arbil",
        "text": "Arbil"
      }
    },
...
```

hkrishna:
Does adding a 'national' key to the outputSchema make sense? Here's the argument - When Pelias used as a product with a dataset thats global (meaning more than one or two countries), doesn't it not make sense to have a national segment to the output schema. This of-course can be optional if not mentioned in ```outputSchema.json``` 

Note: This will cause a bunch of regressions in https://github.com/pelias/acceptance-tests
Fixes #46

### DIFF
--- a/helper/outputSchema.json
+++ b/helper/outputSchema.json
@@ -7,11 +7,11 @@
   "GBR": {
     "local": ["neighborhood", "local_admin", "locality"],
     "regional": ["admin2", "admin1"],
-    "national": ["admin0", "alpha3"]
+    "national": ["alpha3", "admin0"]
   },
   "default": {
     "local": ["local_admin", "locality", "neighborhood", "admin2"],
     "regional": ["admin1_abbr", "admin1"],
-    "national": ["admin0", "alpha3"]
+    "national": ["alpha3", "admin0"]
   }
 }
--- a/helper/outputSchema.json
+++ b/helper/outputSchema.json
@@ -1,14 +1,17 @@
 {
   "USA": {
     "local": ["local_admin", "locality", "neighborhood", "admin2"],
-    "regional": ["admin1_abbr", "admin1", "admin0"]
+    "regional": ["admin1_abbr", "admin1"],
+    "national": ["alpha3", "admin0"]
   },
   "GBR": {
-    "local": ["neighborhood", "admin2", "local_admin", "locality"],
-    "regional": ["admin2","admin0","admin1"]
+    "local": ["neighborhood", "local_admin", "locality"],
+    "regional": ["admin2", "admin1"],
+    "national": ["admin0", "alpha3"]
   },
   "default": {
     "local": ["local_admin", "locality", "neighborhood", "admin2"],
-    "regional": ["admin1_abbr", "admin1", "admin0"]
+    "regional": ["admin1_abbr", "admin1"],
+    "national": ["admin0", "alpha3"]
   }
 }
--- a/test/unit/controller/doc.js
+++ b/test/unit/controller/doc.js
@@ -29,7 +29,7 @@ module.exports.tests.functional_success = function(test, common) {
       admin0: 'country1',
       admin1: 'state1',
       admin2: 'city1',
-      text: 'test name1, city1, state1'
+      text: 'test name1, city1, state1, country1'
     }
   }, {
     type: 'Feature',
@@ -44,7 +44,7 @@ module.exports.tests.functional_success = function(test, common) {
       admin0: 'country2',
       admin1: 'state2',
       admin2: 'city2',
-      text: 'test name2, city2, state2'
+      text: 'test name2, city2, state2, country2'
     }
   }];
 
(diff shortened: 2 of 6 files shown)


missinglink:
You're proposing adding a 4th 'section' to the output text?

eg. the `text` property in the geojson would be in the format `name, local, regional, national` instead of just `name, local, regional`?

one thing to consider is that for mobile clients 'less is more', we introduced the `admin1_abbr` property for this reason; because small screens can't accommodate [such long/descriptive text](https://www.youtube.com/watch?v=1g0Hsp4rVzU).

dianashk:
+1

It makes sense to add it. (nominatim does)

hkrishna:
@missinglink yes, I'm proposing adding a 4th section ```national``` - I thought about mobile clients and if you mention ```alpha3``` before ```admin0``` in outputSchma.json, ```alpha3``` takes precedence. 

```
"default": {
    "local": ["local_admin", "locality", "neighborhood", "admin2"],
    "regional": ["admin1_abbr", "admin1"],
    "national": ["alpha3", "admin0"]
  }
```

This would yield results such as ```Blind Barber, East Village, NY, USA```  or ```123 E 12th st, New York, NY, USA``` :smile_cat:

hkrishna:
@missinglink is this good to go? You can easily turn it off by simply not mentioning the ```national``` key in the outputSchema.json file.

hkrishna:
Nevermind, I found some inconsistencies with ```alpha3``` values missing in a few datasets. Working on it.

hkrishna:
Ugh - I broke tests. WIP

missinglink:
I am still :-1: on this sorry, I personally feel that the string returned is not as good as the previous iteration, based on what I see here: http://rawgit.com/pelias/demo/two-servers/index.html#loc=12,40.7255,-73.9795&q=arbil&t=fine&gb=off

- it's less succinct & therefore more difficult to render on smaller screens, if you consider the impact to the 'mapzen open app', this will probably result in more rows going to a 2nd line of text which means less results per 'page'.
- suffixing the output text with an alpha3 is pretty geeky and not very human, it doesn't pass the 'mum test' ie. does it make sense to a non/semi technical user?
- I personally like the gmaps experience, eg. 'Trafalgar Square, London' gives me enough information to understand where it is by simply providing `name, city`.

Is there a simpler, more human solution we could explore which passes the 'mum test'?

hkrishna:
Thats fair. Yes, we can explore a simpler, more human solution that passes the mum test. I agree with your comment on whether alpha3 values makes any sense to non/semi technical users - it probably doesn't. And if we use ```admin0``` values - it might get too wordy and long - probably wont make sense for someone who always searches within his/her country. 

While we have to think about mobile/ smaller screens and account for possible rendering issues, we should also keep in mind that the clients can compile their own presentational logic based on the document properties. 

Another approach to this problem is to let the user decide what version of output text they would like through the API. Perhaps an additional optional parameter lets say ```outputSchema```. 

* ```&outputSchema=all``` (by default): will return name, local, regional, national as defined in outputSchema.json Ex: ```Arbil, Lye, West Midlands, GBR```
* ```&outputSchema=local,regional```: will return name, local, regional Ex: ```Arbil, Lye, West Midlands```

We can also push this one step further and define formats for mobile ```&outputSchema=mobile``` in a config file which will cater to smaller screens. Thoughts?

sevko:
I like the idea of an `outputSchema` parameter, and especially with some out-of-the-box aliases (like `mobile` and `all`).

sevko:
Should we close this PR in favor of another one that implements `outputSchema`?

hkrishna:
Closing this issue in favor of #100

PR #102, opened 2015-04-14
adding all admin layers to the search query

### LINKED ISSUE #91: boost results by query locality
A [query for `1710 Drew, Houston`](http://pelias.stage.mapzen.com/search?input=1710%20drew,%20Houston) currently ranks `1710 Drew Place, Claremont, CA` higher than `1710 Drew Street, Houston, TX`. According to @hkrishna , we only boost matches for `admin0`, `admin1`,  and `admin1_abbr`, meaning that the exact `locality` match doesn't get taken into account at all here. We should factor it into boosting if it doesn't significantly impact query times.

hkrishna:
Search now looks for all admin layers and boosts ```_score``` accordingly

```
[ 'admin1_abbr',
  'alpha3',
  'admin0',
  'admin1',
  'admin2',
  'neighborhood',
  'locality',
  'local_admin' ]
```

Fixes #91

### DIFF
--- a/config/queryAdminWeights.js
+++ b/config/queryAdminWeights.js
@@ -1,0 +1,15 @@
+/**
+ * Configurable weights that influence how, for a query with an admin value
+ * present, a match on any given admin value affects a document's overall
+ * score.
+ */
+
+module.exports = {
+  admin0: 20,
+  alpha3: 20,
+  admin1: 10,
+  admin1_abbr: 10,
+  admin2: 5,
+  locality: 1,
+  local_admin: 1
+};
--- a/query/search.js
+++ b/query/search.js
@@ -2,7 +2,8 @@
 var logger = require('../src/logger'),
     queries = require('geopipes-elasticsearch-backend').queries,
     get_layers = require('../helper/layers'),
-    sort = require('../query/sort');
+    sort = require('../query/sort'),
+    adminFieldWeights = require( '../config/queryAdminWeights' );
 
 function generate( params ){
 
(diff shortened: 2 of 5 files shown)


sevko:
So, one problem that I haven't considered is that results like "New York County", which are admin2 records and thus don't have any admin names below that, can't possibly match on any of them and thus will usually be ranked lower than, say, address-level records, which *do* have them:

```javascript
{
  name: "New York County",
  admin0: "United States",
  admin1: "New York",
  admin2: "New York County"
}
```

would score lower than:

```javascript
{
  name: "New York City Hall",
  admin0: "United States",
  admin1: "New York",
  admin2: "New York County"
  local_admin: "Manhattan",
  locality: "New York"
}
```

Assuming that both records get the same `_score` from their `name.default` properties alone, since they both contain "New York", *without* any admin-boosting on `local_admin` and `locality` the `New York County` record would end up being first because it has a high population value. *With* it, on the other hand, the `_score` for the second record would become higher, so population wouldn't even get taken into account in the corresponding Groovy script (since there wouldn't be a tie on `_score`).

sevko:
Since a multi-condition `should` clause will boost the document score for each match, which consequently causes the problem outlined in the above comment, it might be better to match on *only* 1 field, starting with `admin0` and moving down to `admin1`, `admin2`, etc., assigning a lower value to each subsequent layer, and **stopping** after the first match.

sevko:
After a lot of trial and error, we've decided to defer this work until we have smarter admin value parsing in place.

PR #103, opened 2015-04-14
Detail Param

### LINKED ISSUE #101: Address details Parameter
If a user wants to just use the text property in the output and doesnt care about the rest - then he/she should be able to request a smaller payload by setting a parameter lets say ```addressDetails``` to ```false```

Currently, each point that gets returned by the API looks like the following

```js
{  
   type:"Feature",
   properties:{  
      id:"127099401",
      layer:"osmway",
      name:"Mays",
      alpha3:"GBR",
      admin0:"United Kingdom",
      admin1:"Bournemouth",
      admin2:"Dorset",
      locality:"Bournemouth",
      neighborhood:"Westbourne",
      text:"Mays, Westbourne, Dorset"
   },
   geometry:{  
      type:"Point",
      coordinates:[  
         -1.902019,
         50.721979
      ]
   }
}
```

with  ```&addressDetails=false``` parameter, the output is slimmed down to the following

```js
{  
   type:"Feature",
   properties:{  
      text:"Mays, Westbourne, Dorset"
   },
   geometry:{  
      type:"Point",
      coordinates:[  
         -1.902019,
         50.721979
      ]
   }
}
```

hkrishna:
adding a parameter called details (defaults to false for all endpoint except reverse) + tests

fixes #101 

### Search

http://localhost:3100/search?input=a&details=false

```js
{
  type: "FeatureCollection",
  features: [
  {
    type: "Feature",
    properties: {
      id: "4145526",
      layer: "geoname",
      text: "A A Hilton Pond, Okaloosa County, Florida"
    },
   ...
  },
  ...
}
```

http://localhost:3100/search?input=a

```js
{
  type: "FeatureCollection",
  features: [
  {
    type: "Feature",
    properties: {
      id: "4145526",
      layer: "geoname",
      name: "A A Hilton Pond",
      alpha3: "USA",
      admin0: "United States",
      admin1: "Florida",
      admin2: "Okaloosa County",
      text: "A A Hilton Pond, Okaloosa County, Florida"
    },
   ...
  },
  ...
}
```

### /suggest, /doc, /reverse is similar to /search

If you pass ```&details=false``` or ```details=0``` you will not see the additional/ detailed payload - in all other cases, you will. 

***By Default details is set to ```true```***


missinglink:
Sorry I didn't see this until after I commented on https://github.com/pelias/api/issues/101

hkrishna:
This is ready to go! @missinglink

sanitiser/_details.js
```diff
@@ -0,0 +1,32 @@
...
+  var params= req.query;
+
+  if (default_value === undefined) {
+    default_value = true;
+  }
+  
+  default_value = !!default_value;
+
+  // ensure the input params are a valid object
+  if( Object.prototype.toString.call( params ) !== '[object Object]' ){
```
missinglink: I started using `npm is-object` for this `var isObject = require('is-object');` https://github.com/pelias/api/blob/master/package.json#L41; because javascript

sanitiser/_details.js
```diff
@@ -0,0 +1,32 @@
...
+    params = {};
+  }
+
+  if (params.details !== undefined) {
```
missinglink: this block of code is super-hard to read; because ternary operators

controller/doc.js
```diff
@@ -23,7 +23,7 @@ function setup( backend ){
       if( err ){ return next( err ); }
 
       // convert docs to geojson
-      var geojson = geojsonify( docs );
+      var geojson = geojsonify( docs, req.clean.details );
```
missinglink: minor comment, if the second arg was an object then it could be more easily extended to allow a range of flags to be passed. as it is, if you want to add another flag-type-thing you'll either need to refactor this or add another new arg.

missinglink:
added some comments about minor nit-picks, other than that :+1: looks good. does this flag also work with `/reverse`?

sanitiser/_details.js
hkrishna: hahah I love your ```because javascript``` comment :) Yes, will replace with ```is-object```

sanitiser/_details.js
hkrishna: gotcha. I'll simplify it

controller/doc.js
hkrishna: should we just pass ```req``` object as a whole?

hkrishna:
yes, it works for all endpoints :smile_cat: 

I'll work on your minor (but completely valid) comments and let you know

hkrishna:
@missinglink check it out - I think its ready now

sevko:
+1

missinglink:
:+1: ping! @hkrishna this is good to merge & release

dianashk:
* Tested in dev.
* Acceptance tests added.
* Good to merge.

Issue #105, opened 2015-04-16
Use categories as a scoring function


hkrishna:
Each point has a field called ```category``` that contains some valuable information, it would be nice to use this information in the scoring algorithm.

We could have a file ```weighted_categories.yml``` where we can define what weight each tag value corresponds to.

```yml
   airport: 12
   railway: 10
   restaurant: 8
   tourism: 11
   poi:landmark: 10
   ...
```
This should be a query time boosting easily do-able by just adding a [script](http://github.com/pelias/scripts) and using it in the API sorting logic. 

Related to pelias/openstreetmap#33 and pelias/pelias#50

PR #106, opened 2015-04-16
category scoring [WIP]

hkrishna:
Weights are all hardcoded to 10 currently - requires to be tweaked

```js
'transport': 10,
'transport:air': 10,
'transport:air:aerodrome': 10,
'transport:air:airport': 10,
'recreation': 10,
'religion': 10,
'education': 10,
'entertainment': 10,
'nightlife': 10,
'food': 10,
'government': 10,
'professional': 10,
'finance': 10,
'health': 10,
'retail': 10,
'transport:public': 10,
'transport:bus': 10,
'transport:taxi': 10,
'transport:sea': 10,
'accomodation': 10,
'transport:station': 10,
'food:bagel': 10,
'food:barbecue': 10
...
```

uses https://github.com/pelias/scripts/pull/6
fixes pelias/api#105


sevko:
@hkrishna , @dianashk , @missinglink , did this merit more discussion before getting merged? I can revert so that we get a chance to collectively look it over.

missinglink:
[question] @sevko @hkrishna will this add a 3kb payload to each db query containing `./helper/category_weights.js`? That sounds not ideal...

sevko:
Currently, yes, and no, it isn't. Could either move those into the Groovy scripts, but that stops us from injecting and thus easily configuring them. Since they're virtually all currently set to `10`, which serves as a default value, we can just remove anything that doesn't have a non-default weight (ie, in the Groovy script, assume a weight of 1 unless a custom weight for the given category was present in the injected weights object).

missinglink:
OK that sounds better, thanks.

PR #107, opened 2015-04-17
Layers be layers not layer (typo in the error message)

hkrishna:
param ```layers``` should be called ```layers``` consistently across the API


missinglink:
:+1: for consistency

PR #108, opened 2015-04-17
Better API index page

hkrishna:
* adding a new file DOCS.md (same as https://github.com/pelias/api/wiki/API-Endpoints)
* modifying controller/index.js to return text/html instead of application/json by default. * Using markdown module to convert MD to HTML. 
* Fixing tests (removed test/jsonp.coffee)


controller/index.js
```diff
@@ -1,16 +1,30 @@
...
 function setup(){
 
   function controller( req, res, next ){
 
-    // stats
-    res.json({
-      name: pkg.name,
-      version: {
-        number: pkg.version
+    fs.readFile('./DOCS.md', 'utf8', function (err, content) {
```
missinglink: This is going to hit disk every time, it might be better to load the file in RAM once when the server starts up since it's a small file that might be served often.

controller/index.js
```diff
@@ -1,16 +1,30 @@
...
+      if (!err) {
+        var header = '# Pelias API\n';
+        var version = '### Version: ['+ pkg.version+ '](https://github.com/pelias/api/releases)\n';
+        var style = '<style>html{font-family:monospace}</style>';
+        
+        res.send(style + markdown.toHTML(header + version + content));
+
+      } else {
```
missinglink: I'm a little uneasy about serving `text/html` sometimes and `application/json` at other times. It might be nicer to check the `Accepts` header from the User Agent and serve content accordingly. Browsers will always send a header such as `Accept:text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8` when they want HTML and client libraries will always send 'Accept:application/json' so no-one gets nasty surprises.

controller/index.js
hkrishna: agreed. WIP

controller/index.js
hkrishna: Fair point. WIP

controller/index.js
```diff
@@ -1,22 +1,51 @@
...
-      version: {
-        number: pkg.version
+  var text = null;
+  function getText( callback ) {
+    if( text ) {
+      process.nextTick( callback.bind( null, null, text ) );
+      return;
+    }
+
+    fs.readFile( './DOCS.md', 'utf8', function ( err, content ) {
```
sevko: Can we use `fs.readFileSync()` instead? It'll only happen once so blocking the process is a non-issue, and would do away with the callback.

sevko: In fact, could just do it in the global scope (line 8):

```
var text = fs.readFileSync('./DOCS.md');
```

and remove `getText()` entirely.

dianashk: good point. I'll make the change.

controller/index.js
```diff
@@ -1,22 +1,32 @@
...
 
-  function controller( req, res, next ){
+  var text = '# Pelias API\n';
+  text += '### Version: ['+ pkg.version+ '](https://github.com/pelias/api/releases)\n';
+  text += fs.readFileSync( './DOCS.md', 'utf8');
 
-    // stats
+  function controller( req, res, next ) {
+    if (req.accepts('html')) {
+      if( text ) {
```
sevko: Sorry, nitpick: this conditional looks unnecessary.

sevko:
+1

controller/index.js
```diff
@@ -1,22 +1,32 @@
...
+        var style = '<style>html{font-family:monospace}</style>';
+        res.send(style + markdown.toHTML(text));
```
sevko: Can we call `markdown.toHTML()` on line 10, rather than once per request?

sevko:
+1

dianashk:
because they were failing when trying to commit (unless API was running) and @missinglink said he feels they aren't necessary now that we have good unit test coverage.

PR #109, opened 2015-04-17
[Major bug fix] using coarse sanitizer for the coarse endpoint

hkrishna:
this fixes the ```/suggest/coarse``` endpoint - ensures it only uses ```admin``` layers


missinglink:
:+1: nice catch

PR #113, opened 2015-04-23
Boost exact matches#5

sevko:
See the commit logs for detailed messages, but tl;dr: modify the API to use the script introduced in pelias/scripts#7.

Fixes pelias/scripts#5.


query/sort.js
```diff
@@ -3,36 +3,53 @@ var population = 'population';
...
+          'weights': weights
+        },
+        'file': 'weights',
+        'type': 'number',
+        'order': 'desc'
+      }
     }
+  ];
+
+  if( typeof params === 'object' && params.hasOwnProperty( 'input' ) ){
```
hkrishna: use ```is-object```

query/reverse.js
```diff
@@ -11,9 +11,9 @@ function generate( params ){
   };
 
   var query  =  queries.distance( centroid, { size: params.size || 1 } );
-  query.sort = query.sort.concat(sort);
+  query.sort = query.sort.concat( sort() );
```
hkrishna: why not pass ```params``` here? I know reverse doesnt have ```input``` field requirement. However, for consistency sake - I'd argue we pass ```params``` because maybe a script in the future might use ```params.lat``` ```params.lon``` for example..

test/unit/query/sort.js
```diff
@@ -50,12 +51,30 @@ var expected = [
...
 module.exports.tests.query = function(test, common) {
   test('valid part of query', function(t) {
-    var sort = generate;
+    var sort = generate();
     t.deepEqual(sort, expected, 'valid sort part of the query');
     t.end();
   });
 };
 
+module.exports.tests.queryWithInput = function ( test, common ){
```
hkrishna: * test for when params object has no ```input``` or when ```input = ''```
* test for when params object is not passed in

sevko: `input = ''` borks, since that should be handled by the sanitizer. The latter is already handled by the `.query` test. Testing for an object with no `input` is doable, though.

hkrishna:
looks good :+1:

sevko:
The API's currently deployed off this branch in dev and all acceptance-tests pass. The following (condensed) elasticsearch query:

```json
{
		"query": {
				"filtered": {
						"query": {
								"bool": {
										"must": [
												{
														"match": {
																"name.default": "318 19th street"
														}
												}
										]
								}
						},
						"filter": { "bool": { "must": [] } }
				}
		},
		"size": 10,
		"sort": [
				"_score",
				{ "_script": { "file": "admin_boost", "type": "number", "order": "desc" } },
				{ "_script": { "file": "population", "type": "number", "order": "desc" } },
				{ "_script": { "file": "popularity", "type": "number", "order": "desc" } },
				{ "_script": { "params": { "weights": { "geoname": 0, "address": 4, "osmnode": 6, "osmway": 6, "poi-address": 8, "neighborhood": 10, "local_admin": 12, "locality": 12, "admin2": 12, "admin1": 14, "admin0": 2 } }, "file": "weights", "type": "number", "order": "desc" } },
				{ "_script": { "params": { "input": "318 19th street" }, "file": "exact_match", "type": "number", "order": "desc" } }
		],
		"track_scores": true
}
```

when executed against our dev elasticsearch instance returns:

```javascript
{
    "_shards": {
        "failed": 0,
        "successful": 2,
        "total": 2
    },
    "hits": {
        "hits": [
            {
                "_id": "address-osmway-248170365",
                "_index": "pelias",
                "_score": 6.310146,
                "_source": {
                    "address": {
                        "number": "318",
                        "street": "19th Street",
                        "zip": "11215"
                    },
                    "admin0": "United States",
                    "admin1": "New York",
                    "admin1_abbr": "NY",
                    "admin2": "Kings County",
                    "alpha3": "USA",
                    "category": [],
                    "center_point": {
                        "lat": 40.660571,
                        "lon": -73.989475
                    },
                    "local_admin": "Brooklyn",
                    "locality": "New York",
                    "name": {
                        "default": "318 19th Street"
                    },
                    "neighborhood": "Windsor Teraace",
                    "suggest": {
                        "input": [
                            "318 19th street"
                        ],
                        "output": "osmaddress:address-osmway-248170365"
                    }
                },
                "_type": "osmaddress",
                "sort": [
                    6.310146,
                    6.0,
                    0.0,
                    0.0,
                    0.0,
                    1.0
                ]
            },
            {
                "_id": "address-osmway-264629154",
                "_index": "pelias",
                "_score": 6.310146,
                "_source": {
                    "address": {
                        "number": "318",
                        "street": "West 19th Street",
                        "zip": "10011"
                    },
                    "admin0": "United States",
                    "admin1": "New York",
                    "admin1_abbr": "NY",
                    "admin2": "New York County",
                    "alpha3": "USA",
                    "category": [],
                    "center_point": {
                        "lat": 40.743087,
                        "lon": -74.001273
                    },
                    "local_admin": "Manhattan",
                    "locality": "New York",
                    "name": {
                        "default": "318 West 19th Street"
                    },
                    "neighborhood": "Chelsea",
                    "suggest": {
                        "input": [
                            "318 west 19th street"
                        ],
                        "output": "osmaddress:address-osmway-264629154"
                    }
                },
                "_type": "osmaddress",
                "sort": [
                    6.310146,
                    6.0,
                    0.0,
                    0.0,
                    0.0,
                    0.0
                ]
            },

```

The `_score`s are both identical, but `318 19th Street` gets an additional 1.0 in its `sort` values from the `exact_match.groovy` script that's guaranteed to boost it to the top.

sevko:
The performance hit is surprisingly non-negligible. An average time of 0.03299 and 0.05139 seconds before/after for a query for "318 19th street", and 0.0107 and 0.0309 seconds for "new york, new york" (this is the average time for each query repeated 100 times).

sevko:
Right, so after some further experimentation it turns out that the performance hit is caused entirely by usage of the `_source` object in [exact_match.groovy](https://github.com/pelias/scripts/blob/boost-exact-matches%235/scripts/exact_match.groovy), which is expected as per the [elasticsearch documentation](http://www.elastic.co/guide/en/elasticsearch/reference/1.5/modules-scripting.html#_source_field):

> Accessing _source is much slower compared to using _doc but the data is not loaded into memory.

The reason that script uses `_source` is that it appears to be the only way of getting the untokenized `name.default` of a given document, since `doc["name.default"]` returns the analyzed token array. If there's any way of retrieving the untokenized name *without* `_source`, this'll become a non-issue. If not, we can try storing a raw, un-analyzed version of `name.default` to use specifically in scripts like these, but we'll have to discuss whether there any other use-cases for it since it seems like a fairly significant change for a small feature.

sevko:
I've [amended the scripts](https://github.com/pelias/scripts/commit/17c2bababbcce478da23bf528d244c8c3fd30516) to avoid using `_source`, with identical improvements in search results. The respective [new test cases](https://github.com/pelias/acceptance-tests/pull/42) pass in staging when the `boost-exact-match` branches are checked out in both pelias/api and pelias/scripts, and fail otherwise, so there's demonstrable improvement.

query/sort.js
```diff
@@ -4,47 +4,65 @@ var popularity = 'popularity';
...
+        'file': 'weights',
+        'type': 'number',
+        'order': 'desc'
+      }
     }
+  ];
+
+  if( isObject( params ) && params.hasOwnProperty( 'input' ) ){
+    scriptsConfig.push({
+      _script: {
```
dianashk: should be consistent with above code in either using or getting rid of the `' '` around property names.

test/unit/query/search.js
```diff
@@ -58,6 +58,16 @@ var sort = [
       'type': 'number',
       'order': 'desc'
     }
+  },
+  {
+    _script: {
```
dianashk: same inconsistency here.

query/sort.js
sevko: Yeah, they shouldn't have been there in the first place. I'd like to avoid doing that in this PR since it'll clutter the diff and is an unrelated cleanup change, though; will gladly do it in another branch/directly in `master`.

sevko: I'll make these properties consistent with the others by quoting them, and reserve removing them all for another PR.

dianashk:
:+1:

hkrishna:
:+1:

PR #115, opened 2015-04-24
Reverse categories

dianashk:
Add `categories` parameter to `/reverse` endpoint. Expected values are comma-separated strings. The categories specified should come from the supported category taxonomy to surface results.

The query will `or` all specified categories and results will be sorted by proximity to `lat,lon` as done in all other `/reverse` query cases.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 node_modules
+coverage
+.idea
 *.log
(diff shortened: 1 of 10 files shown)


missinglink:
:+1: great stuff, super-cool feature, really excited to play with this in production.

sevko:
Will we have separate PRs for `/suggest` and `/search`?

dianashk:
Yes. This is currently just in /reverse to see how we like it. Once we are happy with it, we can move on to the other endpoints.

sevko:
+1

PR #116, opened 2015-04-27
Resolve #114: report runtime errors.

### LINKED ISSUE #114: API (possibly express) swallows runtime exceptions
The API appears to silently swallow runtime exceptions that would otherwise cause it to print a stack trace and crash. For instance, I've changed [query/search.js:17] (https://github.com/pelias/api/blob/master/query/search.js#L17) from:

```javascript
var query = queries.distance( centroid, { size: params.size } );
```

to

```javascript
var query = queries.distance( centroid, { size: param.size } );
```

Note that `params.size` became `param.size`, and, because `param` doesn't exist, you'd expect the process to crash with `ReferenceError: param is not defined`. When running the API via `node index.js`, however:

```
$ curl localhost:3100/search?input=foobar
{"error":{}}
```

The request obviously errored and the status code is 500, yet the API keeps running without any output besides `listening on 3100`. If you try to use the `query/search` module *directly*, on the other hand, it errors as expected:

```javascript
$ node -e 'console.log(require("./query/search")({}));'
/home/sevko/src/mapzen/api/query/search.js:17
  var query = queries.distance( centroid, { size: param.size } );
                                                  ^
ReferenceError: param is not defined
    at generate (/home/sevko/src/mapzen/api/query/search.js:17:51)
    at [eval]:1:38
    at Object.exports.runInThisContext (vm.js:74:17)
    at Object.<anonymous> ([eval]-wrapper:6:22)
    at Module._compile (module.js:460:26)
    at evalScript (node.js:431:25)
    at startup (node.js:90:7)
    at node.js:814:3
```

sevko:
```
middleware/500.js
	-Add code to the 500 catchall middleware to report errors to
	stdout/stderr.
	-Also, stringify the error via `.toString()` before passing it
	to `res.json()`, since otherwise an empty object appears to be
	returned.
```

### DIFF
--- a/middleware/500.js
+++ b/middleware/500.js
@@ -6,7 +6,7 @@ function middleware(err, req, res, next) {
   logger.error( 'Stack trace:', err.trace );
   res.header('Cache-Control','no-cache');
   if( res.statusCode < 400 ){ res.status(500); }
-  res.json({ error: err.toString() });
+  res.json({ error: 'internal server error' });
 }
 
 module.exports = middleware;
--- a/middleware/500.js
+++ b/middleware/500.js
@@ -1,9 +1,12 @@
+var logger = require( '../src/logger' );
 
 // handle application errors
 function middleware(err, req, res, next) {
+  logger.error( 'Error:', err );
+  logger.error( 'Stack trace:', err.trace );
   res.header('Cache-Control','no-cache');
   if( res.statusCode < 400 ){ res.status(500); }
-  res.json({ error: err });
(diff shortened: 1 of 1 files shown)


missinglink:
we shouldn't expose errors to the API consumer, they may contain sensitive information such as paths and usernames.

sevko:
oof, amateur mistake. Good call, amended.

hkrishna:
:+1:

PR #117, opened 2015-04-27
Better way of describing bbox param

hkrishna:
```javascript
"?bbox=southwest_lng,southwest_lat,northeast_lng,northeast_lat"
```

is a standard way of passing around bbox description over a web req. It can be easily produced using the following leaflet magic (if your map is powered by leaflet for example)

```javascript
L.latLngBounds(southwestLatLng, northeastLatLng).toBBoxString()
```

Heres an example:

[http://pelias.mapzen.com/search?input=west&bbox=41.430,-88.30,42.198,-87.47](http://pelias.mapzen.com/search?input=west&bbox=41.430,-88.30,42.198,-87.47)

Its important we document this well.

### DIFF
--- a/sanitiser/_geo.js
+++ b/sanitiser/_geo.js
@@ -35,8 +35,8 @@ module.exports = function sanitize( req, latlon_is_required ){
 /**
  * Parse and validate bbox parameter
  * bbox = bottom_left lat, bottom_left lon, top_right lat, top_right lon
- * bbox = left,bottom,right,top
- * bbox = min Longitude , min Latitude , max Longitude , max Latitude
+ * bbox = bottom, left, top, right
+ * bbox = min Latitude , min Longitude , max Latitude , max Longitude
  *
  * @param {object} clean
  * @param {string} param
--- a/sanitiser/_geo.js
+++ b/sanitiser/_geo.js
@@ -1,10 +1,11 @@
 var isObject = require('is-object');
 
+
 // validate inputs, convert types and apply defaults
-function sanitize( req, latlon_is_required ){
-  
+module.exports = function sanitize( req, latlon_is_required ){
+
   var clean = req.clean || {};
-  var params= req.query;
+  var params = req.query;
   latlon_is_required = latlon_is_required || false;
 
   // ensure the input params are a valid object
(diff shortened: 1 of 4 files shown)


hkrishna:
While this started as an enhancement to the documentation - I just discovered a bug wherein if you pass in the string generated by ```.toBBoxString()``` Pelias API doesn't return anything because it expects southwest **lat**, southwest **lon**, northeast **lat**, northeast **lon** and not "southwest_lng,southwest_lat,northeast_lng,northeast_lat" 

lat and lng are reversed. It would be nice if Pelias API supported the leaflet convention of ```lon,lat``` and not ```lat,lon```. 

Does this make sense?

missinglink:
Fun fact of the day: Twitter are still living with a mistake they made around the ordering of [ lat, lon ] in tweets returned via the API.

As a result you still see this when using their API today:
```javascript
{
  "tweet": {
    "geo":{"type":"Point","coordinates":[44.2927839,-88.41006552]},
    "coordinates":{"type":"Point","coordinates":[-88.41006552,44.2927839]}
  }
}
```

https://twittercommunity.com/t/why-is-lat-long-pair-reversed-in-geo-tag-vs-coordinates-tags/8486

#themoreyouknow #fail #becausegeo

hkrishna:
Should we re-order them to [lon, lat] or keep it as is [lat, lon] or go the twitter route? :smile_cat: 

Thoughts?

sevko:
I'd opt for making it compatible with Leaflet's `.toBBoxString()`.

sanitiser/_geo.js
```diff
@@ -12,87 +13,102 @@ function sanitize( req, latlon_is_required ){
...
-        'message': 'invalid param \'lon\': must be >-180 and <180'
+  req.clean = clean;
+
+  return { 'error': false };
+};
+
+
+/**
+ * Parse and validate bbox parameter
+ * bbox = bottom_left lat, bottom_left lon, top_right lat, top_right lon
```
hkrishna: These comments might need to be updated?

```javascript
/*
 * bbox = bottom_left lon, bottom_left lat, top_right lon, top_right lat
 * bbox = bottom,left,top,right
 */ 
```

hkrishna:
Other than the minor comment change, it looks good to me :+1:

dianashk:
verified on dev

Issue #56, opened 2014-12-15
interpret lon:181 as lon:-179


hkrishna:
https://github.com/pelias/demo/issues/10

orangejulius:
The tests very clearly show lon > 180 should be invalid, is this change still being considered?

hkrishna:
yes. I think this needs to happen at the [sanitizer](https://github.com/pelias/api/blob/master/sanitiser/_geo.js) level and tests be updated accordingly.

missinglink:
![](http://i279.photobucket.com/albums/kk143/faeini1/agscary4.gif)

PR #121, opened 2015-05-04
wrap longitude (#56)

sevko:
Implement longitude wrapping by removing the longitude sanitizer (elasticsearch appears to gracefully handle lat/lon wrapping, such that it treats a longitude of 181 as -179). Note that while I ripped out some of the bits in `sanitizer/_geo`, I simultaneously refactored some convoluted code, so there are some slightly unrelated changes.

fixes pelias/api#56

### DIFF
--- a/sanitiser/_geo.js
+++ b/sanitiser/_geo.js
@@ -81,7 +81,7 @@ function sanitize_bbox( clean, param ) {
  * @param {bool} latlon_is_required
  */
 function sanitize_coord( coord, clean, param, latlon_is_required ) {
-  var value = parseFloat( param, 10 );
+  var value = parseFloat( param );
   if ( !isNaN( value ) ) {
     if( coord === 'lat' && check_lat.is_invalid( value ) ){
       throw new Error( 'invalid ' + check_lat.error_msg );
--- a/sanitiser/_geo.js
+++ b/sanitiser/_geo.js
@@ -51,14 +51,17 @@ function sanitize_bbox( clean, param ) {
 
   if( Array.isArray( bboxArr ) && bboxArr.length === 4 ) {
 
-    var bbox = bboxArr.map( function ( latlon ){
-      return parseFloat( latlon, 10 );
-    });
-    bboxArr.forEach( function( latlon, index ) {
-      if( index % 2 === 1 && check_lat.is_invalid( latlon ) ){
-        throw new Error( 'Invalid lat: ' + latlon );
+    var bbox = [];
+    for( var ind = 0; ind < bboxArr.length; ind++ ){
+      var num = parseFloat( bboxArr[ ind ] );
+      if( isNaN( num ) ){
+        return;
       }
-    });
+      if( ind % 2 === 1 && check_lat.is_invalid( num ) ){
+        throw new Error( 'Invalid lat: ' + num );
+      }
+      bbox.push( num );
+    }
 
     clean.bbox = {
       right: Math.max( bbox[0], bbox[2] ),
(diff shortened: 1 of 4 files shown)


dianashk:
If ES handles both lat and lon, why did you only remove the lon sanitizer? That seems odd to me. We should either remove both or sanitize both.

Also please add acceptance tests for this. Thanks.

sevko:
The relevant issue only requested longitude wrapping, but we may as well allow latitude wrapping. The problem is that longitude wrapping seems more intuitive and a more common problem than latitude wrapping, since `longitude 180 == -180`, so `longitude 181 == -179`, while `latitude 91 == 89`, which means you have to wrap it differently *and* shift longitude 180 degrees. I'm not too familiar with the problems involved, but [this elasticsearch issue](https://github.com/elastic/elasticsearch/issues/1997) sums them up fairly well. In short, latitude wrapping seems generally.. hairier.

Anyway, elasticsearch currently handles latitude wrapping the right way, so assuming that the client does so as well (in that it flips the longitude value when latitude exceeds 90), everything should Just Work if we remove the sanitizer. @missinglink , @hkrishna , any reason to not implement it?

sevko:
Test-cases added.

sanitiser/_geo.js
```diff
@@ -46,26 +47,25 @@ function sanitize_bbox( clean, param ) {
...
-  var bbox = [];
   var bboxArr = param.split( ',' );
 
   if( Array.isArray( bboxArr ) && bboxArr.length === 4 ) {
 
-    bbox = bboxArr.filter( function( latlon, index ) {
-      latlon = parseFloat( latlon, 10 );
-      return !(lat_lon_checks[(index % 2 === 0 ? 'lon' : 'lat')].is_invalid( latlon ));
+    var bbox = bboxArr.map( function ( latlon ){
+      return parseFloat( latlon, 10 );
```
missinglink: `parseFloat()` doesn't accept a second arg like the confusingly mandatory `radix` one from `parseInt()`

sanitiser/_geo.js
```diff
@@ -46,26 +47,25 @@ function sanitize_bbox( clean, param ) {
...
+    });
+    bboxArr.forEach( function( latlon, index ) {
+      if( index % 2 === 1 && check_lat.is_invalid( latlon ) ){
+        throw new Error( 'Invalid lat: ' + latlon );
```
missinglink: [question] I see the unit tests pass, by what magic do we not also need `lng` validation in this block of code?

sevko: Are you just referring to `isNaN()`?

missinglink: makes sense, hard to get your head around...
:+1:

sanitiser/_geo.js
sevko: Good call.

sevko:
@missinglink , @hkrishna , any thoughts on wrapping longitude in addition to latitude?

sevko:
We've decided to wrap latitude as well.

sanitiser/_geo.js
```diff
@@ -78,15 +81,15 @@ function sanitize_bbox( clean, param ) {
  * @param {bool} latlon_is_required
  */
 function sanitize_coord( coord, clean, param, latlon_is_required ) {
-  var value = parseFloat( param, 10 );
+  var value = parseFloat( param );
   if ( !isNaN( value ) ) {
-    if( lat_lon_checks[coord].is_invalid( value ) ){
-      throw new Error( 'invalid ' + lat_lon_checks[coord].error_msg );
+    if( coord === 'lat' && check_lat.is_invalid( value ) ){
```
dianashk: don't do this check anymore. if `lon` isn't being validated, `lat` doesn't need to be either.

dianashk:
:+1:

orangejulius:
This PR is obsoleted by #178

Issue #118, opened 2015-04-28
Add categories and address properties to result doc


dianashk:
Add new properties to result objects:
```javascript
{
  categories: [],
  address: {
    street: 'Chuck St.',
    // etc...
  }
}
```

These should only be returned when `details` flag is set to `true` in request.

PR #122, opened 2015-05-04
Add categories and address properties to result doc

dianashk:
Fixes pelias/api#118

### DIFF
--- a/helper/geojsonify.js
+++ b/helper/geojsonify.js
@@ -77,13 +77,11 @@ function search( docs, params ){
  * @param {object} dest
  */
 function copyProperties( source, props, dest ) {
-  var count = props.length;
-  for ( var i = 0; i < count; i++ ) {
-    var prop = props[i];
+  props.forEach( function ( prop ) {
     if ( source.hasOwnProperty( prop ) ) {
       dest[prop] = source[prop];
     }
-  }
+  });
 }
 
 
--- a/helper/geojsonify.js
+++ b/helper/geojsonify.js
@@ -27,7 +27,7 @@ function search( docs, params ){
   var geodata = docs.map( function( doc ) {
 
     // something went very wrong
-    if( !doc || !doc.center_point ) {
+    if( !doc || !doc.hasOwnProperty( 'center_point' ) ) {
       return warning();
     }
 
--- a/helper/geojsonify.js
+++ b/helper/geojsonify.js
@@ -3,23 +3,33 @@ var GeoJSON = require('geojson'),
     extent = require('geojson-extent'),
     outputGenerator = require('./outputGenerator');
 
-function search( docs, params ){
+// Properties to be copied when details=true
+var DETAILS_PROPS = [
+  'alpha3',
+  'admin0',
+  'admin1',
+  'admin1_abbr',
+  'admin2',
+  'local_admin',
+  'locality',
+  'neighborhood',
+  'category',
+  'address'
+];
 
-  // emit a warning if the doc format is invalid
-  // @note: if you see this error, fix it ASAP!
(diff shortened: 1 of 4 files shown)


dianashk:
If `category === []` or `address === {}` should we not include them in the result or would there be value in showing them? I'm leaning towards not sending them, as to not waste bandwidth. Thoughts?

sevko:
I'd say not sending them is consistent with not sending properties with `undefined` values. I wouldn't worry about the cost of a few additional characters per record, but I guess it adds up when you apply the same thinking to all the other properties we're not sending.

missinglink:
There are pros/cons to leaving the empty lists in the JSON:

**pros**
- default behaviour of `JSON.stringify`
- potentially easier to parse (less client-side existence checks)

**cons**
- adds volume to HTTP response body

If there is a way we could achieve this in js without adding nasty blocking/recursive code in our controllers I'd be keen, otherwise we would be trading HTTP response size for controller CPU cycles.

I envision it looking something not-so-pretty like this, which would also require a bit of memory for each copy:

```javascript
function removeEmptyLists( obj ){
  var copy = {};
  for( var attr in obj ){
    var prop = obj[attr];
    if( Array.isArray( prop ) ){
      if( prop.length ){
        copy[attr] = prop;
      }  
    } else if( typeof prop === 'object' && prop !== null ){
      if( Object.keys(prop).length ){
        copy[attr] = prop;
      }
    }
    else {
      copy[attr] = prop;
    }
  }

  return copy;
}
```

sevko:
It sounds an awful lot like we're trying to prematurely optimize. Let's do what makes intuitive sense and worry about payload size and CPU cycles (I'd argue the latter is presently totally negligible) when they become problems.

helper/geojsonify.js
```diff
@@ -67,4 +68,34 @@ function search( docs, params ){
...
+ * Copy specified properties from source to dest.
+ * Ignore missing properties.
+ *
+ * @param {object} source
+ * @param {[]} props
+ * @param {object} dest
+ */
+function copyProperties( source, props, dest ) {
+  var count = props.length;
+  for ( var i = 0; i < count; i++ ) {
```
sevko: @dianashk , not sure if you're soliciting code review, but would a `.forEach()` make sense here? Might make it more immediately clear what this is doing.

dianashk:
Since there is no strong feelings towards either, I'm going to leave them in for now. If payload size becomes a concern, we can consider cleaning these up then.

helper/geojsonify.js
```diff
@@ -3,23 +3,33 @@ var GeoJSON = require('geojson'),
...
   var details = params ? params.details : {};
   details = details === true || details === 1;
 
   // flatten & expand data for geojson conversion
-  var geodata = docs.map( function( doc ){
+  var geodata = docs.map( function( doc ) {
 
     // something went very wrong
-    if( !doc ) { return warning(); }
+    if( !doc || !doc.center_point ) {
```
missinglink: nit picking.. `!doc.hasOwnProperty('center_point')` it's really not a big deal but the original code I wrote was wrong and should have performed a key check rather than a value check here.

missinglink:
[question] will `.address` and `.category` be on by default or require `?details=1`, I would prefer the latter

dianashk:
They only get sent when `details=true`, which happens to be the default. if the user says `details=false` they will only get `.layer`, `.text`, and `.id`.

sevko:
+1

Can you document this parameter in DOCS.md?

sevko:
ack, my mistake.

missinglink:
:+1:

PR #126, opened 2015-05-06
some DOCS.md refactoring

sevko:
* remove redundant parameter documentation across different endpoints, replacing it with `see documentation for /${endpoint}`
  * replace inline use of \`\`\`code\`\`\` with \`code\`
  * use a consistent format for documenting default values


dianashk:
nice :+1:

dianashk:
We should probably remove the copy of this doc from the wiki so we don't have to update several places each time we make changes.

sevko:
Absolutely, doing that now.

PR #131, opened 2015-05-12
Use multi_match in name.* for /search

### LINKED ISSUE #127: /search against all name[*] properties
currently `/search` is only searching against [name.default](https://github.com/pelias/api/blob/master/query/search.js#L28), it would be better if the search was performed against any of the `name[*]` values.

rel: pelias/pelias#83
re: pelias/schema#49

dianashk:
Closes #127

### DIFF
--- a/query/search.js
+++ b/query/search.js
@@ -22,8 +22,9 @@ function generate( params ){
   query.query.filtered.query = {
     'bool': {
       'must': [{ 
-          'match': {
-            'name.default': params.input
+          'multi_match': {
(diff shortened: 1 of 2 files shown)


hkrishna:
:+1:

sevko:
LGTM, but do you mind if we hold off on merging this until we sort out our current dev import? It's broken a lot of tests, and I don't want to confuse those problems with any breaking changes that this might cause (unless you've already deployed and tested this branch).

PR #154, opened 2015-07-07
remove exact_match script

missinglink:
remove deprecated `exact_match` script

in a basic dev test this deletion improved response time from 32ms to 22ms


dianashk:
:+1:

Issue #139, opened 2015-05-26
[discuss] lat/lon bias


missinglink:
the lat/lon bias is currently set up to find only documents withing 50km of the supplied centroid and then the results are sorted by distance from the centroid.

there are a couple of issue with this:
- 50km is an arbitrary number, we could either:
 - allow the API consumer to adjust this (within sane limits) **or**
 - potentially increase this number to cover a wider area
- tends to rank better linguistic matches lower than geographically closer ones

it might be worth setting the 50km constant higher and removing the sorting completely.

note: `/reverse` will still need the sorting

missinglink:
Had a crack at resolving it this morning here https://github.com/geopipes/elasticsearch-backend/pull/7, once that is merged the module version will need to be updated and the `sort:true` flag added to `/reverse`.

I think we should also expand `50km` to something higher like `200km` or maybe vary it on zoom?

PR #156, opened 2015-07-13
Issue 139

missinglink:
issue detail: https://github.com/pelias/api/issues/139

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,16 +37,16 @@
     "express": "^4.8.8",
     "geojson": "^0.2.1",
     "geojson-extent": "^0.3.1",
-    "geopipes-elasticsearch-backend": "0.0.12",
+    "geopipes-elasticsearch-backend": "^0.2.0",
     "is-object": "^1.0.1",
     "markdown": "0.5.0",
-    "pelias-esclient": "0.0.25",
-    "pelias-logger": "^0.0.8",
-    "through2": "0.6.5",
+    "microtime": "1.4.0",
     "morgan": "1.5.2",
     "pelias-config": "^0.1.4",
-    "microtime": "1.4.0",
-    "pelias-suggester-pipeline": "2.0.2"
+    "pelias-esclient": "0.0.25",
+    "pelias-logger": "^0.0.8",
+    "pelias-suggester-pipeline": "2.0.2",
+    "through2": "0.6.5"
   },
   "devDependencies": {
     "ciao": "^0.3.4",
--- a/query/reverse.js
+++ b/query/reverse.js
@@ -9,7 +9,12 @@ function generate( params ){
     lon: params.lon
   };
 
-  var query = queries.distance( centroid, { size: params.size || 1, sort: true } );
+  var query = queries.distance( centroid, {
+    size: params.size || 1,
+    sort: true,
+    distance: '500km'
+  });
+
   query.sort = query.sort.concat( sort( params ) );
 
   if ( params.categories && params.categories.length > 0 ) {
--- a/query/reverse.js
+++ b/query/reverse.js
@@ -9,7 +9,7 @@ function generate( params ){
     lon: params.lon
   };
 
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -37,16 +37,16 @@
     "express": "^4.8.8",
     "geojson": "^0.2.1",
     "geojson-extent": "^0.3.1",
-    "geopipes-elasticsearch-backend": "0.0.12",
+    "geopipes-elasticsearch-backend": "^0.2.0",
```
missinglink: I only bumped this version, the rest is `npm` re-ordering alphabetically

query/reverse.js
```diff
@@ -9,7 +9,12 @@ function generate( params ){
     lon: params.lon
   };
 
-  var query  =  queries.distance( centroid, { size: params.size || 1 } );
+  var query = queries.distance( centroid, {
+    size: params.size || 1,
+    sort: true,
+    distance: '500km'
```
missinglink: this is the guts of it, enable sorting and increase the search radius

hkrishna:
:+1:

query/reverse.js
hkrishna: why increase search radius?

missinglink:
```bash
$ npm publish
npm http PUT https://registry.npmjs.org/pelias-api
npm http 200 https://registry.npmjs.org/pelias-api
+ pelias-api@2.1.0
```

query/reverse.js
riordan: re: @harish's question about search radius:

Imagine the use case we're dealing with in Open by Mapzen, where we have a
mobile app that's being used (heavily) for driving directions. And it uses
today's geobias, which limits to 50km.


Great for getting to somewhere nearby (a school across town, a new
restaurant), but for longer trips (e.g. road trips to a different
province), that 50km radius won't work. 200's an attempt to patch that. It
doesn't solve for the magical universal geocoder that prefers places
closer, but still has a great global option for when the local places
aren't really matching (e.g starting in NYC and looking for an address in
Ottawa) but it's a start.

@insertcoffee - sound about right?



On Mon, Jul 13, 2015 at 11:31 AM, Harish Krishna <notifications@github.com>
wrote:

> In query/reverse.js
> <https://github.com/pelias/api/pull/156#discussion_r34475295>:
>
> > @@ -9,7 +9,12 @@ function generate( params ){
> >      lon: params.lon
> >    };
> >
> > -  var query  =  queries.distance( centroid, { size: params.size || 1 } );
> > +  var query = queries.distance( centroid, {
> > +    size: params.size || 1,
> > +    sort: true,
> > +    distance: '500km'
>
> why increase search radius?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/pelias/api/pull/156/files#r34475295>.
>



-- 
David Riordan | Product Manager - Search | dave.riordan@mapzen.com |
@riordan <https://twitter.com/riordan> | gpg 235D9DC95EF6277C
<https://keybase.io/riordan>
Mapzen | https://mapzen.com | @mapzen <https://twitter.com/search>

missinglink: hmmm... to be honest I am confused on this one myself.

If we set the `distance` to say 500km (a solid half day drive) it *could* potentially still not show your final destination, causing confusion that could be resolved only really by using `no geo bias` (slower queries and listing many far away places for poorly specified inputs)

The original idea was that if someone sent their lat/lon to the service it would mean that they would like to get results biased towards their location. I'm not sure if this is valid, at least not always true.

We should encourage clients to send the device location when it is available.

Breaking away from the computer and code for a second I think it should work like this.

1) When you specify a query such as `McDonald's` (with no comma, or other admin info)
2) Pelias should default that admin bias to the city in which you reside (so for me Berlin, for you NYC) -- how strong this bias is will need to be balanced
3) When the user types a comma and provides an admin segment (eg. `McDonalds, London`) then the device lat/lon is no longer considered as the user has specified their desired geo bias

thoughts?

PR #163, opened 2015-07-15
Error handling bug fix

hkrishna:
```err``` is of type ```string``` if there is an error within express app - its of type ```object``` (with ```err.message``` being the error string) in case of errors messages from the ES cluster

```var error = err.message ? err.message : err;``` catches both cases nicely.

### DIFF
--- a/middleware/408.js
+++ b/middleware/408.js
@@ -2,7 +2,7 @@
 // handle time out errors
 function middleware(err, req, res, next) {
   res.header('Cache-Control','no-cache');
-  var error = err.message ? err.message : err;
+  var error = (err && err.message) ? err.message : err;
   
   if( res.statusCode === 408 || (error.toLowerCase().indexOf('request timeout') !== -1) ){ 
   	res.status(408); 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by missinglink** [comment]
doesn't `var error = err.message ? err.message : err;` throw a fatal error when `err` is `null`?

**Comment by hkrishna** [comment]
Good Catch! I changed it to ```err & err.message```

File: middleware/408.js [hunk 1]
```diff
@@ -2,9 +2,11 @@
 // handle time out errors
 function middleware(err, req, res, next) {
   res.header('Cache-Control','no-cache');
-  if( res.statusCode === 408 || (err.message.toLowerCase().indexOf('request timeout') !== -1) ){ 
+  var error = (err && err.message) ? err.message : err;
+  
+  if( res.statusCode === 408 || (error.toLowerCase().indexOf('request timeout') !== -1) ){ 
   	res.status(408); 
-  	res.json({ error: err && typeof err.message === 'string' ? err.message : 'request time out' });
+  	res.json({ error: typeof error === 'string' ? error : 'request time out' });
```
dianashk: I think timeout should be one word.

dianashk:
:+1:

PR #164, opened 2015-07-15
fix broken test in master

hkrishna:
I think this got introduced with https://github.com/pelias/api/pull/156


hkrishna:
nevermind - false alarm. closing..

PR #165, opened 2015-07-15
Remove all suggest code

hkrishna:
Code Cleanup

- removed all things that involved suggester including query mixer
- kept the suggest sanitizer for suggest/nearby wherein lat/lon are required
- 843 deletions


missinglink:
is any of this code useful to keep for when ES@2.0 lands?

hkrishna:
not really - we wont need multiple suggesters because ES 2.0 promises to handle precision array ```[5,4,3,2,1]``` gracefully and mix results in the order given to it. If not, we can always refer back in history to see this code and maybe even bring it back. 

Query Mixing is something we may need to bring back if we want to mix only admin layers when the input length is a certain characters and other layers when its longer and other cases. 

However, for now - I think this code should go away and based on what lands in ES 2.0 and how it changes the context suggester - we should write fresh queries. Dead code is bad for maintainability.

missinglink:
:+1:

missinglink:
this can be merged and released now

Issue #128, opened 2015-05-08
filter /search by categories


sevko:
Extend the search logic added in #115 , which added support for filtering `/reverse` results by record category, to `/search` and `/suggest`.

missinglink:
adding the categories to `/search` is difficult yet possible, have a read of this comment https://github.com/elastic/elasticsearch/issues/6444#issuecomment-64408912

missinglink:
this is quite interesting as a opportunity to refactor the 'admin' (coarse) and 'address' query aliases to use categories, meaning we would no longer have to hard-code which datasets contained which data

riordan:
A category may be combined with a `/search?text` so a user could search for `category=bank` and `text=wells fargo`.

orangejulius:
Updating the title as we only want /search to accept the categories parameter (extra parameters on /suggest reduces the effectiveness of caching).

riordan:
Hold for epic: pelias/pelias#150

PR #168, opened 2015-07-20
[DO NOT MERGE] added category sanitiser to /search and updated search query

hkrishna:
This is part of the pair programming between me and @orangejulius (was fun!)
got tests to pass: by not setting an empty array as categories

fixes https://github.com/pelias/api/issues/128

### DIFF
--- a/query/search.js
+++ b/query/search.js
@@ -46,10 +46,8 @@ function generate( params ){
 
   // add category mapping
   if (params.categories) {
-    query.query.filtered.query.bool.should.push({
-      'match': {
-        'category': params.categories
-      }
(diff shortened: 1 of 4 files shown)


missinglink:
looks good, can we add a simple test case which covers this specific behaviour?

something like setting the input to 'foo', 'foo,bar' and '' then checking the generated query to ensure the `query.filtered.filter.bool.must` section is not added?

hkrishna:
yes! agreed @missinglink this PR is not ready for merge.. tests are yet to be written - @orangejulius is aware of that :)

orangejulius:
#186 is a better version of this code :)

PR #169, opened 2015-07-21
Address Parser

hkrishna:
Fixes pelias/pelias#40

This PR aims at taking a first crack at address parsing using the npm module [addressIt](https://www.npmjs.com/package/addressit)

AddressIt is a freeform street address parser, that is designed to take a piece of text and convert that into a structured address that can be processed in different systems.

```bash
> var addressit = require('addressit')

> addressit('123 main st new york ny 10010 usa')
{ text: '123 main st new york ny 10010 usa',
  parts: [],
  unit: undefined,
  number: 123,
  street: 'main st',
  state: 'NY',
  country: 'USA',
  postalcode: 10010,
  regions: [ 'new york' ] }
```

Before the pelias API calls addressit for address parsing, it does some basic checks by parsing query to ensure that we dont slow things down drastically when unnecessary for example the following are the cases where we dont need address parsing - 

- ```input=a``` or ```input=au``` or ```input=aus``` - if the input has 3 or less characters, we could assume its not a fully formed address, in fact - we can go one step further by only targeting admin layers because if we return results such as ```austin```, ```australia``` etc it should be relevant but more importantly fast. 
- ```input=boston``` or ```input=frankfurt``` or ```input=somereallybigname``` or ```input=new york``` - if the input is just one or even two tokens and does not contain a number - we can get away with just targeting ```admin``` and ```poi``` layers

In all other cases, we do address parsing and handle the address parts to query the ES index. Here's a sample mapping 

```
number + street -> name.default
number -> address.number
street -> address.street
postalcode -> address.zip
state -> admin1_abbr
country -> alpha3
regions -> admin2
```

Sometimes, the address parser comes back empty handed

```bash
> addressit('123 chelsea, london')
{ text: '123 chelsea, london',
  parts: [],
  unit: undefined,
  state: undefined,
  country: undefined,
  postalcode: undefined,
  regions: [ '123 chelsea', 'london' ] }
```

In this case, we take fall back to the naive approach we implemented months ago - where we split the address based on a comma and assume everything that follows the comma is an admin part and add a ```match``` block in the ```should``` array. So, we query ```name.default``` with ```123 chelsea``` and the ```should``` array in the query would try to match ```london``` with all the 5 admin fields

- ```admin0```
- ```admin1``` 
- ```admin1_abbr```
- ```admin2```
- ```alpha3```

All of this logic lives in ```helper/query_parser.js``` and is well documented with in-line comments. The query changes can be seen in ```query/search.js```.

An additional 104 test cases were written to test out all the above mentioned logic and to test the query building - bringing the grand total of unit tests for the API to 708! 

This PR works well with US addresses but its the first humble pass at address parsing. Support for other regions can be added to addressit [locale](https://github.com/DamonOehlman/addressit/tree/master/locale)

Additionally, a bunch of acceptance tests were added to test address_parsing in both ```dev``` and ```prod/bigdev``` instances of pelias https://github.com/pelias/acceptance-tests/pull/123

:beers: for anyone who reviews/comments. thanks!


missinglink:
looking good!

helper/query_parser.js
```diff
@@ -0,0 +1,81 @@
+
+var parser     = require('addressit');
+var extend     = require('extend');
+var get_layers = require('../helper/layers');
+var delim      = ',';
+
+module.exports = function(query) {
+  var parsedAddress0 = {};
```
orangejulius: It could be nice to move each of these parsed addresses out into their own function. Presumably each one will be tweaked over time, and having them separated will help keep things clear.

test/unit/helper/query_parser.js
```diff
@@ -0,0 +1,168 @@
...
+      var admin_parts = query.substring(delimIndex + 1).trim();
+
+      t.equal(typeof address, 'object', 'valid object');
+      t.equal(address.name, name, 'name set correctly to ' + address.name);
+      t.equal(address.admin_parts, admin_parts, 'admin_parts set correctly to ' + address.admin_parts);
+      t.end();
+    });
+  };
+
+  for (var keys in queries) {
```
orangejulius: `keys` actually has just a single value out of the array, right? so is `key` a better name? It's easy to misread and think keys contains a collection of some kind

test/unit/helper/query_parser.js
```diff
@@ -0,0 +1,168 @@
...
+      } else {
+        t.deepEqual(address.target_layer, target_layer, 'admin_parts set correctly to ' + target_layer.join(', '));  
+      }
+      
+      t.end();
+    });
+  };
+
+  var queries = one_token_queries.concat(two_tokens_nonum);
+  for (var keys in queries) {
```
orangejulius: same keys vs key again

test/unit/helper/query_parser.js
```diff
@@ -0,0 +1,168 @@
...
+      
+      if (address.text === address.regions.join(' ')) {
+        t.deepEqual(address.target_layer, query.target_layer, 'admin_parts set correctly to ' + query.target_layer.join(', '));  
+      }
+      
+      t.end();
+    });
+  };
+
+  for (var keys in addresses_nonum) {
```
orangejulius: same keys vs key again

test/unit/helper/query_parser.js
hkrishna: good point. i'll change it

test/unit/helper/query_parser.js
```diff
@@ -0,0 +1,168 @@
...
+      var target_layer = get_layers(['admin']);
+
+      t.equal(typeof address, 'object', 'valid object');
+      t.deepEqual(address.target_layer, target_layer, 'admin_parts set correctly to ' + target_layer.join(', '));
+      t.end();
+    });
+  };
+
+  var queries = chars_queries.concat(num_queries).concat(alphanum_q);
+  for (var keys in queries) {
```
orangejulius: oh and one more here

riordan:
Excellent summary of changes and damn fine work.

:+1: 

Looking forward to building on top of this.

controller/search.js
```diff
@@ -21,6 +21,11 @@ function setup( backend, query ){
       cmd.type = req.clean.layers;
     }
 
+    // set type if input suggests targeting a layer(s)
```
orangejulius: Can you think of any easy ways to make the logic for selecting target layers less brittle? Right now it relies on a flag that's set pretty far away, and then there are at least four different values here that determine the final value of cmd.type. lots of paths to think about when making future changes. Even if it's something we don't implement today, but a refactoring we work towards, i'd love to hear some thoughs.

heffergm:
Three cheers for @hkrishna :beers:

missinglink:
:+1: :beers:

orangejulius:
Niiice. definitely a :+1: and some :beers: and some :bread: and some :family: and some :hamburger: and some :ice_cream: and some :bicyclist:

controller/search.js
hkrishna: It would be nice to set ```clean.default_layers``` at the ```_layers``` sanitiser - that way we wont have to set ```default_layers_set``` flag. Furthermore, we can just fall back to the ```default_layers``` when ```target_layers``` and ```clean.layers``` are not set. And that can be done in a one liner in the search controller if required :D 

But this shouldn't delay this PR from getting merged. I think that layer rewrite should be a tech-debt-y issue in the API.. https://github.com/pelias/api/issues/170

dianashk:
:+1:

PR #171, opened 2015-07-27
phrase slop query modifications

missinglink:
this is the API PR to change the query logic slightly to account for https://github.com/pelias/schema/pull/70

I also included an update which explicitly specifies the analyzer to use for ngrams queries (peliasOneEdgeGram) as autocomplete for single letter inputs is currently not working (it should be enabled for admin layers only)

note: the schema PR should be merged before this PR and after a staging build and review.

Closes https://github.com/pelias/schema/issues/66

### DIFF
--- a/query/search.js
+++ b/query/search.js
@@ -102,7 +102,10 @@ function generate( params ){
   // add search condition to distance query
   query.query.filtered.query.bool.must.push({ 
     'match': {
-      'name.default': input
+      'name.default': {
+        'query': input,
+        'analyzer': 'peliasOneEdgeGram'
+      }
     }
   });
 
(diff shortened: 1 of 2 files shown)


query/search.js
```diff
@@ -102,7 +102,10 @@ function generate( params ){
   // add search condition to distance query
   query.query.filtered.query.bool.must.push({ 
     'match': {
-      'name.default': input
+      'name.default': {
+        'query': input,
+        'analyzer': 'peliasOneEdgeGram'
```
dianashk: what do you think about API being more tightly coupled with schema? I don't like the fact that we just have strings like this around without any validation that corresponding schema version contains them. I'm working on coupling schema with API for admin fields.

PR #186, opened 2015-08-04
Add categories filter to /search

### LINKED ISSUE #128: filter /search by categories

orangejulius:
This also adds some tests for the categories filter, since there weren't any.

Fixes #128 
Replaces #168


hkrishna:
looking good :+1: 

(however, as an iteration - we should probably sanitize categories to limit to just the ones that map well with our data and ignore gibberish values such as ```abc``` or ```def```)

riordan:
Hold for pelias/pelias#150

dianashk:
Outdated due to too many changes in API code.
Closing for now.

PR #192, opened 2015-08-10
Search query builder not checking all available admin values

### LINKED ISSUE #187: Search query builder not checking all available admin values
Query builder logic in search doesn't take into account all the available admin values. The logic is flawed in that it starts with an empty list of unmatched admin fields and only adds those that weren't explicitly found.  See [HERE](https://github.com/pelias/api/blob/master/query/search.js#L67-L98) That leaves all _other_ admin fields unsearchable resulting in failures such as this:

__input:__
> 1 Water Street Manhattan NY

__results:__
> 1: "1 Water Street, Hanover, NY"
> 2: "1 Water Street, Charlotte, NY"
> 3: "1 Water Street, Van Buren, NY"
> __4: "1 Water Street, Manhattan, NY"__ <--- should be first!!!

dianashk:
Fixes #187 

* Not all admin fields were being used by the `query builder`. Refactored a bit and added `adminFields.js` where the fields are validated against the current schema.
* Added `admin` to category boosts to help `soho, new york` pass. This should be used in all cases, except when we know the query is an address.
* Added boosting for each address part we've parsed out with certainty. We can play around with the actual boost numbers separately.

### DIFF
--- a/query/search.js
+++ b/query/search.js
@@ -1,6 +1,9 @@
 
 var queries = require('geopipes-elasticsearch-backend').queries,
-    sort = require('../query/sort');
+    sort = require('../query/sort'),
+    adminFields = require('../helper/adminFields').availableFields,
+    addressWeights = require('../helper/address_weights');
+
 
 function generate( params ){
   var centroid = null;
@@ -27,22 +30,7 @@ function generate( params ){
   };
 
   if (params.parsed_input) {
-
-    query.query.filtered.query.bool.should = [];
-
-    var unmatched_admin_fields = [];
-    // qb stands for query builder
-    var qb = function(unmatched_admin_fields, value) {
-      if (value) {
-        unmatched_admin_fields.forEach(function(admin_field) {
-          var match = {};
-          match[admin_field] = value;
-          query.query.filtered.query.bool.should.push({
-             'match': match
-          });
-        });  
-      }
-    };
+    addParsedMatch(query, input, params.parsed_input);
 
     // update input
     if (params.parsed_input.number && params.parsed_input.street) {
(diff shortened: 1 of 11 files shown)


riordan:
That was quick!

query/search.js
```diff
@@ -119,4 +60,67 @@ function generate( params ){
   return query;
 }
 
+function addParsedMatch(query, defaultInput, parsedInput) {
+  query.query.filtered.query.bool.should = [];
```
dianashk: check if doesn't exist

query/search.js
```diff
@@ -119,4 +60,67 @@ function generate( params ){
...
+      boost: boost
+    };
+  }
+  else {
+    match[key] = value;
+  }
+
+  query.query.filtered.query.bool.should.push({ 'match': match });
+
+  var index = unmatched.indexOf(key);
```
dianashk: add comment as to what this is doing

test/unit/helper/adminFields.js
```diff
@@ -0,0 +1,84 @@
...
+
+    var expectedFields = require('../../../helper/adminFields').expectedFields;
+    var schemaMock = { mappings: { _default_: { properties: {} } } };
+
+    // inject all expected fields into schema mock
+    expectedFields.forEach(function (field) {
+      schemaMock.mappings._default_.properties[field] = {};
+    });
+
+    var adminFields = proxyquire('../../../helper/adminFields', {'pelias-schema': schemaMock});
```
dianashk: should we continue using proxyquire?

test/unit/query/search.js
```diff
@@ -267,27 +267,72 @@ module.exports.tests.query = function(test, common) {
...
-                   'alpha3': 'USA'
+                   'alpha3': {
+                     'query': 'USA',
+                     'boost': 5
+                   }
+                 }
+               },
+               {
+                 match: {
+                   admin0: 'new york'
```
dianashk: investigate if these are using the right analyzers

test/unit/query/search.js
```diff
@@ -267,27 +267,72 @@ module.exports.tests.query = function(test, common) {
...
+                     'boost': 3
+                   }
                  }
                },
                {
                  'match': {
-                   'address.zip': 10010
+                   'address.zip': {
+                     'query': 10010,
+                     'boost': 3
```
dianashk: pull boost values out of config file

test/unit/query/search.js
hkrishna: admin0 uses the standard peliasAdmin analyzer https://github.com/pelias/schema/blob/master/mappings/partial/admin.json#L3

This will allow partial matching - in this case, ```new jersey``` and ```new mexico``` will be partial matches. If you want to use ```keyword``` analyzer - you get 0 or 1 match.. for example, if ```New York``` is how its stored in the index - you can only match it with ```New York``` and not ```new york``` (yes, its case sensitive)

query/search.js
```diff
@@ -119,4 +60,67 @@ function generate( params ){
...
+
+  // address
+  // number, street, postalcode
+  addMatch(query, unmatchedAdminFields, 'address.number', parsedInput.number, addressWeights.number);
+  addMatch(query, unmatchedAdminFields, 'address.street', parsedInput.street, addressWeights.street);
+  addMatch(query, unmatchedAdminFields, 'address.zip', parsedInput.postalcode, addressWeights.zip);
+
+  // city
+  // admin2, locality, local_admin, neighborhood
+  addMatch(query, unmatchedAdminFields, 'admin2', parsedInput.admin2, addressWeights.admin2);
```
hkrishna: please check if its ```parsedInput.city``` or ```parsedInput.admin2``` - I remember ```addressIt``` has no notion of admins.. 

I just looked at master - [this](https://github.com/pelias/api/blob/master/query/search.js#L69) may be a bug in master because ```parsedInput.admin2``` would always be ```undefined```

dianashk: Good call! Thanks.

orangejulius:
This is excellent. :+1:

PR #205, opened 2015-08-20
Filter on alpha3

### LINKED ISSUE #112: Add boundary.country parameter to /search and /reverse endpoints
The geometry of some country polygons makes it difficult to use bbox filtering to exclude neighbouring nations.

It would be nice to have an API param called `alpha3` or similar which excluded results from outside that country.

When used in combination with a coarse geocoder it would be very easy to create a coarse search for a single country using the mapzen service.

orangejulius:
This PR adds an alpha3 parameter to the /search endpoint. It ensures Elasticsearch only returns results matching that alpha3 code via a must filter. A sanitiser checks that the alpha3 parameter is a 3 character string, but doesn't check for validity. Invalid alpha3s (i.e. ones not assigned to any country) will not throw an error, but also won't return any results (obviously). That said I'd like to build an npm module that exports a list of all countries filtered by alpha2, alpha3, etc so that a friendly warning could be returned in the case of an invalid alpha3. But that's for another day.

Fixes #112


riordan:
Excellent!

orangejulius:
@missinglink, should the analyzer for the admin3 field be changed to keyword? anything other than an exact match is not interesting

sanitiser/_alpha3.js
```diff
@@ -0,0 +1,25 @@
+// validate inputs, convert types and apply defaults
+function sanitize( req ){
```
dianashk: this needs to now match the new convention. please merge master into the branch. sanitize functions now take `(raw, clean)`

sanitiser/_alpha3.js
```diff
@@ -0,0 +1,25 @@
...
+  var params = req.query;
+  req.clean = req.clean || {};
+
+  if (params === undefined || params.alpha3 === undefined) {
```
dianashk: shouldn't the param name be `boundary.country`?

orangejulius: Really old code in this PR, sorry

sanitiser/_alpha3.js
orangejulius: Really old code in this PR, sorry

orangejulius:
Really old code in this PR

PR #206, opened 2015-08-25
Add deprecation notice to Beta version

dianashk:
Did the simplest thing that would work. Ideally the geojsonify code should be a middle-wear function,
in which case adding the deprecation warning would also be a new middle-wear component that could easily be turned on and off without impacting other existing code. We will address this refactoring in a different PR, probably as part of the v1 infrastructure changes.

Closes pelias/pelias#118


dianashk:
Merging into **`__BETA__`** branch instead. Doesn't belong in **`master`**

PR #213, opened 2015-09-02
Node based proxy for v0 api

### LINKED ISSUE #195: Introduce API semantic versioning
# Versioning Schema
Introduce schema into API URL structure to support multiple versions of the Pelias API.

## Description
Before any API call, the version of the API must be specified.

For instance, to run a `search` query, the url would be `https://pelias.mapzen.com/v1/search`.

This **may** also allow us to introduce [Semantic Versioning](https://semver.org/) into the Pelias API. So `pelias.mapzen.com/v1/1/2/search` could be a patch to version 1.1, which wouldn't break responses for any users of `/v1/0/01`.

Taken from conversations in pelias/api#182.

orangejulius:
Using [express-http-proxy](https://www.npmjs.com/package/express-http-proxy), send requests to the old (v0) API. This is only active when a config entry for legacyUrl is present in the pelias config. Something like the following:

	"api": {
		"legacyUrl": "http://pelias.mapzen.com"
	},

Fixes #195 (at least as much as required for v1)


dianashk:
:+1:

PR #219, opened 2015-09-04
Confidence score calculation

dianashk:
Fixes pelias/pelias#120


dianashk:
Will merge into `master` instead of `responses` branch

PR #221, opened 2015-09-04
Add source parameter

orangejulius:
This adds the source parameter described #181. It allows limiting search results to only those matching a particular data source (geonames, quattroshapes, openaddresses, and openstreetmap are the currently available sources).

The filtering done with this parameter is _cumulative_ with the layers parameter, and it is possible to specify combinations of layers and source parameters that can't possibly return any results, such as source=quattroshapes&layers=address. Additionally, the address parser uses some logic to determine when it is safe to query only the admin layer for performance improvement (for example, queries with a text parameter only a few characters long).

For that reason, this branch also contains a major refactoring to all code relating to layers, as well as the address parser. It splits the address parser functionality into two sections, one dealing with the actual address parsing, and one dealing with determining when the search results can be constrained to the admin layer. It also refactors address parsing and layer parameter handling code to simply submit a requested list of types, rather than directly manipulating the list of types sent to Elasticsearch.
Finally, new logic is added in middleware/types and helpers/types to look at types specified by the layers parameter, the source parameter, and the address parser, and decide how to combine them.

The types middleware detects when a combination of layers and source parameter won't return any results, and provides a friendly error message. This allows the full power of layers and source to be used in combination when it makes sense, while ensuring combinations that don't make sense are discoverable quickly.

Fixes #181

PR #222, opened 2015-09-04
Remove token based layer filter

### LINKED ISSUE #194: Query parser is too restrictive with certain languages
Currently the search doesn't find any addresses in Finnish, if the user doesn't specify a street number, even if both the street name and city have been specified.

[Example](http://pelias.mapzen.com/search?input=Opastinsilta,%20Helsinki)

However, if you explicitly specify to include the address layer it works correctly [(example)](http://pelias.mapzen.com/search?input=Opastinsilta,%20Helsinki&layers=address).

This has to do with the fact that most of Finnish street names are compound words, and thus, "Opastinsilta, Helsinki" becomes only two tokens. It is also confusing to the users that it works some times, e.g. when the street name contains a personal name [(example)](http://pelias.mapzen.com/search?input=Urho%20Kekkosen%20katu,%20Helsinki).

orangejulius:
Primarily as a performance optimization, but also to attempt to return
more relevant results, only admin and POI layers were queried when the
text input consisted of only one or two tokens, and there weren't any
numbers. However as shown in #194 that is a bit too optimistic, especially
in countries other than the USA.

Fixes #194, should only be merged after #221

### DIFF
--- a/helper/query_parser.js
+++ b/helper/query_parser.js
@@ -13,9 +13,6 @@ module.exports.get_layers = function get_layers(query) {
   if (query.length <= 3 ) {
     // no address parsing required
     return get_layers_helper(['admin']);
-  } else if (tokenized.length === 1 || (tokenized.length < 3 && !hasNumber)) {
-    // no need to hit address layers if there's only one (or two) token(s)
-    return get_layers_helper(['admin', 'poi']);
   }
 };
 
--- a/test/unit/helper/query_parser.js
+++ b/test/unit/helper/query_parser.js
@@ -63,6 +63,7 @@ module.exports.tests.parse_one_or_more_tokens = function(test, common) {
   var two_tokens_nonum  = ['small town', 'biggg city', 'another empire'];
   var two_tokens_withnum= ['123 main', 'sixty 1', '123-980 house'];
 
+  // parse address is now always true to fix pelias/api#194
   var testParse = function(query, parse_address) {
     test('query with one or more tokens (' + query + ')', function(t) {
       var address = parser.get_parsed_address(query);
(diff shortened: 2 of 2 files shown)


controller/search.js
```diff
@@ -16,13 +15,9 @@ function setup( backend, query ){
...
-    if (req.clean.layers) {
-      cmd.type = req.clean.layers;
-    }
-
-    // set type if input suggests targeting a layer(s)
-    if (req.clean.default_layers_set && req.clean.parsed_input) {
-      cmd.type = req.clean.parsed_input.target_layer || cmd.type;
+    if (req.clean.type !== undefined) {
+      cmd.type = req.clean.type;
+      delete req.clean.type; // remove type from clean to avoid clutter
```
dianashk: We currently only show `req.clean` object as part of the result _geocode-json_, so if we want to show the client which layers we decided to query, it might need to be set in both places. We should still be returning `layers` array so this might not be super important.

helper/query_parser.js
```diff
@@ -63,13 +56,12 @@ module.exports = function(query) {
...
                           'regions',
-                          'admin_parts',
-                          'target_layer'
+                          'admin_parts'
                         ];
 
   var parsed_input = {};
 
-  address_parts.forEach(function(part){ 
+  address_parts.forEach(function(part){
```
dianashk: this might work with `extend` as well

helper/types.js
```diff
@@ -0,0 +1,41 @@
+var valid_types = require( '../query/types' );
```
dianashk: splitting this out into a separate file is awesome. I wonder if the filename should be a bit more descriptive, since `type` is such an overloaded term.

PR #242, opened 2015-09-10
Update /reverse endpoint params and query as per API spec

### LINKED ISSUE #239: Implement updated parameter names for /reverse endpoint
As per our design doc, update the names of the parameters (and in some cases, behaviors) on the `/reverse` endpoints

- [ ] `point`
 - `point.lon`
 - `point.lat`
- [ ] `layer`
- [ ] `source`
- [ ] `boundary.circle`
 - `boundary.circle.lon`
 - `boundary.circle.lat`
 - `boundary.circle.radius`
- [ ] boundary.country
- [ ] size
- [ ] privacy

dianashk:
* Found small bug where params were expected to have been parsed into objects, however they remain as strings and need to be processed that way. We can refactor it later to parse the params if we'd like, but it wasn't worth the effort at this time.
* Added `boundary.circle.*` processing to the reverse sanitizer and query, and tied `radius` to the distance parameter instead of always defaulting to 500km.
* Added `source` sanitizer to reverse.

Fixes #239

### DIFF
--- a/query/reverse.js
+++ b/query/reverse.js
@@ -21,13 +21,18 @@ function generateQuery( clean ){
 
   // set defaults
   vs.set({
-    'size': 1,
-    'boundary:circle:radius': '500km'
+    'size': 1
   });
 
   // set size
   if( clean.size ){
     vs.var( 'size', clean.size );
+  }
+
+  // set radius, default to 500km if not specified in request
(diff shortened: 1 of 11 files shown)


missinglink:
:+1:

PR #244, opened 2015-09-11
Distance

### LINKED ISSUE #225: Reverse geocoding should return distance from point
When a user reverse geocodes a point, one of the properties that should be returned back to the user should be the distance from the matching reverse point to the origin point of the query.

Should return in meters.

dianashk:
So turns out we don't get a nice `distance` value from elasticsearch. Decided as a simple solution to use an npm module to compute the distance for results.

Fixes #225


missinglink:
:+1:

PR #246, opened 2015-09-11
Add /status controller

### LINKED ISSUE #231: Non-versioned health check endpoint
The API should provide a request path that isn't versioned which can be used as a health check endpoint behind load balancing software/hardware (ELB, proxy, etc).

example: /status

In it's most basic form the uri should return a 200 status code for a GET request, and some body, e.g. 'status: ok'.

orangejulius:
It's at the root level, and simply returns 200 with 'status: ok' as the
body.

@heffergm, this is intended to fix #231. I'm not completely clear on how all the proxy logic between v1 and legacy work, but let me know if this works. I assume to an outside user of the pelias api, this endpoint ISN'T actually reachable, because any non /v1/ routes would get forwarded to the legacy server, which will not have this /status endpoint. However presumably that's ok.

### DIFF
--- a/controller/status.js
+++ b/controller/status.js
@@ -1,0 +1,3 @@
+module.exports = function controller( req, res, next) {
+  res.send('status: ok');
+};
--- a/routes/v1.js
+++ b/routes/v1.js
@@ -19,7 +19,8 @@ var middleware = {
 var controllers     = {
   mdToHTML: require('../controller/markdownToHtml'),
   place: require('../controller/place'),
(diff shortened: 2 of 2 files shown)


heffergm:
This works. The checks aren't run externally... the ELB connects directly to the API, so there is no proxy involved at this level. I'll update the ELBs when it's merged for the v1 endpoint.

Thanks!

missinglink:
`:+1`

PR #248, opened 2015-09-11
default private to false, Refactoring private and details parameters

### LINKED ISSUE #235: Add `private` param to all endpoints
Just provide the parameter for now, don't process it internally until (and if) we start actually storing user input anywhere.

stephenkhess:
default private to false

refactored truthy-with-default behavior out to new function in _truthy.js

Fixes #235


dianashk:
I ended up merging master into the previous commit and didn't realized there was this PR until after having refactored what was in the previous one. Now it seems like a lot of repeated effort to try to merge what's here into that. I'm really sorry! :cry:

PR #251, opened 2015-09-14
refactor sanitizers: first pass

missinglink:
first pass of a structural refactor that will allow the second stage work to be completed.

in this PR:
- deprecate `is-object` module in favour of `check-types` (and refactor type-checking code for clarity)
- refactor all the sanitizers to return an object of `messages = { errors: [], warnings: [] }`
- refactor all sanitizer method signatures to be `function( unclean, clean )`
- clean up constants and move single-execution functions to a single-execution scope
- minimal changes made to test suite in order to ensure backwards compatibility

in the next PR:
- remove implementation specific code from sanitizers, eg. the string for `unclean['point.lat']` should be defined outside the sanitizer itself, so it can be re-used and does not need to be aware of `expressjs` and things like `req.query` (almost complete).
- update response body to include a more comprehensive list of `errors` and `warnings` (at this stage it is backwards compatible and only reports the 1st error message).
- many more unit tests, much of this work was done without causing any tests to fail, this is due to most of the sanitizers not actually having any unit tests written at all!

I will branch the rest of the work from here so this can be merged while that is ongoing


missinglink:
reviewed via teleconference

PR #252, opened 2015-09-14
lodash-ify some array functions; remove 'yes' and 'y' from thruthy

missinglink:
lodash-ify some array functions; remove 'yes' and 'y' from thruthy


dianashk:
:+1:

PR #253, opened 2015-09-14
112 support boundary country parameter

### LINKED ISSUE #112: Add boundary.country parameter to /search and /reverse endpoints

stephenkhess:
Fixed #112

### DIFF
--- a/sanitiser/_boundary_country.js
+++ b/sanitiser/_boundary_country.js
@@ -8,8 +8,8 @@ function sanitize(raw, clean) {
   // init clean.boundary (if not already init)
   clean.boundary = clean.boundary || {};
 
-  if (raw.boundary && raw.boundary.country) {
-    var country = raw.boundary.country
+  if (raw['boundary.country']) {
+    var country = raw['boundary.country'];
 
     if (typeof country !== 'string') {
       messages.warnings.push('boundary.country is not a string');
--- a/sanitiser/_boundary_country.js
+++ b/sanitiser/_boundary_country.js
@@ -1,0 +1,40 @@
+var isObject = require('is-object');
+var iso3166 = require('iso3166-1');
+
+function sanitize(raw, clean) {
+  // error & warning messages
+  var messages = { errors: [], warnings: [] };
+
+  // init clean.boundary (if not already init)
+  clean.boundary = clean.boundary || {};
+
+  if (raw.boundary && raw.boundary.country) {
+    var country = raw.boundary.country
+
+    if (typeof country !== 'string') {
+      messages.warnings.push('boundary.country is not a string');
+      clean.boundary.country = undefined;
+    }
+    else if (!containsIsoCode(country.toUpperCase())) {
+      messages.warnings.push(country + ' is not a valid ISO2/ISO3 country code');
+      clean.boundary.country = undefined;
+    }
+    else {
+      clean.boundary.country = iso3166.to2(country.toUpperCase());
+    }
+
+  } else {
+    clean.boundary.country = undefined;
+  }
+
+  return messages;
+
+}
+
+function containsIsoCode(isoCode) {
+  return iso3166.list().filter(function(row) {
+      return row.alpha2 === isoCode || row.alpha3 === isoCode;
+  }).length > 0;
+}
+
+module.exports = sanitize;
--- a/sanitiser/_boundary_country.js
+++ b/sanitiser/_boundary_country.js
@@ -12,11 +12,11 @@ function sanitize(raw, clean) {
     var country = raw['boundary.country'];
 
     if (!check.string(country)) {
-      messages.warnings.push('boundary.country is not a string');
+      messages.errors.push('boundary.country is not a string');
       delete clean.boundary.country;
     }
     else if (!containsIsoCode(country.toUpperCase())) {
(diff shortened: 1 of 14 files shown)


sanitiser/_boundary_country.js
```diff
@@ -0,0 +1,40 @@
...
+  // error & warning messages
+  var messages = { errors: [], warnings: [] };
+
+  // init clean.boundary (if not already init)
+  clean.boundary = clean.boundary || {};
+
+  if (raw.boundary && raw.boundary.country) {
+    var country = raw.boundary.country
+
+    if (typeof country !== 'string') {
```
dianashk: let's use check-types here `if (check.string(country)) {}`

sanitiser/_boundary_country.js
```diff
@@ -0,0 +1,40 @@
...
+var iso3166 = require('iso3166-1');
+
+function sanitize(raw, clean) {
+  // error & warning messages
+  var messages = { errors: [], warnings: [] };
+
+  // init clean.boundary (if not already init)
+  clean.boundary = clean.boundary || {};
+
+  if (raw.boundary && raw.boundary.country) {
```
dianashk: let's use `check.assigned` here, from `check-types` module

sanitiser/_boundary_country.js
```diff
@@ -0,0 +1,40 @@
...
+    var country = raw.boundary.country
+
+    if (typeof country !== 'string') {
+      messages.warnings.push('boundary.country is not a string');
+      clean.boundary.country = undefined;
+    }
+    else if (!containsIsoCode(country.toUpperCase())) {
+      messages.warnings.push(country + ' is not a valid ISO2/ISO3 country code');
+      clean.boundary.country = undefined;
```
dianashk: I think it's cleaner to delete the property:

```javascript
delete clean.boundary.country;
```

sanitiser/_boundary_country.js
```diff
@@ -0,0 +1,40 @@
...
+    }
+    else {
+      clean.boundary.country = iso3166.to3(country.toUpperCase());
+    }
+
+  } else {
+    clean.boundary.country = undefined;
```
dianashk: also use `delete`

dianashk:
This looks fantastic! Great job!

:+1: :+1: :+1:

sanitiser/_boundary_country.js
```diff
@@ -0,0 +1,42 @@
...
+
+}
+
+function containsIsoCode(isoCode) {
+  return iso3166.list().some(function(row) {
+      return row.alpha2 === isoCode || row.alpha3 === isoCode;
+  });
+}
+
+module.exports = sanitize;
```
missinglink: you could simplify the code a little bit by avoiding doing the `delete` operations and only check `clean.boundary` in the function where you set it.

it could be rewritten as something like this:

```javascript
function sanitize(raw, clean) {
  // error & warning messages
  var messages = { errors: [], warnings: [] };

  // target input param
  var country = raw['boundary.country'];
  
  // must be valid string
  if (!check.unemptyString(country)) {
    messages.warnings.push('boundary.country is not a string');
  }

  // must be a valid ISO 3166 code
  else if (!containsIsoCode(country.toUpperCase())) {
    messages.warnings.push(country + ' is not a valid ISO2/ISO3 country code');
  }

  // valid ISO 3166 country code, set alpha3 code on 'clean.boundary.country'
  else {
    // init clean.boundary (if not already init)
    clean.boundary = clean.boundary || {};

    // the only way for boundary.country to be assigned is if input is
    //  a string and a known ISO2 or ISO3
    clean.boundary.country = iso3166.to3(country.toUpperCase());
  }

  return messages;
}
```

missinglink: should this be a fatal `error` or simply a `warning`? I guess in this case the query can still execute successfully despite an input error, but it will not produce the output the developer expected.

one one hand we could follow the idea of [fail early, fail loud](https://en.wikipedia.org/wiki/Fail-fast) and on the other we can offer [fault tolerance](https://en.wikipedia.org/wiki/Fault_tolerance).

as an engineer I would choose the former but as a user I would expect the latter?

more of a question for @pelias/owners on what constitutes a failure?

dianashk:
I think invalid parameters mean we can't return results. If we can't return results, does it make sense to still return 200 and empty geojson with errors? Or does it make sense to return an HTTP error code with a message about the error? I'm kinda leaning towards the HTTP error now. This way there is no pretence of success.

stephenkhess:
I'd be good with failing fast with a HTTP 400 and an "invalid country"-type
message.  That would solve the issue of a user entering boundary.country=UK
then wondering why results are being returned all over the planet.  Is the
code that calls the sanitizers setup to respond in this way to non-empty
error messages?

Regards,
Stephen

missinglink:
from the call today: `errors` will mean that the query is not executed and the response returned immediately, `warnings` will still allow db execution to happen but are used to warn of thiongs like deprecations and non-fatal errors.

orangejulius:
:+1: we should get this merged :)

PR #254, opened 2015-09-14
Remove geocoding block from properties in GeoJSON responses

### LINKED ISSUE #247: Remove `geocoding` namespace from responses
As described in pelias/geocodejson-spec#1, having this extra namespace makes working with lots of tools like geojson.io, a bit more difficult than we'd like.

orangejulius:
All the attributes that were in this block are now put directly in the
GeoJSON properties object

Fixes #247


dianashk:
That was easy! :+1:

PR #255, opened 2015-09-14
Add autocomplete sanitiser

### LINKED ISSUE #245: Cleanup /autocomplete sanitizers
We should not be accepting any `filter` parameters here, such as `layer` and `source`.

orangejulius:
This is similar to the search sanitiser but with a more limited scope.
It doesn't support layers, sorces, or bounding boxes.

This almost seemed too simple, am I missing anything?

Fixes #245

### DIFF
--- a/sanitiser/autocomplete.js
+++ b/sanitiser/autocomplete.js
@@ -1,0 +1,24 @@
+var sanitizeAll = require('../sanitiser/sanitizeAll'),
+    sanitizers = {
+      text: require('../sanitiser/_text'),
+      size: require('../sanitiser/_size'),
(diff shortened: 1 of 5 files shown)


sanitiser/autocomplete.js
```diff
@@ -0,0 +1,24 @@
+var sanitizeAll = require('../sanitiser/sanitizeAll'),
+    sanitizers = {
+      text: require('../sanitiser/_text'),
+      size: require('../sanitiser/_size'),
+      details: require('../sanitiser/_details'),
```
dianashk: `details` is out, `private` is in :)

dianashk:
Just swap out `details` for `private` and this is golden.

PR #256, opened 2015-09-15
fix CORS support by enabling OPTIONS endpoint

missinglink:
when updating the [compare app](https://github.com/pelias/compare) today I needed to make these changes to enable [CORS](https://en.wikipedia.org/wiki/Cross-origin_resource_sharing) support.

@heffergm this CORS thing is a bit funny in that the client issues an http `OPTIONS` request before sending the actual request, presumably in order to sniff out [these headers](https://github.com/pelias/api/blob/master/middleware/cors.js)

this is a little problematic as the route which responds to the `OPTIONS` http verb requires an `api_key` to be set otherwise the load balancer returns `301`.

so in the mean time, so long as we set an api_key on every request then it works fine, otherwise it fails silently (in the browser).

how would you feel about disabling auth for all `OPTIONS` requests? cc/ @baldur @riordan @louh as this is ditto across the board for all our services which support CORS.

### DIFF
--- a/app.js
+++ b/app.js
@@ -12,6 +12,7 @@ if( peliasConfig.accessLog ){
 
 app.use( require('./middleware/headers') );
 app.use( require('./middleware/cors') );
+app.use( require('./middleware/options') );
 app.use( require('./middleware/jsonp') );
 
 /** ----------------------- routes ----------------------- **/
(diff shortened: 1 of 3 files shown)


missinglink:
diana says :+1 :+1: :+1:

PR #265, opened 2015-09-16
set response data/meta/body on $res instead of $req

missinglink:
this PR just renames/rescopes a few variables for clarity:

- `req.results.*` are now set directly on the response object `res`
 - `req.results.data` -> `res.data`
 - `req.results.meta` -> `res.meta`
 - `req.results.geojson` -> `res.body`


dianashk:
:+1:

PR #267, opened 2015-09-16
expose error messages in geojson

missinglink:
this PR refactors the handling of request `errors` and `warnings` so that:

- there are arrays `req.errors` and `req.warnings` which any middleware can push to
- sanitizers push messages to these arrays
- controllers do not execute when a request errors occur
- we no longer `next(err)` as we want to always reply with geoJSON
- the errors and warnings are exposed to the user in the geoJSON response

note: if `req.errors.length > 1` then controllers will not execute and the db will not be queried

### DIFF
--- a/controller/place.js
+++ b/controller/place.js
@@ -1,10 +1,18 @@
 var service = { mget: require('../service/mget') };
 
 function setup( backend ){
+
   // allow overriding of dependencies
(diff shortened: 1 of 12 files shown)


dianashk:
:+1: from @orangejulius and diana

PR #268, opened 2015-09-16
Fix attribution link

### LINKED ISSUE #263: Fix attribution link in results
The link should point to `v1` and see if it's possible to bypass the api_key for that endpoint.

dianashk:
Fixes #263


orangejulius:
pretty straightforward, :+1:

PR #269, opened 2015-09-16
Remove warning when no size param specified

### LINKED ISSUE #236: Add warnings about invalid `size` param values
* `size`= 0, return 1 and provide warning about invalid value
* `size`> MAX_SIZE (currently 40), return MAX_SIZE and provide warning about invalid value

dianashk:
Originally intended to remove an unwanted warning when size param was not being specified in the query. Ended up adding tests for _size sanitizer while in there. Also looks like the warning about size being out of range were already in place.

Fixes #236

### DIFF
--- a/sanitiser/_size.js
+++ b/sanitiser/_size.js
@@ -1,3 +1,4 @@
+var check = require('check-types');
 
 var MIN_SIZE = 1,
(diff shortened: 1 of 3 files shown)


orangejulius:
:+1:

PR #271, opened 2015-09-16
Refactor to use flat structure in clean object

### LINKED ISSUE #249: Confusing error message for missing point.{lat|lon} parameters on /reverse
Queries against the /reverse endpoint now require the latitude and longitude to be specified with the "point.lat" and "point.lon" parameters, but if one of those are missing, the error is as follows:

    {
      "error": "missing param 'lat'"
    }

This is confusing since the parameter that's missing is actually point.lat

Sample query: http://pelias.mapzen.com/v1/reverse?point.lon=-73.990342&categories=food:cuisine:chinese&details=false&layers=osmnode&api_key=fillmein

### LINKED ISSUE #257: Match query parameter names returned in `geocoding` block
For example, 

`http://pelias.bigdev.mapzen.com/v1/autocomplete?text=hello&focus.point.lat=1&focus.point.lon=2`

results show `lat` `lon` instead of `focus.point.lat` and `focus.point.lon`:

```javascript
{
  "geocoding":{
    "version":"0.1",
    "attribution":"http://pelias.mapzen.com/attribution",
    "query":{
      "text":"hello",
      "parsed_text":{
        "regions":[
          "hello"
        ]
      },
      "size":10,
      "private":false,
      "lat":1,
      "lon":2,
      "categories":[

      ]
    },
    "engine":{
      "name":"Pelias",
      "author":"Mapzen",
      "version":"1.0"
    },
    "timestamp":1442333002389
  },
  "type":"FeatureCollection",
  "features":[...],
  "bbox":[
    0.71667,
    -24.865301,
    152.350793,
    68.23877
  ]
}
```

### LINKED ISSUE #262: Null island bug in reverse query
`/v1/reverse?point.lon=0.0&point.lat=0.0&boundary.circle.radius=1&size=40`

results in the following query:

```javascript
{
  "index": "pelias",
  "searchType": "dfs_query_then_fetch",
  "body": {
    "query": {
      "filtered": {
        "query": {
          "bool": {}
        }
      }
    },
    "size": 40,
    "track_scores": true,
    "sort": [
      "_score"
    ]
  }
}
```
and returns random results

orangejulius:
Both our sanitisers and query builders currently expect the clean object to have nested-sub objects such as

    var clean = {
      point: {
        lat: 40,
        lon: -73
      }
    };

However, all the API parameters will come in as flat strings, like `point.lat=40&point.lon=40`. This is easy to confuse with a nested structure since the strings look exactly like the syntax for nesting in Javascript. A couple of us have already accidentally implemented sanitisers that expect API parameters to come in as nested objects, which is of course not how it works.

So in the name of consistency, this PR refactors all the clean objects to use a consistent, flat structure. The clean object above will now look like this:

    var clean = {
      'point.lat': 40,
      'point.lon': -73
    };

## Other cleanup
A few other things were cleaned up along the way. Of the 10 commits in this PR, the first 4 deal with pure cleanup stuff, so it might be useful to view them seaparately.

In addition to un-nesting all the clean objects, some places already used a flat structure, but with underscores in the names instead of periods. These all use periods now too.

Some tests were cleaned up a bit, mostly when they were doing huge deepEquals comparisons of entire clean objects, but only cared about one single value. They just compare the value the test cares about now. As a result, the `defaultClean` fixture in `test/unit/query/search.js` is now completely gone. I'd love to get rid of the similar one in `test/unit/query/reverse.js` soon.

This also fixes #262 and other similar null island bugs that would have appeared in search, reverse, and autocomplete queries.

Finally, both the `focus.point.{lat|lon}` parameter on `/search` and the `point.{lat|lon}` parameter on `reverse` used to be crammed into `clean.lat` and `clean.lon`. They now go to properties on `clean` that reflect the API parameter name. This fixes #257. As a nice side effect, this also fixes #249.

## What's not included
This PR doesn't touch bounding box code at all. That will be separate as it will require some more major refactoring.

In general, none of the logic of structure of any sanitisers is changed. This entire PR is nearly 100% search and replace-ish changes. Look for that in a separate PR too.

### DIFF
--- a/query/autocomplete.js
+++ b/query/autocomplete.js
@@ -1,6 +1,7 @@
 
 var peliasQuery = require('pelias-query'),
-    defaults = require('./defaults');
+    defaults = require('./defaults'),
+    check = require('check-types');
 
 //------------------------------
 // autocomplete query
@@ -31,7 +32,7 @@ function generateQuery( clean ){
   vs.var( 'size', 10 );
 
   // focus point
-  if( clean['focus.point.lat'] && clean['focus.point.lon'] ){
+  if( check.number(clean['focus.point.lat']) && check.number(clean['focus.point.lon']) ){
     vs.set({
       'focus:point:lat': clean['focus.point.lat'],
       'focus:point:lon': clean['focus.point.lon']
--- a/query/reverse.js
+++ b/query/reverse.js
@@ -1,5 +1,6 @@
 var peliasQuery = require('pelias-query'),
-    defaults = require('./defaults');
+    defaults = require('./defaults'),
+    check = require('check-types');
 
 //------------------------------
 // reverse geocode query
@@ -38,7 +39,7 @@ function generateQuery( clean ){
   }
 
   // focus point centroid
-  if( clean['point.lat'] && clean['point.lon'] ){
+  if( check.number(clean['point.lat']) && check.number(clean['point.lon']) ){
     vs.set({
       // focus point to score by distance
       'focus:point:lat': clean['point.lat'],
--- a/query/search.js
+++ b/query/search.js
@@ -1,6 +1,7 @@
 var peliasQuery = require('pelias-query'),
     defaults = require('./defaults'),
-    textParser = require('./text_parser');
+    textParser = require('./text_parser'),
+    check = require('check-types');
 
 //------------------------------
 // general-purpose search query
@@ -58,7 +59,7 @@ function generateQuery( clean ){
   }
 
   // focus point
-  if( clean['focus.point.lat'] && clean['focus.point.lon'] ){
+  if( check.number(clean['focus.point.lat']) && check.number(clean['focus.point.lon']) ){
     vs.set({
       'focus:point:lat': clean['focus.point.lat'],
       'focus:point:lon': clean['focus.point.lon']
--- a/test/unit/query/autocomplete.js
+++ b/test/unit/query/autocomplete.js
@@ -37,6 +37,20 @@ module.exports.tests.query = function(test, common) {
     t.deepEqual(compiled, expected, 'valid autocomplete query');
     t.end();
   });
+
+  test('autocomplete + focus on null island', function(t) {
+    var query = generate({
+      text: 'test',
+      'focus.point.lat': 0,
+      'focus.point.lon': 0
+    });
+
+    var compiled = JSON.parse( JSON.stringify( query ) );
+    var expected = require('../fixture/autocomplete_linguistic_focus_null_island');
+
+    t.deepEqual(compiled, expected, 'valid autocomplete query');
+    t.end();
+  });
 };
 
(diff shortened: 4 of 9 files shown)


test/unit/query/autocomplete.js
```diff
@@ -27,8 +27,8 @@ module.exports.tests.query = function(test, common) {
   test('autocomplete + focus', function(t) {
     var query = generate({
       text: 'test',
-      lat: 29.49136,
-      lon: -82.50622
+      'focus.lat': 29.49136,
```
dianashk: should this be `focus.point.lat`?

orangejulius: yes, i'll fix it

On Wed, Sep 16, 2015 at 3:49 PM, Diana Shkolnikov <notifications@github.com>
wrote:

> In test/unit/query/autocomplete.js
> <https://github.com/pelias/api/pull/271#discussion_r39676270>:
>
> > @@ -27,8 +27,8 @@ module.exports.tests.query = function(test, common) {
> >    test('autocomplete + focus', function(t) {
> >      var query = generate({
> >        text: 'test',
> > -      lat: 29.49136,
> > -      lon: -82.50622
> > +      'focus.lat': 29.49136,
>
> should this be focus.point.lat?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/pelias/api/pull/271/files#r39676270>.
>



-- 
Thanks,
Julian

dianashk:
:+1:

dianashk:
:ship:

query/autocomplete.js
```diff
@@ -31,10 +31,10 @@ function generateQuery( clean ){
   vs.var( 'size', 10 );
 
   // focus point
-  if( clean.lat && clean.lon ){
+  if( clean['focus.point.lat'] && clean['focus.point.lon'] ){
```
missinglink: need to do a key check here rather than a value check; because null island

query/reverse.js
```diff
@@ -34,20 +33,20 @@ function generateQuery( clean ){
...
   var radius = 500;
-  if (clean.hasOwnProperty('boundary_circle_radius')) {
-    radius = clean.boundary_circle_radius;
+  if (clean.hasOwnProperty('boundary.circle.radius')) {
+    radius = clean['boundary.circle.radius'];
   }
 
   // focus point centroid
-  if( clean.lat && clean.lon ){
+  if( clean['point.lat'] && clean['point.lon'] ){
```
missinglink: null island

query/search.js
```diff
@@ -59,10 +58,10 @@ function generateQuery( clean ){
   }
 
   // focus point
-  if( clean.lat && clean.lon ){
+  if( clean['focus.point.lat'] && clean['focus.point.lon'] ){
```
missinglink: ditto

Issue #272, opened 2015-09-16
Error when querystring parameter specified multiple times or are not scalars


orangejulius:
Express uses the [querystring](https://nodejs.org/docs/latest-v0.12.x/api/querystring.html) module from Node.js, which eventually comes from [this](https://github.com/Gozala/querystring) github repo.

As it happens this module has a [fancy feature](https://github.com/Gozala/querystring/blob/master/decode.js) where specifying the same parameter multiple times will turn the corresponding key in `req.param` into an array containing all the specified values.

For example, an HTTP query like `GET /foo?id=1` would create a `req.params` object that looks like this:

    {
      id: "1"
    }

That's as one would expect. But a query like `GET /foo?id=1&id=2` will set `req.params` to the following:

    {
      id: ["1", "2"]
    }

This is nice, except that now all of our sanitisers have to consider that the parameters coming into them could be arrays, either unintentionally or intentionally.

I propose that we effectively ban usage of this feature by doing the following in every sanitiser, or ideally in some shared code:

1.) Warn that the parameter was specified more than once, so that unsuspecting users can fix their querystrings.
2.) Take the first specified value by doing something like

    if(_.isArray(req.params.myparam)) {
      req.params.myparam = req.params.myparam[0];
    }

On a similar note, [querystring parameters with brackets end up in nested objects under req.params](http://expressjs.com/4x/api.html#req.query). We might want to explicitly disable this as well.

PR #273, opened 2015-09-16
Rename /place `id` parameter to `ids`

### LINKED ISSUE #259: Rename `place` `id` parameter to `ids`
...to indicate that you can request multiples

orangejulius:
This is ~~mostly a very straightforward rename~~ a big refactor of the place and _ids sanitisers.

It cleans up the tests and code considerably, improves the error messages, and splits tests for just the _ids sanitiser into its own file.

It also adds a fix for the issue described in #272 (for this parameter only).

Fixes #259


sanitiser/_id.js
```diff
@@ -18,23 +18,33 @@ function sanitize( raw, clean ){
...
-
-  // 'raw.id' can be a string
-  if( check.unemptyString( raw.id ) ){
-    rawIds.push( raw.id );
+  // 'raw.ids' can be an array if ids is specified multiple times
+  // see https://github.com/pelias/api/issues/272
+  var rawIdsString;
+  if (check.array( raw.ids )) {
+    rawIdsString = raw.ids[0];
+    messages.warnings.push( '`ids` parameter specified multiple times. Using first value.' );
```
dianashk: I'd rather see this as an error since the user might not see the warning, but it has a significant impact on results.

PR #274, opened 2015-09-16
243 reverse confidence scores

### LINKED ISSUE #243: Implement confidence score for reverse endpoint

stephenkhess:
Fixes #243


orangejulius:
:+1: now that point.lat/lon are used :)

PR #275, opened 2015-09-16
Construct gids for reponses and remove id

### LINKED ISSUE #238: Inadequate data to construct request to /place endpoint with data from /search
We should normalize the way we mint our IDs (or at least display them to users) so that the `id` field of a geocode response can be used to retrieve a complete place record on the `/place` endpoint.

Steps to reproduce:
- Search for `http://pelias.bigdev.mapzen.com/v1/search?text=hackney&size=1'
 - get `id : '937569400"' as document ID. I know more than the average user, and know how we used to construct these. I get back `source : "osm"`.
- Try to retrieve document at: `http://pelias.bigdev.mapzen.com/v1/place?id=937569400`. Fails.
 - Gets non-GeocodeJSON error: ```{error: "invalid: must be of the format type:id for ex: 'geoname:4163334'"}'''
- Try to retrieve document at: `http://pelias.bigdev.mapzen.com/v1/place?id=osm:937569400`
 - Gets non-GeocodeJSON error: ```{error: "osm is invalid. It must be one of these values - [geoname, osmnode, osmway, admin0, admin1, admin2, neighborhood, locality, local_admin, osmaddress, openaddresses]"}```

orangejulius:
This is a simple change to fix #238

Mostly it creates gids as one might expect, for example with osm they
look like

    osmway:23458099
    osmnode:21269005

Or for openaddresses

    openaddresses:3028382
    openaddresses:3028381

Similarly geonames

    geoname:4930956

Quattroshapes though, is a little weird, but these gids do all work, you
can plug them into /place?ids= just fine

    locality:7693:locality:us:usa:boston

### DIFF
--- a/helper/geojsonify.js
+++ b/helper/geojsonify.js
@@ -179,13 +179,13 @@ function copyProperties( source, props, dst ) {
 }
 
 /**
- * Determine and set place id, type, and source
+ * Determine and set place gid, type, and source
  *
  * @param {object} src
  * @param {object} dst
  */
 function addMetaData(src, dst) {
-  dst.id = src._id;
+  dst.gid = src._type + ':' + src._id;
   dst.layer = lookupLayer(src);
   dst.source = lookupSource(src);
 }
--- a/test/unit/helper/geojsonify.js
+++ b/test/unit/helper/geojsonify.js
@@ -140,7 +140,7 @@ module.exports.tests.search = function(test, common) {
           ]
         },
         'properties': {
-          'id': 'id1',
+          'gid': 'type1:id1',
           'layer': 'type1',
           'source': 'type1',
           'label': '\'Round Midnight Jazz and Blues Bar, test3, Angel',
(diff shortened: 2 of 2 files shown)


helper/geojsonify.js
```diff
@@ -179,13 +179,13 @@ function copyProperties( source, props, dst ) {
...
 /**
- * Determine and set place id, type, and source
+ * Determine and set place gid, type, and source
  *
  * @param {object} src
  * @param {object} dst
  */
 function addMetaData(src, dst) {
-  dst.id = src._id;
+  dst.gid = src._type + ':' + src._id;
```
orangejulius: It's not the most complicated line of code, but it might be nice to move this to a createGid helper method. Thoughts?

missinglink: you might be able to re-use this: https://github.com/pelias/model/blob/master/Document.js#L57-L60

missinglink:
code review :+1:

PR #286, opened 2015-09-21
Remove unused test file

orangejulius:
Found this sitting around and not being called. Looks like _sources.js does what we want now.


dianashk:
:+1:

PR #289, opened 2015-09-21
272 error on duplicate parameter names

### LINKED ISSUE #272: Error when querystring parameter specified multiple times or are not scalars

stephenkhess:
fixed #272

### DIFF
--- a/sanitiser/_single_scalar_parameters.js
+++ b/sanitiser/_single_scalar_parameters.js
@@ -1,0 +1,23 @@
+
+var _ = require('lodash'),
+    check = require('check-types');
+
+// validate inputs
+function sanitize( raw, clean ){
+  // error & warning messages
+  var messages = { errors: [], warnings: [] };
+
+  Object.keys(raw).forEach(function(key) {
+    if (_.isArray(raw[key])) {
+      messages.errors.push('\'' + key + '\' parameter can only have one value');
+    } else if (_.isObject(raw[key])) {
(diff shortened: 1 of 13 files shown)


dianashk:
Look at all the shiny tests!

:+1:

PR #291, opened 2015-09-21
Remove admin matching when address is not parsed

dianashk:
Always performing this check for every query was skewing the results when the query was not for an address.

### DIFF
--- a/test/unit/fixture/autocomplete_linguistic_focus.js
+++ b/test/unit/fixture/autocomplete_linguistic_focus.js
@@ -44,7 +44,7 @@ module.exports = {
                       'lon': -82.50622
                     },
                     'offset': '1km',
-                    'scale': '100km',
+                    'scale': '50km',
                     'decay': 0.5
                   }
                 }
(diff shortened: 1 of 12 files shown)


test/unit/helper/query_parser.js
```diff
@@ -50,111 +47,64 @@ module.exports.tests.parse_three_chars_or_less = function(test, common) {
...
+  });
+  test('query with many tokens', function(t) {
+    var address = parser.get_parsed_address('main particle new york');
+    t.equal(address, null, 'nothing address specific detected');
+    t.end();
+  });
 };
 
 module.exports.tests.parse_address = function(test, common) {
-  var addresses_nonum  = [{ non_street: 'main particle', city: 'new york'},
```
orangejulius: Say farewell to these wild tests

test/unit/fixture/autocomplete_linguistic_focus.js
```diff
@@ -44,7 +44,7 @@ module.exports = {
                       'lon': -82.50622
                     },
                     'offset': '1km',
-                    'scale': '50km',
+                    'scale': '100km',
```
orangejulius: How related is this to the admin matching change?

dianashk: I don't think it is. I can remove this change from here and commit separately.

PR #293, opened 2015-09-22
Fix lost confidence score

dianashk:


orangejulius:
yay, welcome back confidence!

PR #294, opened 2015-09-22
admin boost for sqrt(popularity), autocomplete focus not restricted to phrase matches

missinglink:
there are 2x changes in this PR:

- add a boost to `admin0` `admin1` and `admin2` values based on their `sqrt(population)` (`/autocomplete` & `/search`)
- minor change to the `/autocomplete` focus subquery, it used to only score `'near'` matches if they matched **both** the `ngrams` *and* the `phrase` matching, now it only takes `ngram` in to account, in effect this means that only partially matched results will also gain focus (may have a perf hit)

on staging env now for testing


query/autocomplete.js
```diff
@@ -13,7 +13,8 @@ query.score( peliasQuery.view.ngrams, 'must' );
 
 // scoring boost
 query.score( peliasQuery.view.phrase );
-query.score( peliasQuery.view.focus );
+query.score( peliasQuery.view.focus( peliasQuery.view.ngrams ) );
+query.score( peliasQuery.view.popularity(['admin0','admin1','admin2']) );
```
orangejulius: this is actually really cool to see how this one line transforms into a whole nice big section of the query

orangejulius:
sweet! :goat:

PR #300, opened 2015-09-23
[on hold] temporary hack to remove numbers (without a suffix) eg. housenumber, …

missinglink:
**[on hold] until the release is out of the way**

temporary hack to remove numbers (without a suffix) eg. housenumber, postcode from the ngrams analysis only; while not removing the numbers from other types of analysis.

eg. https://github.com/pelias/api/blob/temp_ngrams_strip_housenumbers/test/unit/fixture/autocomplete_full_address.js#L10-L19

see notes in view file for more details:

```bash
  This is (should be!) only a temporary solution.

  It is intended to strip housenumbers from input text BUT
  should only apply to the ngrams analysis and not affect
  the other textual analysis.
  eg: 'phrase' matching should still include the housenumber

  This file can go away once the peliasOneEdgeGram and peliasTwoEdgeGram
  analysers have been modified in pelias/schema, but as would require
  a full re-index and (potentially) break backwards compatibily with the
  v0 legacy codebase it, unfortunately, has to wait until that legacy
  service has been fully decomissioned.
```

note this 'stripping' of numbers does not mutate the `$vs` object and so does not adversely effect any other part of the query or the query parsing logic.


missinglink:
staging this for review now

dianashk:
This feels like a step back in result quality, so I'd like to hold off merging this. Let's increase the timeout threshold between elaticsearch and api, however, so we don't continue seeing empty results come back with the timeout error in the geojson. That doesn't look right.

missinglink:
what exactly is the issue, could you post some examples here so we can track this issue?

the performance improvements for numeric grams is significantly better so maybe we can just tweak some of the variables to achieve a better balance between local and distant places?

missinglink:
this is on semi-permanent hold, could be deleted, let's discuss

orangejulius:
This PR is now pretty old. I'd love to hear if there are any interesting innovations here still (I can't remember the discussions anymore). If there are, let's definitely revisit them with a fresh PR. Closing for now.

PR #304, opened 2015-09-23
focus.viewport sanitiser

### LINKED ISSUE #281: Hook up sanitizer for focus.viewport
Add sanitizer so the values can actually make it into `clean` and consequently the controller.

orangejulius:
This adds the appropriate sanitizer code to `/search` for the focus.viewport parameters. Internally it simply calculates a centerpoint, and then is effectively the same as having specified that centerpoint as focus.point. Specifying both `focus.viewport` and `focus.point` returns an error.

## Bonus refactoring

* `sanitise_coord` has the same signature as the other `sanitise_*` methods:
a string key,
raw object,
clean object,
value is required boolean

* tests specifically about bounding box functionality moved from `search` sanitizer tests to `geo_common` tests. Only code specifically testing what geo parameters search takes are in the `search` tests now.

Fixes #281

### DIFF
--- a/query/search.js
+++ b/query/search.js
@@ -1,7 +1,8 @@
 var peliasQuery = require('pelias-query'),
     defaults = require('./defaults'),
     textParser = require('./text_parser'),
-    check = require('check-types');
+    check = require('check-types'),
+    geolib = require('geolib');
 
 //------------------------------
 // general-purpose search query
@@ -74,10 +75,10 @@ function generateQuery( clean ){
       check.number(clean['focus.viewport.min_lon']) &&
       check.number(clean['focus.viewport.max_lon']) ) {
     // calculate the centroid from the viewport box
-    // simply set focus:point:lat/lon, until we improve this with a radius
     vs.set({
       'focus:point:lat': clean['focus.viewport.min_lat'] + ( clean['focus.viewport.max_lat'] - clean['focus.viewport.min_lat'] ) / 2,
-      'focus:point:lon': clean['focus.viewport.min_lon'] + ( clean['focus.viewport.max_lon'] - clean['focus.viewport.min_lon'] ) / 2
+      'focus:point:lon': clean['focus.viewport.min_lon'] + ( clean['focus.viewport.max_lon'] - clean['focus.viewport.min_lon'] ) / 2,
+      'focus:scale': calculateDiagonalDistance(clean) + 'km'
     });
   }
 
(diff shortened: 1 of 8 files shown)


dianashk:
Please add the radius with this change. Without that we might as well not have the viewport feature at all.

missinglink:
:+1: very nice

missinglink:
good to go?

stephenkhess:
added focus:scale for viewport.* requests calculated from diagonal with minimum of 1

query/search.js
```diff
@@ -69,13 +70,15 @@ function generateQuery( clean ){
...
+      check.number(clean['focus.viewport.max_lat']) &&
+      check.number(clean['focus.viewport.min_lon']) &&
+      check.number(clean['focus.viewport.max_lon']) ) {
+    // calculate the centroid from the viewport box
     vs.set({
-      'focus:point:lat': vp.min_lat + ( vp.max_lat - vp.min_lat ) / 2,
-      'focus:point:lon': vp.min_lon + ( vp.max_lon - vp.min_lon ) / 2
+      'focus:point:lat': clean['focus.viewport.min_lat'] + ( clean['focus.viewport.max_lat'] - clean['focus.viewport.min_lat'] ) / 2,
+      'focus:point:lon': clean['focus.viewport.min_lon'] + ( clean['focus.viewport.max_lon'] - clean['focus.viewport.min_lon'] ) / 2,
+      'focus:scale': calculateDiagonalDistance(clean) + 'km'
```
orangejulius: Does focus:scale expect what is essentially the diameter of a circle centered on focus:point:{lat|lon}? If so that's super convenient. If not we just need to divide by two, right? @missinglink can confirm i'm sure

orangejulius:
I tested this out locally and it does seem to do a pretty good job of focusing. I'd love thoughts on acceptance tests that can exercise this functionality, but in the meantime :+1: lets merge this!

dianashk:
:+1: merge it

query/search.js
missinglink: yep but radius not diameter

sanitiser/_geo_search.js
```diff
@@ -9,10 +11,22 @@ module.exports = function sanitize( raw, clean ){
   // error & warning messages
   var messages = { errors: [], warnings: [] };
 
+  // disallow specifying both focus.point and focus.viewport
+  if ( ( raw['focus.viewport.min_lat'] ||
```
missinglink: these should be key checks using `hasOwnProperty()` as value checks will fail for zero values such as `0,0`

missinglink:
so what did we end up deciding about zoom:18+ @stephenkhess ? does this mean you are only able to find records within ~1 block of you when zoomed in close?

dianashk:
Let's get this merged and create related follow-up issues as discussed in retrospection.

@stephenkhess, please do the honors.

stephenkhess:
merged

Issue #258, opened 2015-09-15
Fatal error if only part of a geo param is provided


dianashk:
For example autocomplete allows only `focus.point.lat` to be specified and doesn't complain about missing `focus.point.lon`. This should be validated for all `focus` and `boundary` params.

orangejulius:
This looks fixed, can anyone else confirm? 

    julian@julian-mapzen ~/repos/pelias/api $  curl "http://localhost:3100/v1/autocomplete?text=foo&focus.point.lat=50"
    {
    "geocoding": {
    "version": "0.1",
    "attribution": "http:\/\/pelias.mapzen.com\/v1\/attribution",
    "query": {
      "text": "foo",
      "parsed_text": {
        
      },
      "size": 10,
      "private": false
    },
    "errors": [
      "missing point param 'focus.point' requires all of: 'lat','lon' to be present"
    ],
    "engine": {
      "name": "Pelias",
      "author": "Mapzen",
      "version": "1.0"
    },
    "timestamp": 1443043989231
     },
    "type": "FeatureCollection",
    "features": [
    
      ]
    }

missinglink:
confirmed it was fixed in https://github.com/pelias/api/commit/82ab899eccbcff1d6af7823eeee3076a29579b26#diff-0ff1e8ae717210a3c2a5e98c22935c76

PR #305, opened 2015-09-23
Parameter groups

orangejulius:
I had a cool idea a few days ago to refactor all the grouping logic into separate sanitizers. I was looking into #258 and tried my hand at it. Here's the result.

It supports optional and required groups and comes with nice error messages that are generated automatically. 

I remiplemented the geo sanitizers using the group code and it looks significantly cleaner and is much shorter. Thoughts?

##  Example usage

    var groups = require('./_groups);

    // will require all point params to be specified (could be used for reverse)
    groups.required(raw, ['point.lat'], ['point.lon']);
    // error message if not all are present: "parameters point.lat and point.lon must both be specified"

    // will require either all, or none, but not some viewport params to be specified
    groups.optional(raw, ['focus.viewport.min_lat', 'focus.viewport.max_lat', 'focus.viewport.min_lon', 'focus.viewport.max_lon']);
    // error message if only some are present: "parameters focus.viewport.min_lat, focus.viewport.max_lat, focus.viewport.min_lon and focus.viewport.max_lon must all be specified"

### DIFF
--- a/sanitiser/_geo_common.js
+++ b/sanitiser/_geo_common.js
@@ -15,12 +15,13 @@ var groups = require('./_groups'),
  * @param {bool} bbox_is_required
  */
 function sanitize_rect( key_prefix, clean, raw, bbox_is_required ) {
-
-  // the names we use to define the corners of the rect
+  // calculate full property names from the key_prefix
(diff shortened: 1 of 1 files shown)


missinglink:
@orangejulius looks really good, could we do a call today, I had some minor questions, nothing that would prevent a merge.

missinglink:
:+1:

PR #307, opened 2015-09-24
switched ES lookup field from "type" -> "_type", fixes issue with /pl…

stephenkhess:
…ace lookup on Eiffel Tower gid


orangejulius:
Excellent, thank you for fixing my bug :P :rocket:

PR #313, opened 2015-09-25
added unit tests for verifying valid mapping between sources and layers

stephenkhess:


orangejulius:
This is a good set of tests, but I wonder if we want to decouple the data from the code and test each separately. In fact a lot of these tests will already fail since I just changed the layers in #316, and it would be a pain to have to keep updating them.

It would add a bit of complexity, so I'm also ok with merging these tests (once updated for geonames) and doing it later.

stephenkhess:
What do you you propose?  I wrote the tests to help me understand the schema a bit more and that it seemed pretty good to test that inputs and translated correctly for the ES query, but I'm certainly open to decoupling.

missinglink:
hmm.. I know what you mean @orangejulius, I kinda like these tests as a 'source of truth' about our system and I guess when they fail then we can see that something which *used* to be true is not false.

that's kinda what we want from our tests, especially when working remotely because we can ping each other and say "hey, I broke this test you wrote, can you please confirm I didn't bust everything else?"

It's also pretty annoying when your source of truth changes all the time, this is the major frustration with the acceptance tests and so I agree maybe there is a better way to do these tests, the specific functionality is unnecessarily complex.

In this case they are nice and clear, easy to edit and all in one place which is really nice, it's the ones where they are all over the codebase which are a nightmare to refactor.

in the next few weeks we will be re-factoring and deleting most of this source/layer stuff anyway, so why not have these tests in place to make that refactoring easier?

orangejulius:
@missinglink made a comment somewhere, maybe on flowdoc, reminding me that all this code will go away soon. Merge it :)

PR #314, opened 2015-09-25
removed redundant is-ids-an-array check 

stephenkhess:
it's already covered by _single_scalar_parameter sanitizer


missinglink:
:+1: lookin' good

PR #316, opened 2015-09-25
Remove the geoname _type from all layers except venue

### LINKED ISSUE #315: Reverse geocode doesn't honor layers=country correctly
There's an issue with reverse geocoding and layers=country in that it's returning results in layers other than country.  Here's an example request:

http://search.mapzen.com/v1/reverse?api_key=search-Bc6zSak&point.lat=42.888663&point.lon=-0.609304&layers=country

That lat/lng is just inside the France in the France/Spain border.  Until about 2pm on 9/25/2015, it was returning 1) Andorra, 2) Spain, 3) France (then others), which is correct.  Now it's returning gn:venue results that are clearly incorrect.  It has been doing this on bigdev since at least Thursday, 9/24/2015 (when I first starting playing around with reverse).

orangejulius:
We can't distinguish between geonames of different layers due to an
ambiguity in our Elasticsearch schema that we unfortunately won't be
able to fix for a few weeks.

So, while it's technically true that there are countries, cities, etc
contained in the geonames dataset, it's still better for now to remove
geonames from these layers. We have good coverage of most coarse layers
from quattroshapes alone, so the impact isn't too bad.

Fixes #315

### DIFF
--- a/helper/type_mapping.js
+++ b/helper/type_mapping.js
@@ -48,13 +48,13 @@ var SOURCE_TO_TYPE = {
  */
 var LAYER_TO_TYPE = {
   'venue': ['geoname','osmnode','osmway'],
-  'address': ['osmaddress','openaddresses', 'geoname'],
-  'country': ['admin0', 'geoname'],
-  'region': ['admin1', 'geoname'],
-  'county': ['admin2', 'geoname'],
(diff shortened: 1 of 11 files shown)


missinglink:
:+1: yep, has to be done unfortunately

PR #318, opened 2015-09-28
Circle deploy

heffergm:
... testing required ...

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -5,7 +5,7 @@ test:
   override:
     - npm test
     - git clone git@github.com:pelias/acceptance-tests
-    - echo "{\"mapzen\":{\"api_key\":{\"pelias.mapzen.com\":\"$PELIAS_API_KEY\"}}}" >~/pelias.json
+    - echo "{\"mapzen\":{\"api_key\":{\"pelias.mapzen.com\":\"$SEARCH_API_KEY\",\"pelias.mapzen.com\":\"PELIAS_API_KEY\"}}}" >~/pelias.json
     - cd acceptance-tests && npm install && npm test -- -e prod
 deployment:
   prod_build:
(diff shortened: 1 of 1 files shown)


missinglink:
wow, wait a sec, so this will trigger a prod_build rebuild on every push?

heffergm:
No, it will deploy the API.

missinglink:
agh yes, you are right, is this what we want @pelias/owners?

this will mean every time we push (+ the unit tests pass) we will update the production API.

in cases such as below we are going to end up with a broken production environment:
- we forgot to write a test and the code now `exit(1)`'s in some situations not covered
- there is a configuration issue which needs to be tested outside the dev environment
- there is a non fatal but wide-sweeping effect on the system, eg: no records returned
- et cetera

there are also a couple more things to consider:
- what is the rollback strategy *when* the above happens?
- will this be run every time anyone clones this repo and pushes their fork to github?

heffergm:
The only thing this affects is when someone pushes to the production branch of our repo, in which case it deploys said branch in production and the build environment (although at present I have no idea if all the plumbing that went into making this feasible is actually going to work yet, but that's the working premise). This is what was requested, or such was my understanding.

Circle presently only runs 'npm test' as a default for nodejs apps. If you want it to run some other test suite, let me know.

Rollback involves manually deploying a known working commit hash, or rolling back the changes from the production branch that caused the problem and pushing, which will result in another deployment.

dianashk:
We will want it to run the acceptance tests. Eventually also the fuzzy tests, but not yet. We need to test the production branch of acceptance tests against production branch of API.

heffergm:
So as this stands now, the following will happen:
- push to any branch other than production and circle will simply run npm test on the api code and then leave well enough alone
- push to production branch, and the following happens:

```
deployment:
  prod:
    branch: production
    commands:
      - cd acceptance-tests && npm install && npm test -- -e prod
      - git clone git@github.com:mapzen/pelias-deploy.git && cd pelias-deploy && bundle install
      - cd pelias-deploy && bundle exec rake deploy:api[prod]
      - cd pelias-deploy && bundle exec rake deploy:api[prod_build]
```

Which is an ugly way of saying:
- run the acceptance tests, if they pass...
- clone pelias-deploy and bundle, and if successful...
- deploy the api to prod and then prod_build

missinglink:
this makes me a little uneasy, but sure, it gets us to continuous integration so that's a good thing.

how quickly will we get a nagios/pingdom alert if things get broken in prod?

heffergm:
Things will start complaining within a few minutes.

heffergm:
resubmitting as https://github.com/pelias/api/pull/326 to avoid the commit mess...

PR #320, opened 2015-09-29
support node v4.0

missinglink:
[do not merge] on hold till after the dust settles from the imminent v1 release.

this PR allows all the code to run on the latest **stable** version of nodejs `4.0`


orangejulius:
I'm in favor of closing this and using #423 instead, what do you think @missinglink?

missinglink:
no problem, is that the last module we needed to upgrade? :tada:

PR #322, opened 2015-09-29
remove cache control headers

missinglink:
apologies for pushing a PR at the final hour but this one needs to make it's way to production so fastly can cache for longer than 60s.


orangejulius:
yes. good idea. :cactus:

heffergm:
:+1:

PR #331, opened 2015-10-02
fix ciao tests

missinglink:
cc/ @orangejulius 

I noticed a couple oddities when testing the sanitizers groups:

> /v1/reverse?point.lat=40.744243&point.lon=-73.990342&boundary.circle.radius=999.9

this one has the sanitized inputs but with a warning that they are not supported:

```
query: {
  boundary.circle.radius: 999.9,
  boundary.circle.lat: 40.744243,
  boundary.circle.lon: -73.990342
},
warnings: [
  "boundary.circle is currently unsupported"
]
```

and this one:

> /v1/reverse?point.lat=40.744243&point.lon=-73.990342&boundary.circle.radius=foo

has an error regarding something which is not supported:

```
warnings: [
  "boundary.circle is currently unsupported"
],
errors: [
  "missing param 'boundary.circle.radius'"
],
```

also I noticed that we no longer support `boundary.circle` for `/reverse`, so I commented out some of those assertions, any idea why we chose to do that? It seems like the ideal endpoint to use that type of boundary with? cc/ @riordan 

the warning message `boundary.circle is currently unsupported` sounds like we plan to support it in the future, is that the case?

### DIFF
--- a/test/ciao/autocomplete/focus_point_missing_lat.coffee
+++ b/test/ciao/autocomplete/focus_point_missing_lat.coffee
@@ -24,7 +24,7 @@ json.features.should.be.instanceof Array
 
 #? expected errors
 should.exist json.geocoding.errors
-json.geocoding.errors.should.eql [ 'missing point param \'focus.point\' requires all of: \'lat\',\'lon\' to be present' ]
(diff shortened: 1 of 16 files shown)


orangejulius:
Aww you missed my PR #330

PR #333, opened 2015-10-02
http friendly status codes

### LINKED ISSUE #321: return non 200 error codes
at the moment we return `200` for all requests, even if errors and warnings occur. this needs to be changed to a more `HTTP` friendly solution, and that solution needs to take in to account upsteam proxies like fastly and browser agents.

missinglink:
this PR resolves #321

during the busy refactoring to get everything going through the `geocodeJSON` middleware we ended up having all requests coming back with a `200` status code.

with this PR we will support the following status codes:

- `200 OK`: self-explanatory; alles gut
- `400 Bad Request`: an input param was invalid, there is an error message explaining which one
- `404 Not Found`: this means the URI path is invalid, not that the document was not found.
- `408 Request Timeout`: elasticsearch took too long to respond. as per: `esclient.requestTimeout`
- `500 Internal Server Error`: generic fatal error
- `502 Bad Gateway`: the api lost connection to the elasticsearch cluster

... the only new code from version `v0` is `502 Bad Gateway`

cc/ @heffergm 

I can write up some docs for this once it's merged, plus we need to update the CORS docs


orangejulius:
excellent

dianashk:
:+1:

PR #336, opened 2015-10-07
remove generated ciao docs

missinglink:
remove all the markdown docs created by ciao. these seem to have been committed back in December, and they are super out-of-date.

the reason I want to remove them is that they show up in Search Engine results for queries such as 'pelias cors' which is bad.

no code in this PR, just markdown files, the whole PR is a result of one command:
 ```
pelias/api$ rm -rf docs/
```


orangejulius:
:+1:

PR #340, opened 2015-10-08
more label improvements

missinglink:
some more label improvements

see the included test cases for more details


missinglink:
simple stuff, I take it there are no objections here?

PR #345, opened 2015-10-12
Use tap-dot

### LINKED ISSUE #343: Display only unit test failures
Currently when we run our unit tests, there is output for all (currently) 840 tests. This means when a few tests are failing, one normally has to scroll super far up to view the failure. Alternatively some of us comment out lines in the `test/unit/run.js` file, which isn't ideal either.

Many test suites only display output for failing tests, and there are quite a few ways to do that with tape/TAP. There's a list of TAP parsers that do this. Unfortunately [tap-spec](https://github.com/scottcorgan/tap-spec), the one we're using, doesn't.

Looking at the [list](https://github.com/substack/tape#pretty-reporters) on the tape readme, the one I like the most so far is [tap-dot](https://github.com/scottcorgan/tap-dot).

orangejulius:
[tap-dot](https://www.npmjs.com/package/tap-dot) is a TAP(test anything protocol) formatter. It's quite a bit less verbose than what we're using now, [tap-spec](https://github.com/scottcorgan/tap-spec). It only includes dots for test output, with the exception of failures, where it shows at least as good of a diff of what was seen vs expected as tap-spec.

Please try out this branch, run the tests, and make a few fail, to see how it feels. One thing I noticed is we may want to turn off logging output when running tests. Does anyone know how to do that?

This means we don't have to scroll up endlessly or comment out tests to see just failures, and closes #343

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   "scripts": {
     "start": "node index.js",
     "test": "npm run unit",
-    "unit": "node test/unit/run.js | tap-spec",
+    "unit": "node test/unit/run.js | tap-dot",
     "ciao": "node node_modules/ciao/bin/ciao -c test/ciao.json test/ciao",
     "coverage": "node_modules/.bin/istanbul cover test/unit/run.js",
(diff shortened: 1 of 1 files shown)


dianashk:
:+1: :+1:

PR #347, opened 2015-10-13
Finish Renaming address_parser to text_parser

### LINKED ISSUE #278: 'from_address_parsing' block present when address parsing not relevant
in the query below the `geocoding.query.types['from_address_parsing']` block probably shouldn't be there for an input such as `a`?
http://localhost:3100/v1/search?text=a&layers=coarse

```javascript
{
  "geocoding": {
    "version": "0.1",
    "attribution": "http:\/\/pelias.mapzen.com\/v1\/attribution",
    "query": {
      "text": "a",
      "parsed_text": {
        
      },
      "types": {
        "from_address_parsing": [
          "admin0",
          "admin1",
          "admin2",
          "neighborhood",
          "locality",
          "local_admin"
        ],
        "from_layers": [
          "admin0",
          "admin1",
          "admin2",
          "neighborhood",
          "locality",
          "local_admin"
        ]
      },
      "size": 10,
      "private": false,
      "boundary": {
        
      },
      "type": [
        "admin0",
        "admin1",
        "admin2",
        "neighborhood",
        "locality",
        "local_admin"
      ]
    },
    "engine": {
      "name": "Pelias",
      "author": "Mapzen",
      "version": "1.0"
    },
    "timestamp": 1442565412941
  },
```

orangejulius:
There were some lingering files named query_parser, as well as the `from_address_parser` key in the clean.types object.

Incidentally, there was also a bug in this code that was causing the text parser's preferences for which Elasticsearch types to search to not be respected.

The intent of the code in `helper/text_parser.js`, line 11, is to avoid searching everything except addresses for very short input texts (3 characters or less). Now that this is actually working, we'll have to decide how we like it.

Fixes #278


dianashk:
Cool :+1:

PR #351, opened 2015-10-20
german street/number naming convention

missinglink:
resolves https://github.com/pelias/pelias/issues/141

this PR simply flips the housenumber and street name for german addresses.

eg. '101 Grolmanstraße' -> 'Grolmanstraße 101'

### DIFF
--- a/middleware/localNamingConventions.js
+++ b/middleware/localNamingConventions.js
@@ -1,0 +1,39 @@
+
+function setup() {
+  return applyLocalNamingConventions;
+}
+
+function applyLocalNamingConventions(req, res, next) {
+
+  // do nothing if no result data set
+  if (!res || !res.data) {
+    return next();
+  }
+
+  // loop through data items and flip relevant number/street
+  res.data.filter(function(place){
+    // only relevant for German addresses
+    if( 'DEU' !== place.alpha3 ){ return false; }
+    if( !place.hasOwnProperty('address') ){ return false; }
+    if( !place.address.hasOwnProperty('number') ){ return false; }
+    if( !place.address.hasOwnProperty('street') ){ return false; }
+    return true;
+  })
+  .forEach( flipNumberAndStreet );
+
+  next();
+}
+
+// DE address should have the housenumber and street name flipped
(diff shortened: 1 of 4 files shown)


middleware/localNamingConventions.js
```diff
@@ -0,0 +1,39 @@
...
+
+  // do nothing if no result data set
+  if (!res || !res.data) {
+    return next();
+  }
+
+  // loop through data items and flip relevant number/street
+  res.data.filter(function(place){
+    // only relevant for German addresses
+    if( 'DEU' !== place.alpha3 ){ return false; }
```
dianashk: I think this should be more config-driven. `DEU` and filter names should be part of a config file where you specify which get which analyzers. For example:

```javascript
{
  "DEU": ["flipNumberAndStreet"]
}
```

then in code something like:

```javascript
if (check.isSet(config, place.alpha3)) {
  filters[config[place.alpha3](place);
}
```

missinglink: fair point, would you prefer to block this release to future-proof the functionality or be more 'lean' and wait to do it when required? ;)

dianashk: Team consensus is to move forward and merge this as is.

dianashk:
:+1:

middleware/localNamingConventions.js
orangejulius: Yup, maybe the next time someone adds a country, or maybe even the time after that, this will need some sort of config. Until then this code nicely fixes an issue for our friends in Germany :)

riordan:
:de: :+1:

hannesj:
Wouldn't it be better to swap the number and name already when generating the data as suggested in pelias/openstreetmap#68, rather than doing it on the fly during the queries? Would be much cleaner at least in my mind.

missinglink:
merging this as an interim solution, as per @hannesj comment it would be better to do it at indexing time rather than query time.

doing like this means we can get it out quicker, we are also having a discussion about not creating a `name` field for addresses, instead just querying on the street name and number as separate fields.

https://github.com/pelias/openaddresses/pull/24

PR #358, opened 2015-11-02
remove ngram analysis from search endpoint

missinglink:
DO NOT MERGE

remove ngram analysis from search endpoint
the start of good things to come ;)
requires dev testing

[EDIT] did a little testing and having second thoughts on this one, see:
- http://pelias.github.io/compare/#/v1/search%3Ftext=whole%20foods%20NY
- http://pelias.github.io/compare/#/v1/search%3Ftext=new%20zealand

it would probably be better to use a `shingles` analysis for the `must` condition rather than `phrase`


orangejulius:
I think we kinda just merged something similar with #538, if not let's discuss!

PR #368, opened 2015-11-06
Update addressit module to not strip 0s from zipcodes

dianashk:
Fixes pelias/pelias#166

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "elasticsearch": ">=1.2.1"
   },
   "dependencies": {
-    "addressit": "1.3.0",
+    "addressit": "git://github.com/dianashk/addressit.git@9da3f04128c22b28710a993a4cedd0ad4f5aa4a3",
     "async": "^0.9.0",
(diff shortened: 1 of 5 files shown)


orangejulius:
Strings for everything!

https://twitter.com/velartrill/status/615714789499637760

:+1: 

One note though: it looks like NPM does not have a way to tell the difference between copies of the same module with the same version. So after switching to this branch to test, running `npm install`, npm had your version cached. When I was done, ran `git checkout master` and then `npm install` again (even after a `rm -rf node_modules`), the modified version of addressit was still cached. This caused the tests on master to fail since they were still expecting housenumber and zip to be numbers. So it might make sense to change the version string in your branch of addressit. Rule [10](http://semver.org/) of semver says you can add metadata to a version that doesn't affect its precedence with other versions after a `+` so perhaps the version should be changed to something like `1.3.0+pelias`

dianashk:
Good catch! Thanks. Will update.

PR #369, opened 2015-11-09
Fix postalcode confidence score

orangejulius:
I noticed when testing #368 that confidence scores were being capped at 0.5, and not matching on zipcode. Looks like we changed the mapping at some point but forgot to update this.

Note: I couldn't find any documents that didn't at least have the address property defined as an empty object, even with an Elasticsearch query like the one below, so I didn't add an explicit check that address is set.

```
GET /pelias/_search
{
  "query": {
    "filtered": {
      "filter": {
        "bool": {
          "must_not": [
            { 
              "exists": {
                "field": "address"
              }
            }
          ]
        }
      }
    }
  }
}
```

### DIFF
--- a/middleware/confidenceScore.js
+++ b/middleware/confidenceScore.js
@@ -100,7 +100,7 @@ function checkForDealBreakers(req, hit) {
     return true;
   }
 
-  if (check.assigned(req.clean.parsed_text.postalcode) && req.clean.parsed_text.postalcode !== hit.zip) {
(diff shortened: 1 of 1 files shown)


dianashk:
:+1: 

Is there a way to cover this in an acceptance test? Just something that is an address and has a score of `1` might work.

orangejulius:
Yes, it's possible! Sortof. We can treat a confidence score of 0.5 as an unexpected property.

PR #378, opened 2015-12-07
autocomplete overhaul

missinglink:
**Note:** this is the twin ticket of https://github.com/pelias/schema/pull/84 both should be deployed at the same time.

This PR heavily refactors the `/autocomplete` functionality to fix some usability issues. It brings the product closer to the 'feeling' of the big-G maps.

A demo is available here: http://missinglink.embed.s3.amazonaws.com/pelias-dev.html

There are quite a lot of changes in this PR and some are quite hard to explain, I've done my best to go back through the code and comment every line which I found confusing on second read.

If there are parts which are still not clear please add a comment below and I'll try to make it clearer.

PR https://github.com/pelias/schema/pull/84 introduced some improved *analysis* techniques for correcting 'jitter' by rethinking the way we deal with synonyms and the way we handle stop words.

Following on from that, this PR deals with the *logical* changes to how we build the query:
1) minimizing *jitter* (seeing a label enter and leave the list as you type)
2) improved *local* focus (more emphasis on local places)
3) no degradation in finding *distant* places (a balancing act with 2.)
4) the ability to localize by specifying a region after the comma (such as "pizza, paris")
5) input parsing (tokens after the comma are not considered part of the name) see https://github.com/pelias/api/pull/382
6) requiring fewer characters to be entered before finding your result.
7) reducing the amount of results returned (no 'long tail' in the results list).
8) performance improvements (using full token matching instead of ngrams for whole words)

there are some other nice subtle effects but I won't go in to too much detail.

#### 1. jitter

The PR notes on https://github.com/pelias/schema/pull/83 do a good job of explaining what's going wrong there.

In this PR we introduce the idea of 'splitting' the `input:name` in to the *last token* and the *preceding tokens*, so for the input "40 west 26th st" we only send "st" to the `ngram` index, the remaining tokens "40 west 26th" are sent to the `phrase` index.

This drastically reduces the 'jitter' as once you've typed a token and have started on your next token, the first token is effectively 'locked-in' and so prefixes of that token are no longer considered. 

#### 2. local focus

the decay scale has been expanded from `offset 1km, scale: 50km` to `offset 10km, scale: 250km` which makes a huge difference for drivers searching for their destination.

the weight has also been increased from `2` to `3` which gives it more emphasis than previous builds.

local focus is now only applied to [certain 'layers'](https://github.com/pelias/api/pull/378/files#diff-70bb685035d312ef458e2a6d4daac6b6R25)

#### 3. distant places

care has been taken to ensure that distant places are still discoverable. the risk with setting 2. values too high is that local venues such as "london drycleaners" etc. would drown out 'London, UK'.

based on my 'feels' I like this balance, it allows you to still find London from New York in the top 10 results but when searching 'London' from *inside* London you find that the region is not visible and you instead get local venues.

I am able to find smaller areas in New Zealand when focussed on Berlin, albeit by specifying more characters than when I am focussed on NZ. This 'feels' like the right behavior.

#### 4. localize with a comma

to make things even more complex, we also have to account for users who fully-specify a region using a comma to delimit the 'admin' components such as "pizza, paris"

in this case we are not only taking in to account 2. and 3. but also trying to skew the results towards Paris.

the 'admin' weights are set extremely high in order to 'force' the results to that locality. the weighting seems to be sufficient to overcome the effects of 2. although the user will experience a mix of both local results and Paris results.

#### 5. input parsing

as above, this is now enabled. it requires use of a comma to delimit segments of the query.

we can improve this further with https://github.com/openvenues/node-postal

#### 6. requiring fewer characters

due to the effects of the above, you now have to specify fewer characters to find what you want.

in the past 'didi dumpling' required you to type almost the whole name in order to find it, now you only need type 'didi' when in Manhattan to find it.

#### 7. reducing the amount of results returned

in the past the results have been quite 'busy', due to the behavior outlined in 1. combined with using 'type:phrase' analysis and 'operator:and' we are much stricter in removing lower quality matches.

specifying `type:phrase` means that the tokens must appear in the same *order* in both the input and the index. a slop of 2 is allowed, this is sufficient to counter for omitting a single token or swapping the position of 2 tokens.

Eg. 
```
'korea south' -> 'South Korea'
'victoria university wellington' -> 'Victoria University of Wellington'
'la bouche' -> 'l'eau a la bouche'
```

specifying `operator:and` means that *all* the tokens provided by the user **must** be matched in the results. this behavior is really nice, if you search 'samsung acc' then you will now only get one result.

the downside of 'operator:and' is that when you type an address where the streetnumber is not known you will receive 0 results rather than a list of partial matches (as we did in the past)

Eg. ```10000 26th street```

This issue is better solved using street range interpolation techniques than returning fuzzy matches.

#### 8. performance improvements

yet to be measured properly, reducing the amount of ngram matches, using 'operator:and' and having (sometimes) 2 clauses in the 'MUST' section will mean faster execution times and (hopefully) more requests per second on the same hardware. 

happy holidays!


orangejulius:
After looking at this today I'd be curious to see how each of the sub-changes in this PR affect things on their own. We can pair on this sometime hopefully.

riordan:
Huge progress!

1__tmux__tmux_


It's actually "sticking" now too!

One area where I've noticed the "stickiness" isn't sticking is on street names ending in `s`

(apologies for small screenshots)

1__tmux__tmux_

1__tmux__tmux_

1__tmux__tmux_

1__tmux__tmux_

```5505 Stevens Way, San Diego, CA, 92114
FFFFFFF8111FFF000000000000000000000000```

```1050 North Hills Boulevard, Reno, NV, 89506
FFFFFFFFFFFFF00FFF0000000000000000000000000```

We'll dig into it more next week and isn't a reason to stop this from rolling to production.

**Comment by orangejulius** [comment]
I looked at some of the queries for the remaining jumpiness that @riordan mentioned and i think it's because we're using the standard analyzer in the query, which does not remove pluralization, but the peliasPhrase analyser at index time which does. so there are never any matchs when you type the final 's'. We should be able to fix this easily as I think the standard analyser was just an experiment and anything positive about using over peliasPhrase (like avoiding most stemming), can be built into peliasPhrase

**Comment by missinglink** [comment]
There seems to be a jitter/failure case when using directionals such as "north" or "southeast", it might be better to mark it as a regression and fix it in a subsequent release.

Eg. '30 west'. @orangejulius this happened when I changed to using `peliasPhrase` instead of `standard`, I don't really want to get in to refactoring `peliasPhrase` as this time because it weill effect the `/search` endpoint.

File: query/autocomplete.js [hunk 1]
```diff
@@ -39,8 +51,29 @@ function generateQuery( clean ){
...
+  // if the input parser has run and suggested a 'parsed_text.name' to use.
+  if( clean.hasOwnProperty('parsed_text') && clean.parsed_text.hasOwnProperty('name') ){
+
+    // mark the name as complete (user has already typed a comma)
+    vs.var( 'input:name:isComplete', true );
+
+    // use 'parsed_text.name' instead of 'clean.text'.
+    text = clean.parsed_text.name;
+  }
+
```
orangejulius: much cleaner now :)

hkrishna:
:heart_eyes:

riordan:
That's about how I feel

orangejulius:
@missinglink, I can just barely see the jitter in addresses with directionals, so i think if there are any major problems we can work it out later. I ran another round of the autocomplete acceptance tests against dev and they are slightly better than before the cleaup, so I say it's time to merge all this stuff.


Here's a screenshot. The previous dev score was 958 pass, 22 jumpy.
screenshot from 2015-12-28 17 11 16

riordan:
Honestly these PR's are the best holiday gift of all.

PR #390, opened 2015-12-17
Fix deduping for non-ascii strings

dianashk:
:no_mouth: my bad. @missinglink thanks for keeping me honest.


orangejulius:
:+1: looks good. I couldn't find any real world examples where this would happen, but the unit test looks perfect.

Issue #395, opened 2016-01-11
npm run docs fail


armadillabs:
Hi there:

I am trying to generate the API docs, and errors when trying to run `npm run docs`.

Here is the stack trace and logs:

```
> pelias-api@2.2.0 docs /Users/pma/Palantir/repo/garnet/pelias-api
> rm -r docs; cd test/ciao; node ../../node_modules/ciao/bin/ciao -c ../ciao.json . -d ../../docs

rm: docs: No such file or directory
 /Users/pma/Palantir/repo/garnet/pelias-api/node_modules/ciao/bin/ciao:42
  if( err ) throw new Error( err );
            ^

Error: Unexpected end of input
  at ciao (/Users/pma/Palantir/repo/garnet/pelias-api/node_modules/ciao/bin/ciao:42:19)
  at RequestChain.done (/Users/pma/Palantir/repo/garnet/pelias-api/node_modules/ciao/lib/Script.coffee:14:12)
  at /Users/pma/Palantir/repo/garnet/pelias-api/node_modules/ciao/lib/RequestChain.coffee:81:17
  at /Users/pma/Palantir/repo/garnet/pelias-api/node_modules/ciao/node_modules/async/lib/async.js:726:13
  at /Users/pma/Palantir/repo/garnet/pelias-api/node_modules/ciao/node_modules/async/lib/async.js:52:16
  at done (/Users/pma/Palantir/repo/garnet/pelias-api/node_modules/ciao/node_modules/async/lib/async.js:241:17)
  at /Users/pma/Palantir/repo/garnet/pelias-api/node_modules/ciao/node_modules/async/lib/async.js:44:16
  at /Users/pma/Palantir/repo/garnet/pelias-api/node_modules/ciao/node_modules/async/lib/async.js:723:17
  at /Users/pma/Palantir/repo/garnet/pelias-api/node_modules/ciao/node_modules/async/lib/async.js:167:37
  at Process.<anonymous> (/Users/pma/Palantir/repo/garnet/pelias-api/node_modules/ciao/lib/RequestChain.coffee:72:18)
  at emitMany (events.js:108:13)
  at Process.emit (events.js:182:7)
  at ChildProcess.<anonymous> (/Users/pma/Palantir/repo/garnet/pelias-api/node_modules/ciao/lib/Process.coffee:32:8)
  at emitTwo (events.js:87:13)
  at ChildProcess.emit (events.js:172:7)
  at maybeClose (internal/child_process.js:821:16)
[...]

PR #403, opened 2016-02-03
Make doc generation a bit easier

orangejulius:
While looking into #395, I discovered it's a bit hard to generate the pelias docs. This PR...

* Fixes some Ciao tests that were failing because of overspecified expectations
* Moves the commands to generate docs into its own file since it was already quite long
* Fixes a case where trying to generate the docs will fail if the docs aren't already there (oh the irony)
* Checks if Pelias is running before starting doc generation, and reminds you to start it if not

### DIFF
--- a/bin/generate-docs
+++ b/bin/generate-docs
@@ -2,5 +2,8 @@
 
 rm -r docs || true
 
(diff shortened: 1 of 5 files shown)


dianashk:
:+1:

trescube:
:+1: Good call on extracting all that out instead of jamming into package.json

PR #404, opened 2016-02-08
Handle geonames gids

orangejulius:
Instead of checking if the source/layer combination for geonames is
valid, always use the geoname type if the source is set to geonames,
regardless of which layer is in the gid.

This helps us better handle the case where people directly take gids
from our API responses.

Note: this functionality shouldn't be permanent and should be removed
once our new source/layer mapping system is in.

Fixes pelias/pelias#226
Fixes pelias/pelias#176


dianashk:
:ship:

PR #406, opened 2016-02-09
Update TravisCI config

orangejulius:
Adds Node 4.x versions, and containerized builds (faster).

Connects pelias/pelias#258


dianashk:
:100:

PR #415, opened 2016-02-23
Use patched branch of tap-dot

orangejulius:
Our test printer has a bug where a zero-test suite will pass. Usually,
and definitely for this repo, the only way a test suite will not have
any tests is if there was an error in the test program before it even
ran any tests ([example](https://travis-ci.org/pelias/api/jobs/110970031)).

Until [my PR](https://github.com/scottcorgan/tap-dot/pull/10) or something similar is merged into tap-dot, we should use
this branch.

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "nsp": "^0.3.0",
     "precommit-hook": "^1.0.7",
     "proxyquire": "^1.4.0",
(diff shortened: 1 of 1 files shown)


trescube:
After hours of thoughtful contemplation of this PR:  :+1:

PR #417, opened 2016-02-24
Debug log full queries and responses between Pelias and Elasticsearch

vesameskanen:
Ability to see full query/response content is very useful when debugging. Logging all queries in Elasticsearch seems to be difficult.

### DIFF
--- a/controller/search.js
+++ b/controller/search.js
@@ -30,6 +30,8 @@ function setup( backend, query ){
       cmd.type = req.clean.type;
     }
 
+    logger.debug( '[ES req]', JSON.stringify(cmd) );
+
     // query backend
     service.search( backend, cmd, function( err, docs, meta ){
 
@@ -42,7 +44,7 @@ function setup( backend, query ){
         res.data = docs;
         res.meta = meta;
       }
-
+      logger.debug('[ES response]', JSON.stringify(docs));
       next();
     });
 


orangejulius:
Hey @vesameskanen,
This is helpful, thanks! We also have `autocomplete`, `place`, and `reverse` endpoints with their own queries so, if those are interesting to you too, we could print out the queries in more places.

vesameskanen:
I added logging to controller/search, which should cover also autocomplete and reverse endpoints. A new pull request for the place endpoint logging is under way.

orangejulius:
Ah, you're right! Truth be told, our code might be a little _too_ well abstracted there.

vesameskanen:
Well abstracted indeed - I am right now separating configuration from code inside pelias api, and that adds one more layer of complexity on top of the abstraction. The purpose is that the query structure can be defined in the pelias config file.

missinglink:
Sorry, a little late to this conversation. The [elasticsearch client](https://www.elastic.co/guide/en/elasticsearch/client/javascript-api/current/configuration.html) has a configurable `log` property which can be used to control the verbosity of logging between the client and server.

These settings are provided by `~/pelias.json` under the property named `esclient`, that whole block is passed to the elasticsearch client when a connection is opened.

orangejulius:
@missinglink, man this is a bummer: the debug mode of output for esclient does output the query...almost

```
Elasticsearch DEBUG: 2016-03-28T22:11:01Z
  starting request { method: 'POST',
    path: '/pelias/neighbourhood/_search',
    body: 
     { query: { filtered: [Object] },
       size: { [Number: 20] '$': 20 },
       track_scores: { [Boolean: true] '$': true },
       sort: [ '_score' ] },
    query: { search_type: 'dfs_query_then_fetch' } }
```

PR #418, opened 2016-02-24
fix focus weighting

missinglink:
**note:** for all the screenshots in the PR you'll find the current **production server on the left**, the current **dev build in the centre** and **my local machine on the right**.

**note:** the `focus.point` for all the screenshots is a point in Manhattan, NYC.

#### issue

This PR fixes a regression in the 'focus bias' applied to autocomplete queries, you can see what I mean in the screenshot below:

mcdonalds

In the above screenshot, when you compare the left and the centre maps you'll see a big difference, on the live server you can see results clustering towards the `focus.point` whereas on the dev server they don't.

#### query changes

The fix for that was pretty simple:

```diff
query/view/focus_selected_layers.js
 @@ -22,10 +22,8 @@ module.exports = function( subview ){
      if( view && view.hasOwnProperty('function_score') ){
        view.function_score.filter = {
          'or': [
 -          { 'type': { 'value': 'osmnode' } },
 -          { 'type': { 'value': 'osmway' } },
 -          { 'type': { 'value': 'osmaddress' } },
 -          { 'type': { 'value': 'openaddresses' } }
 +          { 'term': { 'layer': 'venue' } },
 +          { 'term': { 'layer': 'address' } }
          ]
        };
      }
```

... in the past we were using the '_type' to pick which documents got a focus boost, now we are using the 'layer' value, so big win there, we don't have to maintain a list of magic variables in our code and it will 'just work' with new datasets.

Changing the above 'mostly' fixed the issue but the weightings seem to no longer be strong enough to ensure the delicate local-over-regional balance we had before, eg:

balance

the focus is too strong in this case as we would like 'London, UK' in the mix.

#### weighting changes

So I've changed some of the default values, I upped the `focus.weight` from 3 to 10 and then I had to also up the `population.weight` from 2 to 3 in order to get a similar 'feeling' balance to what we had before.

london

you can see in the image above (map on the right) that *most* of the results are now around the focus point *but* we also find the UK city; winning.

#### tighter clustering

I also changed `focus:offset` from `10km` to `0km`, I'm not sure why we had it like this in the first place, likely a mistake. What this was doing is saying "everything within 10km of the focus point get's the same boost", which resulted in Brooklyn venues getting returned when there was a closer match in Manhattan.

This should actually improve local searches significantly.

@riordan let's have a chat about this, it could end up making it hard for long-distance drivers to find their destination although it shouldn't have too much effect as I didn't touch `focus.scale`? I'm sure it would make taxi drivers happier.

donut

26_street

you can see that the grouping is now 'tighter' around the focus point:

grouping1

grouping2

This will have to go up on to the dev server as we're never really sure about the side effects of changing the weights until we've run the full regression test suite against it.

### DIFF
--- a/query/autocomplete_defaults.js
+++ b/query/autocomplete_defaults.js
@@ -30,10 +30,10 @@ module.exports = _.merge({}, peliasQuery.defaults, {
   'phrase:slop': 2,
 
   'focus:function': 'linear',
-  'focus:offset': '10km',
+  'focus:offset': '0km',
   'focus:scale': '250km',
   'focus:decay': 0.5,
-  'focus:weight': 3,
+  'focus:weight': 10,
 
   'function_score:score_mode': 'avg',
   'function_score:boost_mode': 'multiply',
@@ -90,6 +90,6 @@ module.exports = _.merge({}, peliasQuery.defaults, {
   'population:field': 'population',
   'population:modifier': 'log1p',
   'population:max_boost': 20,
-  'population:weight': 2
+  'population:weight': 3
 
 });
--- a/query/view/focus_selected_layers.js
+++ b/query/view/focus_selected_layers.js
@@ -22,10 +22,8 @@ module.exports = function( subview ){
     if( view && view.hasOwnProperty('function_score') ){
       view.function_score.filter = {
         'or': [
-          { 'type': { 'value': 'osmnode' } },
-          { 'type': { 'value': 'osmway' } },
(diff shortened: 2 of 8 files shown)


dianashk:
Code looks good. Any idea why circle is failing?

missinglink:
unfortunately not, see https://github.com/pelias/api/issues/419

orangejulius:
The weighting changes make sense, lets put this on dev today and test it out!

PR #420, opened 2016-02-24
Upgrade to precommit-hook 3.0.0

orangejulius:
It handles keeping track of the tasks better (you can't do `npm run lint` with precommit-hook 1.x


trescube:
:+1:

PR #421, opened 2016-02-25
Remove out of date EXAMPLES.md file

orangejulius:
I just notice this file which is extremely out of date. It's a nice idea, but is probably much more comprehensively covered by our documentation.


dianashk:
:+1:

PR #422, opened 2016-02-25
Add query/response logging in debug mode to place controller, too

vesameskanen:


orangejulius:
Nice, thanks for the second PR!

PR #423, opened 2016-02-25
Use the official elasticsearch module

orangejulius:
This replaces the pelias-esclient library which has not been updated much in a while.

It bring us closer to both Node 4/5 and Elasticsearch 2 support


dianashk:
team says :sheep:-it

PR #430, opened 2016-03-01
Add iso3 to usa results

trescube:
Fixed pelias/pelias#216

This PR modifies label generation by not seeding the label with the `default.name` field when the source is geonames or whosonfirst and layer is region.  

Editorially, this code is confusing because it relies on a lot of nuance about what we're doing and what constitutes results at various layers.  That is, the last step is to dedupe the array of label parts, and knowledge that (for example) a locality `default.name` and `locality` fields are duplicates is implied.  In this case, `default.name` is only meaningful for venues and addresses but it's always included then subtly removed.  I'm adding comments to clarify, but it should be refactored to be less clever.


helper/labelGenerator.js
```diff
@@ -28,8 +23,45 @@ module.exports = function( record ){
...
+  //  `Lancaster, Lancaster, PA, USA`
+  //
+  //  then the `unique` call reduces this to:
+  //  `Lancaster, PA, USA`
+  //
+  //  this code doesn't have the same effect in the case of a venue or address
+  //  where the `name` field would contain the address or name of a point-of-interest
+  //
+  //  Also see https://github.com/pelias/api/issues/429 for other ways that this is bad
+  return _.unique( labelParts ).join(', ').trim();
```
orangejulius: Lodash 4 something gets rid of _.unique in favor of _.uniq. Might as well update now.

test/unit/helper/labelGenerator_SWE.js
```diff
@@ -0,0 +1,53 @@
...
+// SWE city
+module.exports.tests.skane2 = function(test, common) {
+  test('skåne 2', function(t) {
+    var doc = {
+      'name': { 'default': 'Malmö' },
+      'country_a': 'SWE',
+      'country': 'Sweden',
+      'region': 'Skåne',
+      'county': 'Malmö',
+      'locality': 'MalmÃ¶'
```
orangejulius: is this encoding error intentional?

test/unit/helper/labelGenerator_USA.js
```diff
@@ -0,0 +1,230 @@
...
+    var doc = {
+      'name': { 'default': 'California' },
+      'country_a': 'USA',
+      'country': 'United States',
+      'region': 'Region Name',
+      'region_a': 'Region Abbr',
+      'source': 'whosonfirst',
+      'layer': 'region'
+    };
+    t.equal(generator(doc),'Region Abbr, USA');
```
orangejulius: wait so the label for the state of california should be "CA, USA"?

test/unit/helper/labelGenerator_USA.js
```diff
@@ -0,0 +1,230 @@
...
+    t.end();
+  });
+};
+
+// USA non-geonames/whosonfirst state
+module.exports.tests.region_other_source = function(test, common) {
+  test('default name should not be prepended when source=whosonfirst and layer=region', function(t) {
```
orangejulius: i think this test description doesn't match the test

test/unit/helper/labelGenerator_SWE.js
trescube: It came from master.  I'll fix in the branch though.

test/unit/helper/labelGenerator_USA.js
```diff
@@ -0,0 +1,229 @@
+var generator = require('../../../helper/labelGenerator');
```
orangejulius: these are all awesome tests

test/unit/helper/labelSchema.js
```diff
@@ -12,36 +12,766 @@ module.exports.tests.interface = function(test, common) {
...
-  };
-
-  for (var keys in schemas) {
-    isValid(keys, schemas[keys]);
-  }
+module.exports.tests.supported_countries = function(test, common) {
+  test('support countries', function(t) {
+    var supported_countries = Object.keys(schemas);
+
+    t.notEquals(supported_countries.indexOf('USA'), -1);
```
orangejulius: `t.ok(_.includes(supported_countries, 'USA'))` seems clearer, what do you think?

When ES7 rolls around we'll be able to write `supported_countries.contains('USA')` which will be amazing.

orangejulius:
:+1: looks great to me

PR #431, opened 2016-03-02
Deduper improved

vesameskanen:
It now considers locality and neighborhood, too. BTW, why not all existing fields? I guess there is still work to do here.


vesameskanen:
Sorry, made the request again against a wrong repo :(

orangejulius:
You've got to stop teasing us with these nice improvements :P

PR #434, opened 2016-03-03
Fix dedupe

vesameskanen:
Deduper should really consider locality and neighborhood, too. Especially with POI data, which is not necessarily bound to street addresses, current deduping works really badly.
 
It also seems that a missing field should not make an item different if it is otherwise identical with another item.  A more complete item gets higher scores, so it is safe to drop the document which just duplicates some fields.

Further thoughts: maybe the distance should be considered, too. This is not a  straightforward thing, because distance between street addresses should match well (say, with a 500 m limit), whereas location of a city can vary several kilometers.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,6 +18,8 @@ RUN npm install && \
   npm test && \
   chmod -R a+rwX .
 
+ADD pelias.json.docker pelias.json
+
 # Don't run as root, because there's no reason to (https://docs.docker.com/engine/articles/dockerfile_best-practices/#user).
 # This also reveals permission problems on local Docker.
 RUN chown -R 9999:9999 ${WORK}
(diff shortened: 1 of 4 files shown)


vesameskanen:
It seems that the base branch for this commit had also other commits - should I make a new cleaner pull request or can you cherry-pick from this?

dianashk:
Hey @vesameskanen, it would be excellent if you could split out the change specific to the dedupe step. 
Also, `master` is now using new admin property names, so we would need to update the property names to match the new ones. Sorry if this is a hassle. If you don't have the bandwidth to update it, we can try cherry-pick the change as you suggested. That might get messy with all the changes in `master` at this time.

vesameskanen:
No problem @dianashk - I will make a new, clean pull request against your latest master branch.

Issue #432, opened 2016-03-02
bulk geocoding proposal


dimfeld:
I'm starting to implement a "bulk geocoder" feature for the Pelias API. My current design doesn't make any major changes to the API, except that it allows you to POST a JSON array to the `/v1/search` endpoint, where each item in the array is a single request. The API will then run the middleware over each item in the array and use the Elasticsearch `msearch` endpoint to quickly perform all the searches at once. There will also be some way to configure a maximum amount of searches that can be submitted in one request.

I'll be glad to submit a PR for this once I have it working; are you interested in integrating such a feature into Pelias? And if so, do you have any feedback on how you'd like it to look?

riordan:
@dimfeld this is really awesome and actually something we've been thinking a bit about ourselves. It'd be great to hear a bit more about your perspective on this. Is there any chance we could chat next week? I think our team could really benefit.

dimfeld:
Sure, sounds good. I think Monday or Tuesday will work best; send me an email at my username at gmail.com and we can set up a time. Also I have my WIP code at github.com/Carevoyance/pelias-api, if you care to check it out.

orangejulius:
Closing this, we have https://github.com/pelias/scripts-batch-search which uses existing Pelias APIs in a convenient package. I don't believe Pelias needs a dedicated batch endpoint, as the main determinant of performance will be the hardware given to Pelias, rather than the way queries are sent to Pelias.

PR #436, opened 2016-03-08
Bulk search feature

dimfeld:
EDIT: Ignore this first paragraph. ~~I think it's ready now. DO NOT MERGE YET -- This still needs more testing before I'm confident that it's ready, but it is working in simple tests at this point so I'm posting it for feedback.~~

This PR allows you to POST a JSON array to the `/v1/search` endpoint. In response Pelias will send an `msearch` to Elasticsearch and then return an array of GeoJSON objects. The format of each GeoJSON object is unchanged.  See #432 as well. The maximum number of searches allowed in a single request can be configured by setting `bulkSizeLimit` in the Pelias JSON config; the default is currently to not enforce a limit but this probably isn't really tenable without extra functionality to split the batch into multiple msearch calls, since Elasticsearch doesn't like it if you give it too many searches at once.

Major changes:
* Addition of new msearch service, used from the search controller.
* sanitizeAll creates `req.clean` as an array. If the original request is a singleton or comes from a GET request, `sanitizeAll` marks the request with this information so that the `sendJSON` middleware can make sure that it returns an object instead of an array.
* New `iterate` helper function to allow the middlewares to work well when 

I still need to:
* Write new ciao tests to exercise the new functionality
* Look into a couple ciao tests that were failing, though I suspect these are because the Pelias index on my ES instance is still using the schema from a few months ago and needs to be recreated.
* Probably set a real default value (50? 500?) on bulkSizeLimit.

Additionally, I think the `search` ES endpoint is actually not used anymore since req.clean is always an array now. I left in the code that uses it right now but wanted to solicit feedback on how to proceed there.


dimfeld:
Things are looking good so far. Here's one question: The existing behavior of the API is to not do a search when a parsing error is encountered, and this PR brings that forward into the bulk search scenario, not performing any of the searches if a parse error was encountered on any item. 

In a bulk search scenario, do you think this is still desirable, or would it be better to only cancel the searches that had errors? I'm leaning toward changing it to still perform whichever searches for whichever items were valid.

dimfeld:
Added some ciao tests. The existing failures all passed once I created a new index that used the new schema. I also set the default limit on the POSTed array length to 50. I think this is ready to go. Please let me know if you have any comments  :)

missinglink:
hi @dimfeld this is super exciting, we have been talking about it in our meetings here at mapzen.

our product manager @riordan will reach out to you shortly to discuss in more detail.

dimfeld:
Any updates here? I haven't heard from anyone since the comment above and my window to do extra integration work on this is closing.

dianashk:
Hey @dimfeld, we are so super duper sorry about how long it's been since we've last replied. 😔  🐢 

You've really done a tremendous amount of work for this and we really appreciate it! ❤️  🎆 

It's a very big change that unfortunately doesn't align with our current roadmap. However, we'd love to chat with you in a hangout about the motivation behind this work. It would help to understand why you need such a feature and how we can accommodate those needs in another way in the short-term. We'd love to connect via search@mapzen.com to setup a call if you have some time to chat with the team.
Thanks again for all of your support and hard work! 

And again, we're really really really sorry to keep you waiting for so long. 💐

PR #443, opened 2016-03-09
Expose bbox

### LINKED ISSUE #398: API returns GeoJSON bounding boxes for individual features
Was talking this through with @orangejulius and we thought it'd benefit from a writeup to remember what we said:

### Background
Who's on First stores polygons for administrative areas. In addition to storing the points, we can store the bounding boxes to convey the extent of the place.

@trescube's been working on getting this data into ES in pelias/schema#97 and pelias/model#23.

### API Impact
[GeoJSON: Bounding Boxes](http://geojson.org/geojson-spec.html#bounding-boxes)

#### Today

Today, when we return a set of results as GeoJSON FeatureCollection (an array of Features), we pass back a single bounding box for all the documents in the set of results.

It's what lets us zoom a map to encompass the centroids of all the places in the results.

It's the difference between 2 results for `London, England` and `London`:

pelias_github_io_compare___v1_search_3fsize_2_text_london__20england

pelias_github_io_compare___v1_search_3fsize_2_text_london

But what it doesn't let us do is zoom in on the extent of a SINGLE place (like this search for `Canada`, that is very, very accurate in its portrayal).
pelias_github_io_compare___v1_search_3fsize_1_text_canada

#### WoF-enabled API
The GeoJSON spec allows [2 kinds of bounding boxes](http://geojson.org/geojson-spec.html#bounding-boxes):
- bounding boxes describing the extent of an entire FeatureCollection (what we do today)
- bounding boxes for an individual Feature

These can be mixed together in a single document, so an API consumer can know that if they're referencing just a single result or all the results.

**Individual Matches**

We should pass along the BBOX being stored in ES so it's a part of the match for each place. This lets us convey the shape without needing to pass along the entire geometry. Yay efficiency!

** FeatureCollection Result Set **
@orangejulius points out that the current approach provides a bounding box for all centroids, and if we were to continue doing so, it could clip the outer extent of those matches. So instead we should figure out the BBOX of all the BBOXes.

This will require re-serializing the strings to ints in the API (which we'll have to do coming back from ES anyway), then calculating the outermost extents of all those bboxes.

Then the FeatureCollection BBox is the BBOX of all the BBOXes.

dianashk:
- [x] expose `bounding_box` property when available in results
- [x] convert the `bounding_box` stored in ES as a string to an object
- [x] update the FeatureCollection bounding box computation to encompass all the bounding boxes in the results, not only the centroids
- [x] test stuff

___
:notes: **NOTE:** I have some reservations about changing the FeatureCollection bounding box to be larger than what the geometry property suggests. If users are relying on that to be the same as the generic geojson they see elsewhere, it could be really confusing. They might think they are missing data. I think for now we can keep it this way and document **REALLY** well. 
___
Closes #398

### DIFF
--- a/helper/geojsonify.js
+++ b/helper/geojsonify.js
@@ -29,7 +29,8 @@ var DETAILS_PROPS = [
   'locality_id',
   'locality_a',
   'neighbourhood',
-  'neighbourhood_id'
+  'neighbourhood_id',
+  'bounding_box'
 ];
 
 
@@ -50,11 +51,16 @@ function geojsonifyPlaces( docs ){
       return !!doc;
     });
 
+  // get all the bounding_box corners as well as single points
+  // to be used for computing the overall bounding_box for the FeatureCollection
+  var extentPoints = extractExtentPoints(geodata);
+
   // convert to geojson
-  var geojson = GeoJSON.parse( geodata, { Point: ['lat', 'lng'] });
+  var geojson             = GeoJSON.parse( geodata, { Point: ['lat', 'lng'] });
+  var geojsonExtentPoints = GeoJSON.parse( extentPoints, { Point: ['lat', 'lng'] });
 
   // bounding box calculations
-  computeBBox(geojson);
+  computeBBox(geojson, geojsonExtentPoints);
 
   return geojson;
 }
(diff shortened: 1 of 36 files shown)


riordan:
:clap:

riordan:
cc: @louh

trescube:
:+1:

louh:
:astonished: :eight_spoked_asterisk: :smile_cat: 

I can use this!
Out of curiosity why _wouldn't_ the bbox be in the geometry property?

riordan:
Because it can be one of many features, as opposed to the geometry. 

> To include information on the coordinate range for geometries, features, or feature collections, a GeoJSON object may have a member named "bbox"

PR #448, opened 2016-03-11
Pass esclient config to elasticsearch module

orangejulius:
This fixes a regression in #423 that breaks our configuration in dev and
production.


orangejulius:
TODO: how can we test this?

missinglink:
:+1: you can test this by stopping your local ES service and starting it on another port, the easiest way is to use an SSH port forward to dev and then change your `~/pelias.json` port to match that.

orangejulius:
This PR now includes tests using proxyquire!

dianashk:
:package:

PR #461, opened 2016-03-18
Modify region specific label generation

### LINKED ISSUE #410: Add support for macroregions and macrocounties
They're in ES docs, need to be returned by API

### LINKED ISSUE #454: Label generation / admin-lookup
Searching for `Wales` results in a strange label `Powys` because geonames refers to Wales as a region whereas WOF thinks it's a macroregion. So the admin-lookup does the wrong thing and then the label generator removes name and leaves only region and that shows up wrong, even though we found the right result.

Similar situation with London now having City of Westminster in the label.

trescube:
This fix does 3 things:

- exposes `macroregion` and `macrocounty` document fields
- adds `macroregion` to GBR and default label generation
- makes `region`-specific logic in initial label confined to US and CA

Fixes #454
Fixes #410


trescube:
While this PR addresses the root issue of enabling `macroregion` for labels, it still doesn't fix the problem of "Wales" returning the label "Wales, Powys, United Kingdom".  The Geonames record for "Wales" is categorized as `ADM1` so when the record lat/lon is reverse geocoded, "Powys" is listed as the `region` whereas `macroregion` is bigger so `region` should not be considered.

missinglink:
:+1: looks good to me

Issue #449, opened 2016-03-11
Add layers parameter to Autocomplete endpoint


riordan:
We should add the option to select by layer type in autocomplete, allowing users to select results of specific types (e.g. `venue`, `address`, `neighbourhood`, `locality`, `coarse`, `region`, etc.).

One of the things we discovered during our user testing is that some of our users will filter out autocomplete results so they can get the specific type of results they want. While our current autocomplete approach is designed to improve cache likelihood by permitting only limited configuration (query + optional focus.point position), it's too limiting for many intended autocomplete uses (see @louh's [blog post](https://mapzen.com/blog/find-your-community) asking for just addresses in New York City).

## What it'd look like

`search.mapzen.com/v1/autocomplete?layer=coarse&query=la leng`

When combined with #373, this will allow highly targeted and cache-able autocomplete queries.

## Outstanding Questions:
- [ ] What are the performance / cache likelihood implications of this change?
- [ ] Do we allow the selection of multiple layers at once with this query or just a single layer?
 - [ ] How bad is it to ask a user to make multiple queries to retrieve across multiple layers?

louh:
:+1: Incidentally, the Leaflet plugin now filters autocomplete by the requested layers, if provided (https://github.com/mapzen/leaflet-geocoder/releases/tag/v1.5.0). This might not work well if the autocomplete response contains no features in the requested layers. They will all get filtered out and appear in the UI as "no results found".

The reasoning here is that `layers` as an option for the plugin is expected to "just work" everywhere. So someone might specify multiple layers and expect the same experience from both `/search` and `/autocomplete`. If the plugin had to make 2 consecutive requests (or more) to `/autocomplete` and concatenate the results to provide a seamless experience, I'm worried that the additional requests will slow down the autocomplete experience overall.

missinglink:
I am :+1: on adding this API feature, let's pair this ticket with one to improve the key generation algorithm at the end cache.

The dev work on this is trivial, it's simply a product/caching question.

PR #468, opened 2016-03-24
Autocomplete api

missinglink:
This PR contains 3 commits:

commit https://github.com/pelias/api/pull/468/commits/63558eb15dbf1f6fb98d5dc5ab42ef6545fe65cd simply makes it easier to run ciao tests by providing a minimal set of test data which prevents the failures caused when running those tests against an empty index. it also includes a README section on how that all works.

commit https://github.com/pelias/api/pull/468/commits/8431763a122736a571fbacc468cc92721e88e051 fixes the `?size=` bug for `/v1/autocomplete`. this was a bug that allowed users to change the amount of records being returned from that endpoint, this commit fixes `size` to 10 and **doesn't allow users to change it**, a warning is returned in the JSON.

*note: I couldn't find an issue for this, does anyone know if there is a github issue for this ^?*

commit https://github.com/pelias/api/pull/468/commits/00b123adfbec4f362c66772ea5d0e3c66b3e0d42 enables both `?sources=` and `?layers=` filtering for the `/v1/autocomplete` endpoint. it's behaviour is identical to how it works on `/v1/search`.

*note: this will require us to update our API documentation*

[edit] I had to add one more commit https://github.com/pelias/api/pull/468/commits/0abda099a6722cde040c60251ed7310f5e3dc193 which fixed a bug which prevented the sources from not matching correctly! our tests were not covering some important cases, so I added more tests to prevent this issue in the future.

closes https://github.com/pelias/api/issues/449
related: https://github.com/pelias/api/pull/221
related: https://github.com/pelias/api/issues/236

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,3 +34,41 @@ The API recognizes the following properties under the top-level `api` key in you
 
 Please fork and pull request against upstream master on a feature branch. Pretty please; provide unit tests and script
 fixtures in the `test` directory.
+
+## Unit tests
+
+You can run the unit test suite using the command:
+
+```bash
+$ npm test
+```
+
+## HTTP tests
+
+We have another set of tests which are used to test the HTTP API layer, these tests send expected HTTP requests and then
+assert that the responses coming back have the correct geoJSON format and HTTP status codes.
+
+You can run the HTTP test suite using the command:
+
+```bash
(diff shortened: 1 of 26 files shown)


dianashk:
:+1:

Issue #440, opened 2016-03-09
Reverse endpoint filtering has no effect


orangejulius:
In the master branch, the API currently doesn't correctly filter sources or layers: http://pelias.github.io/compare/#/v1/reverse%3Fsources=geonames&point.lat=49.546418&point.lon=5.818429

missinglink:
it turns out that `?layers=` was working fine, it was just `?sources=` that wasn't working as expected.

PR #469, opened 2016-03-24
bugfix: enable support for ?sources= filtering on /reverse

missinglink:
This PR re-enables `?sources=` filtering for the `/v1/reverse` endpoint.

resolves https://github.com/pelias/api/issues/440


orangejulius:
Looks good!

PR #476, opened 2016-03-29
quattroshapes deprecation warning and wof fallback

### LINKED ISSUE #442: layers=(qs or quattroshapes) should forward to layers=whosonfirst
Several of our existing users make calls to quattroshapes. When we disable Quattroshapes, they'll need to be able to still make those calls while getting the same kind of records. We should forward these requests to Who's on First.

missinglink:
this PR fixes some bugs and tests associated with the 'quattroshapes deprecation' sanitizer.

Fixes #442


orangejulius:
I updated the error message to present tense, and tested it on prod_build with the acceptance tests.

:+1:

PR #482, opened 2016-04-05
Change properties.bounding_box to bbox at top level

### LINKED ISSUE #479: Feature-level bbox is geoJSON compliant
Currently, results from Who's on First and OSM Ways can return a bounding box for the feature (in addition to its centerpoint). This core work was done in #443.

`bbox` should be at the same level as `geometry` and `properties` in a feature, like

```
{ "type": "Feature",
    "bbox": [-10.0, -10.0, 10.0, 10.0],
    "geometry": {
      "type": "Polygon",
      "coordinates": [[
        [-10.0, -10.0], [10.0, -10.0], [10.0, 10.0], [-10.0, 10.0]
        ]]
      }
    ...
    }
```

dianashk:
Feature bounding boxes were added under the properties object and had the format of `min_lon`, `min_lat`, `max_lon`, `max_lat`. However it didn't match the geojson specification which calls for the array format with the minimums of each axis first, and maximum last, as follows: [`min_lon`, `min_lat`, `max_lon`, `max_lat`]. According to the spec, this `bbox` property should also reside on the same level as the `type` and `geometry` objects.

Fixes #479

### DIFF
--- a/helper/geojsonify.js
+++ b/helper/geojsonify.js
@@ -35,8 +35,7 @@ var DETAILS_PROPS = [
   'locality_id',
   'locality_a',
   'neighbourhood',
-  'neighbourhood_id',
-  'bounding_box'
+  'neighbourhood_id'
 ];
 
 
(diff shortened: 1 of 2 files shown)


orangejulius:
:+1:

orangejulius:
Tested this locally, looks like it works great

PR #497, opened 2016-04-12
Refactored the label generation code to be simpler

### LINKED ISSUE #496: Rework label generation to mimic google
Our current label generation is a series of fallbacks that can lead to very strange looking labels.  To mimic google, refactor for simplicity.  US and CA should include province, US should include NYC borough when applicable, UK should include macroregion.

trescube:
Labels are now, outside US/CA/UK, just name+city+country.

Deduping now only occurs on the first 2 parts of the labels which fixes the "Luxembourg, Luxembourg" issue.

The general strategy is:

- UK labels include macroregion (England, Northern Ireland, etc)
  - London, England, United Kingdom
  - Glasgow, Scotland, United Kingdom
- US labels include NYC borough (where applicable) AND state abbreviation
  - 102 South Charles Street, Red Lion, PA, United States
  - Truth or Consequences, NM, United States
  - 30 W 26th St, Manhattan, New York, NY, United States
- CA labels include province abbreviation (when available)
  - 8 HOFFMAN ST, Kitchener, Ontario, Canada
  - Tim Hortons, Kitchener, Ontario, Canada
- all other countries are just name+city+country
  - 74 rue de rivoli, Paris, France
  - Tour Eiffel, Paris, France
  - Cape Town, South Africa
  - Luxembourg, Luxembourg

Functionality has been retained to return full US/CA state/province name when the result layer is `region`.  

Fixed #496 and pelias/pelias#299


helper/labelGenerator.js
```diff
@@ -4,37 +4,28 @@ var _ = require('lodash'),
...
-  //
-  //  this code doesn't have the same effect in the case of a venue or address
-  //  where the `name` field would contain the address or name of a point-of-interest
-  //
-  //  Also see https://github.com/pelias/api/issues/429 for other ways that this is bad
-  //
-  // de-dupe, join, trim
-  return _.uniq( labelParts ).join(', ').trim();
+  // retain only things that are defined
+  labelParts = labelParts.filter(function(v) { return !_.isUndefined(v); });
```
orangejulius: could you replace this with

`labelParts = _.compact(labelParts)`? 

https://lodash.com/docs#compact

I think so, as long as any falsy value is ok to remove

trescube: Oh! I did not know that existed!

helper/labelGenerator.js
```diff
@@ -4,37 +4,28 @@ var _ = require('lodash'),
...
+
+  // first, dedupe the name and 1st label array elements
+  //  this is used to ensure that the `name` and first admin hierarchy elements aren't repeated
+  //  eg - `["Lancaster", "Lancaster", "PA", "United States"]` -> `["Lancaster", "PA", "United States"]`
+  var dedupedNameAndFirstLabelElement = _.uniq([labelParts.shift(), labelParts.shift()]);
+
+  // second, unshift the deduped parts back onto the labelParts
+  labelParts.unshift.apply(labelParts, dedupedNameAndFirstLabelElement);
```
orangejulius: doesn't `labelParts.unshift(dedupedNameAndFirstLabelElement)` do the same thing?

orangejulius:
:+1: after the minor comments are addressed. I like how the example tests are split out from the more synthetic ones

helper/labelGenerator.js
trescube: No, because `dedupedNameAndFirstLabelElement` is an array whereas `unshift` takes `varargs`-style parameters.

orangejulius: ahh. what about `labelParts = dedupedNameAndFirstLabelElement.concat(labelParts)` then?

trescube: Sure, that works, too.  I chose `unshift` because `concat` creates a new array.  Certainly it's an academic exercise.

orangejulius: Hah, funny, I was thinking `concat` is clearer for exactly that reason. `apply` takes a bit of extra thinking to figure out what it will do.

At the very least, does something like

`Array.unshift.apply(labelParts, dedupedNameAndFirstLabelElement)` work?

That way the reader doesn't have to wonder what's special about the unshift method on `labelParts`

PR #503, opened 2016-04-14
Fix timeout error handling

orangejulius:
Timeout errors come back from the Elasticsearch module as Error objects,
so instead of pushing the entire object onto the errors array, we should
just push the message.

I'm not sure if there are other cases where errors are just strings,
errors that aren't objects with a message property are handled as
before.


dianashk:
Looks good. :+1:

PR #504, opened 2016-04-14
Use new express sendStatus method

orangejulius:
According to a message on the console:

```
express deprecated res.send(status): Use res.sendStatus(status) instead
middleware/options.js:12:9
```


dianashk:
Cool beans! :+1:

PR #507, opened 2016-04-14
added special case for USA in labels

### LINKED ISSUE #506: Special case US result labels
All other countries should be full country name but for US results, use 'USA' instead of 'United States'

trescube:
if record is for literal United States, use the full name, otherwise use USA

Fixed #506


orangejulius:
:+1:

PR #517, opened 2016-04-18
fixed extraneous comma labels for results with only a `name`

### LINKED ISSUE #516: labels with a single part should not have a trailing comma
In the document below, we need to avoid adding the ", " to the end of labels when their is no 'part' to print after, such as: `"label": "Northern Europe, "`.

```javascript
    {
      "type": "Feature",
      "geometry": {
        "type": "Point",
        "coordinates": [
          2.90039,
          61.68987
        ]
      },
      "properties": {
        "id": "7729883",
        "gid": "geonames:venue:7729883",
        "layer": "venue",
        "source": "geonames",
        "name": "Northern Europe",
        "confidence": 0.662,
        "label": "Northern Europe, "
      }
    }
```

trescube:
Fixed #516 

This fixes geonames records like "Northern Europe" where there are no admin fields, just a name.  The old code was introducing a null due to how parts of the label were deduped.

### DIFF
--- a/helper/labelGenerator.js
+++ b/helper/labelGenerator.js
@@ -16,18 +16,27 @@ module.exports = function( record ){
   // retain only things that are truthy
   labelParts = _.compact(labelParts);
 
-  // first, dedupe the name and 1st label array elements
-  //  this is used to ensure that the `name` and first admin hierarchy elements aren't repeated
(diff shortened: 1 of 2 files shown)


orangejulius:
:+1:

PR #522, opened 2016-04-21
Search borough when searching admin fields

orangejulius:
Also includes some nice refactoring to use the placetypes helper module


trescube:
👍

trescube:
this didn't affect any tests?

orangejulius:
No :( , we probably want to add some to the bit of code where we assign text parser output to query variables.

PR #526, opened 2016-04-25
autocomplete milestone

missinglink:
This PR contains all the work that's gone in to the [Autocomplete Improvements Milestone](https://github.com/pelias/api/milestones/Autocomplete%20Improvements), it is paired with a [corresponding schema update](https://github.com/pelias/schema/pull/127).

The PR is focused around fixing bugs and flaky behavior in `/v1/autocomplete` that's been reported by our users, I've also taken the opportunity to triage all the incoming requests for autocomplete, clean up some messy parts and add more code coverage.

The major difference in behavior is due to how we handle analysis and especially synonym substitution for autocomplete, an example of a query which was not previously possible using phrase matching:

#### improved analysis

```
/v1/autocomplete?text=vic uni wellington

1)	Victoria University of Wellington, New Zealand
...
```

more information about the analysis changes can be found in: https://github.com/pelias/schema/issues/105

#### final token is a stopword bug

This reported bug has been squashed: https://github.com/pelias/pelias/issues/211

```
/v1/autocomplete?text=green lane

 1)	Green Lane, PA, USA
 2)	Green Lane Farms, Lower Allen, PA, USA
...
````

#### improved server-side tokenizer

Improved parsing of source data containing commas or slashes as delimiters, such as in `Bedell Street/133rd Avenue`. see: https://github.com/pelias/schema/pull/113

#### improved client-side tokenizer

In order to generated the best/most efficient queries possible, we've added a client-size tokenizer which mimics the functionality of the server-side tokenizer. see: https://github.com/pelias/api/pull/529

#### improved handling of ordinals

This allows users to partially type an address such as "33rd street" as "33r", see https://github.com/pelias/schema/pull/96

#### improved admin matching, particularly for non-delimited queries

this PR also includes a fix for `parent.borough` not being targeted during admin patching. see https://github.com/pelias/api/pull/527

```
/v1/autocomplete?text=200 dean st brooklyn

 1)	200 Dean Street, Brooklyn, New York, NY, USA
```

#### indexing of single digit numbers

in the past we did not index single digit numbers, they are now present in the index without causing a major increase in disk/ram usage.

```
/v1/autocomplete?text=grolmanstr 8

1)	Grolmanstr. 8, Köln, Germany
...
```

#### tighter 'focus' clustering and improved localized matching

see: http://missinglink.embed.s3.amazonaws.com/pelias_clustering.png

#### house number can be provided in 1,2 or 3rd position

mainly for for Germanic addresses schemes. see https://github.com/pelias/api/pull/489

```
/v1/autocomplete?text=wilmersdorfer str 51

1)	Wilmersdorfer Straße 51, Kiel, Germany
...
```

#### boosting of exact matches

an additional query segment was added to ensure that outputs that more closely match the input text appear higher than those which only partially match.

### CODE REVIEW

File: query/autocomplete_defaults.js [hunk 1]
```diff
@@ -82,6 +82,10 @@ module.exports = _.merge({}, peliasQuery.defaults, {
   'admin:neighbourhood:field': 'parent.neighbourhood',
   'admin:neighbourhood:boost': 200,
 
+  'admin:borough:analyzer': 'peliasAdmin',
+  'admin:borough:field': 'parent.borough',
+  'admin:borough:boost': 800,
```
dianashk: why so high?

query/search.js
```diff
@@ -30,6 +30,7 @@ query.score( peliasQuery.view.admin('country_a') );
 query.score( peliasQuery.view.admin('region') );
 query.score( peliasQuery.view.admin('region_a') );
 query.score( peliasQuery.view.admin('county') );
+query.score( peliasQuery.view.admin('borough') );
```
orangejulius: Should we remove this to avoid having any effect on search from this PR?

query/text_parser.js
```diff
@@ -9,6 +9,7 @@ or postalcode because we should only try to match those when we're sure that's w
  */
 var adminFields = placeTypes.concat([
   'region_a',
+  'borough'
```
dianashk: this doesn't need to be done since placeTypes already has borough in the list.

query/view/boost_exact_matches.js
```diff
@@ -0,0 +1,48 @@
...
+  // passed to other views.
+  var vsCopy = new peliasQuery.Vars( vs.export() );
+
+  // copy phrase:* values from search defaults
+  vsCopy.var('phrase:analyzer').set(searchDefaults['phrase:analyzer']);
+  vsCopy.var('phrase:field').set(searchDefaults['phrase:field']);
+
+  // split the 'input:name' on whitespace
+  var name = vs.var('input:name').get(),
+      tokens = name.split(' ');
```
orangejulius: perhaps split on any whitespace so that multiple spaces, tabs, or any other weird stuff is handled ok? Or would that be cleaned up by a sanitizer before this?

dianashk: this could result in empty tokens if there are multiple spaces in a row. just doesn't feel like a very clean solution.

query/view/pop_subquery.js
```diff
@@ -0,0 +1,16 @@
+
+var peliasQuery = require('pelias-query');
+
+/**
+  Population / Popularity subquery
+**/
+
+module.exports = function( vs ){
```
dianashk: could this possibly go into the pelias-query module? seems like the type of thing we typically keep in there.

orangejulius:
We noticed a regression from pelias/pelias#164. When searching for "New York, NY" from Stamford, CT, some of the below queries work well on prod and prodbuild, but other's dont:


[Search for New York, NY](http://pelias.github.io/compare/#/v1/search%3Ffocus.point.lat=41.05343&focus.point.lon=-73.53873&text=new%20york,%20NY)
[Search for New York](http://pelias.github.io/compare/#/v1/search%3Ffocus.point.lat=41.05343&focus.point.lon=-73.53873&text=new%20york)
[Autocomplete for New York, NY](http://pelias.github.io/compare/#/v1/autocomplete%3Ffocus.point.lat=41.05343&focus.point.lon=-73.53873&text=new%20york,%20NY)
[Autocomplete for New York](http://pelias.github.io/compare/#/v1/autocomplete%3Ffocus.point.lat=41.05343&focus.point.lon=-73.53873&text=new%20york)

query/autocomplete_defaults.js
missinglink: I simply copied the value from 'country', I will drop it down to a value for 'region' instead if you feel it's too high. What would you suggest as a good value here?

query/text_parser.js
missinglink: good catch!

query/search.js
missinglink: moving this to https://github.com/pelias/api/pull/527

orangejulius:
:+1:

PR #527, opened 2016-04-28
enable borough for search

missinglink:
Enable matching for borough from search endpoint.

acceptance test: https://github.com/pelias/acceptance-tests/blob/master/test_cases/address_matching.json#L105-L120


orangejulius:
:+1:

PR #529, opened 2016-04-28
improve api-side tokenizer

missinglink:
a tokenizer is required in the API code in order to ascertain which parts of the input text are 'complete' tokens and which tokens are only partially complete (AKA. incomplete).

in the past we had a couple of different techniques for doing this, neither of which was very robust compared to the analysis we do on the elasticsearch side.

this PR introduces a simple `tokenizer` step which splits up the tokens and classifies them as either 'complete' or 'incomplete'.

as a result, a bunch of messy code was removed (eg. the concept of `'input:name:tokens_incomplete'` ) and some improvements were made to how we generate queries (see test/unit/fixture/* for more info).

the code is also now significantly better covered by unit tests

note: the tokenizer splits on commas, this will likely have no effect as those commas are swallowed by the text parser, I've left the comma in anyway so that the list of delimiters matches exactly the ones we use on the elasticsearch side.

the 'guts' of the PR can be found in:
- [sanitiser/_tokenizer.js](https://github.com/pelias/api/blob/missinglink_complete_incomplete_refactor/sanitiser/_tokenizer.js)
- [test/unit/sanitiser/_tokenizer.js](https://github.com/pelias/api/blob/missinglink_complete_incomplete_refactor/test/unit/sanitiser/_tokenizer.js)

### DIFF
--- a/sanitiser/_tokenizer.js
+++ b/sanitiser/_tokenizer.js
@@ -1,0 +1,95 @@
+
+var check = require('check-types');
+
+/**
+  simplified version of the elaticsearch tokenizer, used in order to
+  be able to detect which tokens are 'complete' (user has finished typing them)
+  or 'incomplete' (the user has possibly only typed part of the token).
+
+  note: we don't need to strip punctuation as that will be handled on the
+  elasticsearch side, so sending a token such as 'st.' is not an issue, these
+  tokens should *not* be modified as the anaylsis can use the punctuation to
+  infer meaning.
+
+  note: this sanitizer should run *after* the '_text' sanitizer so it can
+  use the output of clean.parsed_text where available.
+**/
+function sanitize( raw, clean ){
+
+  // error & warning messages
+  var messages = { errors: [], warnings: [] };
+
+  // this is the string we will use for analysis
+  var text = clean.text;
+
+  // a boolean to track whether the input parser successfully ran; or not.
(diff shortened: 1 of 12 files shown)


test/unit/sanitiser/_tokenizer.js
```diff
@@ -0,0 +1,425 @@
...
+
+    // no errors/warnings produced
+    t.deepEquals(messages.errors, [], 'no errors');
+    t.deepEquals(messages.warnings, [], 'no warnings');
+
+    t.end();
+  });
+};
+
+module.exports.all = function (tape, common) {
```
orangejulius: These are some excellent tests

sanitiser/_tokenizer.js
```diff
@@ -0,0 +1,95 @@
...
+    match the ngrams index (such as single grams not stored in the index).
+  **/
+
+  // split the tokens in to 'complete' and 'incomplete'.
+  if( clean.tokens.length ){
+
+    // if all the tokens are complete, simply copy them from clean.tokens
+    if( inputParserRanSuccessfully ){
+
+      // all these tokens are complete!
```
orangejulius: I believe the input parser runs on autocomplete as well. so is there a guarantee that all tokens are complete in that case?

orangejulius:
:+1:

Issue #531, opened 2016-05-02
Incorrect error code for elasticsearch timeouts


missinglink:
I noticed today that API errored with the message `Request Timeout after 10000ms` and the status code of `400`.

The correct behaviour here is that API should return status `408 Request Timeout` as per the code here: https://github.com/pelias/api/blob/master/middleware/sendJSON.js#L28

It's important to fix this so that tools consuming the API don't incorrectly think that the error is `400 Bad Request` (a client error) when it's actually a server-side error.

```
Unexpected status code 400.

Failed for test case:
{
    "statusCode": 400,
    "body": {
        "geocoding": {
            ...
            "errors": [
                "Request Timeout after 10000ms"
            ],
        },
        "type": "FeatureCollection",
        "features": []
    },
    "headers": {
        "access-control-allow-credentials": "true",
        "access-control-allow-headers": "X-Requested-With,content-type",
        "access-control-allow-methods": "GET, OPTIONS",
        "access-control-allow-origin": "*",
        "cache-control": "public",
        "charset": "utf8",
        "content-type": "application/json; charset=utf-8",
        "date": "Mon, 02 May 2016 11:38:51 GMT",
        "etag": "W/\"191-78EYeNDDZWE6nCIXcOQ7uA\"",
        "server": "Pelias/2.2.0",
        "x-powered-by": "mapzen",
        "content-length": "401",
        "connection": "Close"
    },
    "request": {
        "uri": {
            "protocol": "http:",
            "slashes": true,
            "auth": null,
            "host": "pelias.prodbuild.mapzen.com",
            "port": null,
            "hostname": "pelias.prodbuild.mapzen.com",
            "hash": null,
            "search": "?text=30%20w%2026%20st",
            "query": "text=30%20w%2026%20st",
            "pathname": "/v1/search",
            "path": "/v1/search?text=30%20w%2026%20st",
            "href": "http://pelias.prodbuild.mapzen.com/v1/search?text=30%20w%2026%20st"
        },
        "method": "GET",
        "headers": {
            "accept": "application/json"
        }
    }
}
```

orangejulius:
We also send 400 errors for some other Elasticsearch errors, so something in our HTTP status code handling is definitely wrong.

PR #533, opened 2016-05-03
refactor elasticsearch error detection, improve test coverage

missinglink:
this PR refactors the way we detect which elasticsearch error was returned by the elasticsearch client.

prior to this PR we were using `err.constructor.name` to try to detect the Error type, this seems to be ineffective/broken and there were no tests covering this functionality.

added a bunch of unit tests to cover all the cases I could think of.

all ciao tests passing.

closes https://github.com/pelias/api/issues/531

### DIFF
--- a/middleware/sendJSON.js
+++ b/middleware/sendJSON.js
@@ -1,4 +1,5 @@
-var check = require('check-types');
+var check = require('check-types'),
+    es = require('elasticsearch');
 
 function sendJSONResponse(req, res, next) {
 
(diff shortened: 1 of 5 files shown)


orangejulius:
most excellent bugfix!

dianashk:
:+1:

orangejulius:
Hm. Some elasticsearch errors still result in 400 status codes. The easiest one to reproduce is to drop the schema and then load a really old one (so that the analyzers mismatch).

missinglink:
you're right @orangejulius third time's a charm? I added this commit https://github.com/pelias/api/pull/533/commits/0caca08787170977bb2d3fc16b1a62100fe0150c which handles the case where the javascript library returns a string error instead of an error object.

dianashk:
:+1: I like the refactoring to use elasticsearch-exception instead of hardcoding them like you had before.

PR #542, opened 2016-05-16
Respect privacy headers

### LINKED ISSUE #460: Respect privacy headers
Stop logging requests when DO_NOT_TRACK header is sent.

orangejulius:
In order to provide a super privacy conscious experience for Pelias users, this PR changes the way our logging works when the Do Not Track header is enabled.

Custom handlers for fields in our access logs are added. They use a new logging module that can detect whether or not DNT is enabled. When DNT is enabled, IP addresses, lat/lons, and the value in the `text` field are removed from the logs.

It also removes some debug logs that were always printing info from queries. 

Fixes #460

### DIFF
--- a/app.js
+++ b/app.js
@@ -5,7 +5,7 @@ var peliasConfig = require( 'pelias-config' ).generate().api;
 
 
 if( peliasConfig.accessLog ){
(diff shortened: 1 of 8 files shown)


trescube:
👍 great for throughness

PR #543, opened 2016-05-17
Add source_id property

### LINKED ISSUE #541: `source_id` property is missing in search results
Not sure where it went, but we need to bring it back!

dianashk:
Connected to #541 

Still TBD:
- [ ] fix documentation to include `source_id` in response examples
- [ ] add acceptance test to validate all expected properties to be part of the response


orangejulius:
:+1:

Issue #510, opened 2016-04-15
API incorrectly warns against using `boundary.radius` parameter in reverse


orangejulius:
See http://pelias.github.io/compare/#/v1/reverse%3Fpoint.lat=40&point.lon=-65&boundary.circle.radius=400&sources=wof

If you change the radius to something bigger, results will show up, so this parameter definitely has an effect. And it is listed in our API spec. I think this warning carries over from /search, where it is _not_ supported yet.

dianashk:
Code has been merged and we will close this issue once it rolls up to production. Cheers!

PR #546, opened 2016-05-24
Implement categories filter

dianashk:
Connected to pelias/pelias#150

### DIFF
--- a/sanitiser/_categories.js
+++ b/sanitiser/_categories.js
@@ -1,34 +1,50 @@
 
 var check = require('check-types');
 
+var ERRORS = {
+  empty: 'Categories parameter cannot be left blank. See documentation of service for valid options.',
+  invalid: 'Invalid categories parameter value(s). See documentation of service for valid options.'
+};
+
 // validate inputs, convert types and apply defaults
-function sanitize( raw, clean ){
+function sanitize( raw, clean, validCategories ) {
 
   // error & warning messages
-  var messages = { errors: [], warnings: [] };
+  var messages = {errors: [], warnings: []};
 
-  // default case (no categories specified in GET params)
-  clean.categories = [];
+  // it's not a required parameter, so if it's not provided just move on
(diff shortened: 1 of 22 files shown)


dianashk:
TBD: figure out why `reverse` was set to not handle `boundary.circle` filter. This is needed for the new `nearby` and I'd like to know the technical implications of the previous decision before enabling it for `nearby`. Important to note that it will not be enabled in `reverse`. In fact the goal of this entire PR is to preserve `reverse` as-is to the greatest extent possible.

orangejulius:
related: https://github.com/pelias/api/issues/510

Also, I just realized, based on the [wiki page for what we want from reverse geocoding](https://github.com/pelias/pelias/wiki/Reverse-Geocoding,-Part-Deux) that we may eventually want to depricate boundary.circle. Reverse should only return things that are extremely close (a few meters) to the point specified, so a boundary radius doesn't make sense anymore. Does that sound right?

missinglink:
@orangejulius I think you're right, the `boundary.circle` API will need to remain internally but we can remove it from our external APIs.

The original intention of that parameter was to do something like `nearby` where you might need to expand the search area (something like squash courts for instance) to get a decent list of results.

dianashk:
I would like handle the `boundary.circle.radius` work separately and leave it unchanged in this PR. I think we will need to update the default for this param but probably can't remove it fully until v2.0.
I've assigned the incorrect warning ticket to myself and will also handle it separately.

trescube:
:shipit: good call on splitting out parts of geojsonify to new files

sanitiser/_categories.js
```diff
@@ -1,34 +1,53 @@
 
 var check = require('check-types');
+var categoryTaxonomy = require('pelias-categories');
+
+var ERRORS = {
```
orangejulius: I like this! Much better than having strings all over the code like we probably do in other sanitisers.

orangejulius:
LGTM and works locally on my machine!

PR #549, opened 2016-05-25
Lower default reverse search radius, add a bigger radius for coarse reverse

orangejulius:
Defaulting to 500km for the reverse search radius means lots of records will often be searched, when we only care about the nearest ones. This PR reduces that radius to 1km (it's a bit more difficult to specify different units in the two defaults, but 1km is still quite performant).

It also adds a second default, for when any coarse layer was manually specified. This remains at 500km.

Fixes pelias/pelias#340

### DIFF
--- a/query/reverse_defaults.js
+++ b/query/reverse_defaults.js
@@ -12,7 +12,7 @@ module.exports = _.merge({}, peliasQuery.defaults, {
   'sort:distance:order': 'asc',
   'sort:distance:distance_type': 'plane',
 
-  'boundary:circle:radius': '500km',
+  'boundary:circle:radius': '1km',
   'boundary:circle:distance_type': 'plane',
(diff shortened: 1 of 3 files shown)


trescube:
👍

orangejulius:
Update on performance:

Using our [loadtest](https://github.com/pelias/loadtest) framework, our dev instance (4 r3.xlarge AWS boxes for elasticsearch), we were previously only getting 1 request/second throughput when using NYC address points from OpenAddresses to generate reverse queries.

Now we can consistently serve 40 requests/second for NYC, and 70 for a desolate area in new mexico!

orangejulius:
Another data point: looking at the query generated with this PR, Elasticsearch would previously have to score 24 _million_ documents on a full planet build, now with the 1km radius it scores only 10k

dianashk:
:+1:

PR #555, opened 2016-06-03
Redirect to /v1 if no legacy proxy specified

### LINKED ISSUE #511: Make root path give more friendly instructions
If you visit the root path such as http://pelias.prodbuild.mapzen.com/ or http://search.mapzen.com/?api_key=search-yourkey, it gives a rather cryptic error.

http://search.mapzen.com/v1/?api_key=search-yourkey is better, but still could be more friendly.

orangejulius:
For when we have multiple active API versions, we have code to redirect
from one running API server to another. Right now, we've disabled the
beta/legacy API server and haven't released a V2 yet, so we currently
have no proxying configured.

Right now the default with no configuration is to simply return an
error. However that means that request to http://search.mapzen.com/
return that unfriendly error. We have a nice friendly /v1 page, so in
the case of no proxy being set up, it makes sense to redirect to that.

Fixes #511

### DIFF
--- a/routes/legacy.js
+++ b/routes/legacy.js
@@ -4,8 +4,8 @@ function addRoutes(app, peliasConfig) {
   var sendToLegacy;
 
   if (!peliasConfig.hasOwnProperty('legacyUrl')) {
-    sendToLegacy = function doNothing(req, res, next) {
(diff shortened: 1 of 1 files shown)


trescube:
👍

PR #562, opened 2016-06-13
fix: Explicitly depend on packages' optional dependencies

orangejulius:
[istanbul](https://www.npmjs.com/package/istanbul) and
[escodegen](https://www.npmjs.com/package/escodegen) have optional
dependencies, which for some reason `npm ls` in npm version 3 fails to
handle properly. We don't need these packages, but installing them does
little harm so it seems to be the best way to make our precommit-hook
scripts pass for all users.

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,8 +66,10 @@
     "nsp": "^2.2.0",
     "precommit-hook": "^3.0.0",
     "proxyquire": "^1.7.7",
+    "source-map": "^0.5.6",
     "tap-dot": "1.0.5",
-    "tape": "^4.5.1"
+    "tape": "^4.5.1",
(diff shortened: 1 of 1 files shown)


orangejulius:
I'm open to other ideas of how to fix this as well, I spent about 30 minutes looking into it, but I'm sure there's something better.

dianashk:
LGTM 🌴 ☀️ 🌊

orangejulius:
So I did a little bit more investigation into this, and I'm less confident it's a NPM bug of some kind, and more likely it's because we have some really old dependencies in some of our submodules. See https://github.com/pelias/api/issues/564

PR #567, opened 2016-06-22
Focus offset minimum

missinglink:
Set all `focus.offset` values to a minimum of `0km` instead of `1km`.

This was a bug and an oversight, the `/autocomplete` behavior was set to `0km` but wasn't changed for `/search` or `/reverse`.

The effect of this was scoring all matching documents within 1km of the `focus.point` with the same score!

This should make some of the intermittent test failures settle down a bit and will give much better scoring for results within 1km of the target.


orangejulius:
:+1:

trescube:
👍

PR #573, opened 2016-06-30
Updating list of layers for Geonames

### LINKED ISSUE #569: Update allowed list of layers for Geonames
The list of [allowed layers](https://github.com/pelias/api/blob/master/helper/type_mapping.js#L50) for Geonames is out of date. Over time we've added better [layer mappings](https://github.com/pelias/geonames/blob/master/lib/streams/layerMappingStream.js) to the Geonames importer, but haven't kept the API list up to date.

We need to ensure that every layer that a Geonames record can be is listed in the API. I think we have to add `macroregion` and `localadmin`, but there may be more.

Compare these two queries to see the issue. The first one searches Geonames for Detroit, and the second result is of type localadmin. The second query searches Geonames for only localadmin records, but it returns an error saying no such records exist (they obviously do)

http://pelias.github.io/compare/#/v1/search%3Fsources=gn&size=2&text=detroit,%20mi
screenshot from 2016-06-29 15-03-54
screenshot from 2016-06-29 15-04-06

http://pelias.github.io/compare/#/v1/search%3Fsources=gn&layers=localadmin&size=2&text=detroit,%20mi
screenshot from 2016-06-29 15-03-16

Related to https://github.com/pelias/api/issues/484 (and could probably be done at the same time)

avulfson17:
Added macroregion, localadmin and neighbourhood to the list of layers that are available in geonames.
Fixes #569

### DIFF
--- a/helper/type_mapping.js
+++ b/helper/type_mapping.js
@@ -47,7 +47,8 @@ var SOURCE_MAPPING = addStandardTargetsToAliases(SOURCES, SOURCE_ALIASES);
 var LAYERS_BY_SOURCE = {
  openstreetmap: [ 'address', 'venue' ],
  openaddresses: [ 'address' ],
- geonames: [ 'country', 'region', 'county', 'locality', 'venue' ],
+ geonames: [ 'country','macroregion', 'region', 'county','localadmin',
+  'locality', 'neighbourhood', 'venue' ],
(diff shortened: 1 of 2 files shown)


trescube:
👍

orangejulius:
LGTM

orangejulius:
:shipit:

orangejulius:
Bleh, LGTM is being weird again.

PR #576, opened 2016-07-05
Shows county in label if there isnt a locality or local admin

### LINKED ISSUE #575: Use county in label when no locality is found
From the discussion in https://github.com/pelias/pelias/issues/167 we've found a case where records that don't have a locality lead to poor labels. For example, an address _just_ outside the city limits of Davis, California has the label `110A Sage Street, CA, USA`, which is not so good.

In this case, we want to fall back and use the county, so the label would be `110A Sage Street, Yolo County, CA, USA`. Of course, we don't want to add the county to the label in general, since that's generally redundant and not particularly interesting to most people.

avulfson17:
Fixes #575

### DIFF
--- a/helper/labelSchema.js
+++ b/helper/labelSchema.js
@@ -12,7 +12,7 @@ module.exports = {
   },
   'USA': {
     'borough': getFirstProperty(['borough']),
-    'local': getFirstProperty(['locality', 'localadmin']),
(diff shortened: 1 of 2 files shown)


trescube:
👍

orangejulius:
:shipit:

PR #580, opened 2016-07-07
Using new layers view to filter layers

### LINKED ISSUE #579: Use layer field, not _type, to filter by layers
Currently, records of different layers have their layer stored both in a [field called layer](https://github.com/pelias/schema/blob/master/mappings/document.js#L11), and it's also stored as the Elasticsearch "type", which is a feature of elasticsearch that allows you to store different kinds of data in the same index. We want to move away from this and simplify the API code in the process.

Sources also have their [own field](https://github.com/pelias/schema/blob/master/mappings/document.js#L10) as well, and we already filter on that field using a [query view](https://github.com/pelias/query/blob/master/view/sources.js), so this could be used as a model.

This should allow us to unify some code, and remove code like [this](https://github.com/pelias/api/blob/master/controller/search.js#L34-L37). It also will give us more flexibility in the future regarding how we use Elasticsearch.

## Notes
This will also require making changes to the [query module](https://github.com/pelias/query). It probably makes sense to start with a pull request in that repo that adds a new view for `terms` queries, and modifies the [sources](https://github.com/pelias/query/blob/master/view/sources.js) view to use it. In that same PR a new `layers` view could be added, and then a pull request to API to use that new view could be made.

## Relevant Elasticsearch docs:
https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-type-field.html
https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping.html#mapping-type
https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-term-query.html

avulfson17:
Needs https://github.com/pelias/query/pull/28 to be merged first
Fixes #579

### DIFF
--- a/controller/search.js
+++ b/controller/search.js
@@ -31,11 +31,6 @@ function setup( backend, query ){
       body: query( req.clean )
     };
 
-    // use layers field for filtering by type
-    if( req.clean.hasOwnProperty('layers') ){
-      cmd.type = req.clean.layers;
-    }
-
     logger.debug( '[ES req]', cmd );
 
     // query backend
--- a/query/search.js
+++ b/query/search.js
@@ -43,6 +43,7 @@ query.score( peliasQuery.view.admin_multi_match(adminFields, 'peliasAdmin') );
 query.filter( peliasQuery.view.boundary_circle );
 query.filter( peliasQuery.view.boundary_rect );
 query.filter( peliasQuery.view.sources );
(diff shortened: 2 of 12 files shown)


trescube:
:shipit:

orangejulius:
I rebased this PR after merging https://github.com/pelias/query/pull/31 and https://github.com/pelias/api/pull/590, and the tests now pass!

LGTM

PR #585, opened 2016-07-12
remove cluster2 module

missinglink:
remove the `cluster2` module

the package dependency was linking to a branch on my fork, so happy to remove that, also its been abandoned for some time.

I actually didn't realize it would be so easy to replace.

note: the `cluster2` module came with a healthcheck http API, I assume we don't use that?

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,16 +1,28 @@
 
-var Cluster = require('cluster2'),
+var cluster = require('cluster'),
     app = require('./app'),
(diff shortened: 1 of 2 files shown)


trescube:
:shipit:

orangejulius:
LGTM, YAY!

We definitely have to remember to give this a good testing over on dev or dev2 (or prod_build at least) before deploying it to production.

PR #594, opened 2016-07-21
Add Czech Republic

dianashk:
Czech Republic addresses have housenumbers after street names, similar to Germany. This will add it to the list of countries where we need to handle that behavior.


trescube:
:shipit:

orangejulius:
LGTM

dianashk:
Hooray! 🌴 🌅

Issue #334, opened 2015-10-06
Allow override of default Elasticsearch index name via pelias config file


riordan:
_From @heffergm on September 22, 2015 10:54_

The solution to the schema change scenario that came up yesterday in chat: allow the ability to change the database name that the API uses to issue requests to the backend via the pelias config file (currently hardcoded(?) as ‘pelias’). 

This would allow a manually rollout when there’s a breaking schema change where the live API would continue to point at the correct index until everything is in place to move to the new one.

_Copied from original issue: pelias/schema#73_

orangejulius:
This is all done and has been successfully used in production to do a migration that would otherwise have required at least a little downtime! :tada:

PR #595, opened 2016-07-21
Allow Elasticsearch index name to be set from pelias-config

orangejulius:
The index name value will be read from pelias config, under the `config.api.indexName` property. https://github.com/pelias/config/pull/30 (released in pelias-config-2.1.0) sets the default for this value to `pelias`, so it will always be defined and will have the same behavior as before with no changes in `~/pelias.json`.

I've tested this locally quite a bit in addition to the unit tests which should cover everything, but am especially curious to know if there's anything I missed.

Connects https://github.com/pelias/api/issues/334

### DIFF
--- a/controller/place.js
+++ b/controller/place.js
@@ -1,7 +1,7 @@
 var service = { mget: require('../service/mget') };
 var logger = require('pelias-logger').get('api:controller:place');
 
-function setup( backend ){
+function setup( config, backend ){
 
   // allow overriding of dependencies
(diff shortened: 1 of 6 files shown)


orangejulius:
Note: after merging to master I'll be merging this PR directly into the production branch, so that we can use it to adjust our indices in production

missinglink:
LGTM

trescube:
:shipit:

PR #598, opened 2016-07-22
Set up layer filtering for autocomplete and reverse

orangejulius:
This was missed by me when working on https://github.com/pelias/api/pull/580, but caught by the acceptance tests!

Unfortunately it was caught after going to production, so until this is merged and deployed we won't have any layer filtering in production(!).

### DIFF
--- a/query/autocomplete.js
+++ b/query/autocomplete.js
@@ -47,6 +47,7 @@ query.score( peliasQuery.view.population( views.pop_subquery ) );
 
 // non-scoring hard filters
 query.filter( peliasQuery.view.sources );
+query.filter( peliasQuery.view.layers );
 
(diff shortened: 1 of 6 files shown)


orangejulius:
For the record, here's the acceptance test query that caught this. Notice that production returned an address result for a query that filters to only venues.

screenshot from 2016-07-22 14-59-46
screenshot from 2016-07-22 14-59-58

trescube:
:shipit:

orangejulius:
Fixed now! [query](http://pelias.github.io/compare/#/v1/reverse%3Fpoint.lon=-73.989231&point.lat=40.744022&layers=venue)
screenshot from 2016-07-22 15-51-49

PR #599, opened 2016-07-25
Handle empty parent properties

orangejulius:
After https://github.com/pelias/model/pull/37, we have significantly cleaner Elasticsearch documents that don't have empty values for every possible field when there's no data. This is good.

However, some places in the API code were still relying on these empty arrays always being present. It takes some careful query crafting, but it's possible to cause the API to return a 500 error with the right combination of query and data.

I believe this PR fixes all cases where a property on the parent object in an Easticsearch document was assumed to be an array. These properties can now be undefined, so a quick check has to be added.


trescube:
:shipit:

dianashk:
LGTM

PR #607, opened 2016-07-28
fix: URL for Gitter does not appear correctly on NPM

orangejulius:
NPM's markdown parser doesn't work the same as Github's it seems. The space in the URL causes things to look bad:
screenshot from 2016-07-28 19-28-58


dianashk:
:shipit: good catch!

PR #608, opened 2016-07-29
wrap coordinates at poles

### LINKED ISSUE #570: handle coordinate wrapping
prior to elasticsearch2 the wrapping of longitudes was handled automagically inside elasticsearch, this is no longer true and the API now responds:

```
[query_parsing_exception] illegal longitude value [286.00536346435547] for [geo_distance], with { index=pelias line=1 col=178 }
```

note there are acceptance tests in a file called `reverse_coordinate_wrapping` marked as failing until this issue is resolved.

missinglink:
wrap coordinates at poles.

see tests for examples.

Fixes #570

### DIFF
--- a/test/unit/sanitiser/_geo_common.js
+++ b/test/unit/sanitiser/_geo_common.js
@@ -55,7 +55,7 @@ module.exports.tests.wrapping = function(test, common) {
     };
     sanitize.sanitize_point( 'point', clean, params, false );
     t.equal(clean['point.lat'], +88.9, 'equal to (+90 - 1.1)');
-    t.equal(clean['point.lon'], -178.9, 'equal to (-180 - 1.1)'); // polar flip
+    t.equal(clean['point.lon'], -178.9, 'equal to (-180 + 1.1)'); // polar flip
     t.end();
   });
   test('negative latitudinal wrapping', function (t) {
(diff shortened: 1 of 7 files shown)


test/unit/sanitiser/_geo_common.js
```diff
@@ -12,6 +12,64 @@ module.exports.tests.interface = function(test, common) {
...
+    sanitize.sanitize_point( 'point', clean, params, false );
+    t.equal(clean['point.lat'], params['point.lat']);
+    t.equal(clean['point.lon'], params['point.lon']);
+    t.end();
+  });
+  test('positive longitude wrapping', function (t) {
+    var clean = {};
+    var params = {
+      'point.lat': 40.7,
+      'point.lon': 195.1
```
orangejulius: Would it be clearer if this was 286.1 and the expected value was -73.9 like the control test?

orangejulius:
LGTM!

_BUT_ I think it would be better if the expected value for all the tests were the same as the control, to make it more clear that the wrapping makes all the different values behave indentically. **Disclaimer**: I believe the acceptance-tests do not currently follow this guideline, I wrote the acceptance tests, and I mentioned that it would make sense to base the unit tests for wrapping off the acceptance tests. :) I'm happy to update the acceptance tests to reflect this change as well :P

trescube:
:shipit:

dianashk:
LGTM

trescube:
i'm typing for the heck of it!

PR #612, opened 2016-08-01
Add borough as a possible layer for geonames

avulfson17:
Fixes pelias/geonames#99

### DIFF
--- a/test/unit/sanitiser/_sources_and_layers.js
+++ b/test/unit/sanitiser/_sources_and_layers.js
@@ -53,6 +53,17 @@ module.exports.tests.no_errors = function(test, common) {
 
 test('valid combination', function(t) {
     var raw = {};
+    var clean = { sources: ['geonames'], layers: ['borough'] };
+
+    var messages = sanitize(raw, clean);
+
+    t.equal(messages.errors.length, 0, 'should return no errors');
+    t.equal(messages.warnings.length, 0, 'should return no warnings');
+    t.end();
+  });
+
+  test('valid combination', function(t) {
+    var raw = {};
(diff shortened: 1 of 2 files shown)


orangejulius:
What, no tests? :)

avulfson17:
I think there is a test for it 
`test('alias layer mapping', function(t) {
    t.deepEquals(type_mapping.layer_mapping.coarse,
                 [ 'continent', 'country', 'dependency', 'macroregion',
                   'region', 'locality', 'localadmin', 'macrocounty', 'county', 'macrohood',
                   'borough', 'neighbourhood', 'microhood', 'disputed' ]);
    t.end();
  });`

orangejulius:
Well, there's only one file changed, and it's the implementation code. The test didn't fail in the past, so it couldn't have covered the case you added :P

The test you copied is for the `coarse` alias. I think you want to add another test like this one for `borough`: https://github.com/pelias/api/blob/master/test/unit/sanitiser/_sources_and_layers.js#L54-L63

test/unit/sanitiser/_sources_and_layers.js
```diff
@@ -53,7 +53,7 @@ module.exports.tests.no_errors = function(test, common) {
 
 test('valid combination', function(t) {
     var raw = {};
-    var clean = { sources: ['geonames'], layers: ['macroregion'] };
+    var clean = { sources: ['geonames'], layers: ['borough'] };
```
orangejulius: please duplicate this test, keeping the macroregion one, and adding a new borough one :P

dianashk:
LGTM

PR #614, opened 2016-08-03
cleaner code: replace while() statements

missinglink:
it occurred to me while brushing my teeth this morning that the `while()` loops were not necessary :)


trescube:
:shipit:

orangejulius:
Please brush your teeth more often :shipit:

missinglink:
I get more time to think when I floss

PR #617, opened 2016-08-03
Updated warning for boundary.circle

avulfson17:
Fixes (or at least modifies) #510

### DIFF
--- a/test/unit/sanitiser/_geo_reverse.js
+++ b/test/unit/sanitiser/_geo_reverse.js
@@ -16,7 +16,7 @@ module.exports.tests.sanitize_boundary_country = function(test, common) {
     t.equals(clean['boundary.circle.lat'], 12.121212, 'should be set to point.lat');
     t.deepEquals(errorsAndWarnings, {
       errors: [],
-      warnings: ['boundary.circle is currently unsupported']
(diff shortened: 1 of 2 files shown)


orangejulius:
LGTM!

test/unit/sanitiser/_geo_reverse.js
```diff
@@ -50,7 +50,7 @@ module.exports.tests.sanitize_boundary_country = function(test, common) {
     // t.equals(clean['boundary.circle.radius'], 12.121212, 'should be set to point.lat')
     t.deepEquals(errorsAndWarnings, {
       errors: [],
-      warnings: ['boundary.circle is currently unsupported']
+      warnings: []
```
dianashk: this test is explicitly checking for a warning that `boundary.circle.radius` is going to be ignore, which it is. the code changes you've made don't (or shouldn't) change that behavior. this test is actually proving that the code is broken.

dianashk: sorry, ignore the first comment. i was reading this wrong.

however, the name of the test should be updated to reflect what the new expectation is.

dianashk:
:shipit:

Issue #601, opened 2016-07-27
Single shared instance of text-analyzer


trescube:
For pelias/pelias#152, we'll need a singleton-ish instance of text-analyzer to share because the API is multicore.  Potential solutions (but not limited to):

- microservice
- parent/child messaging like in WoF PiP lookup

trescube:
Decided to remove multicore support instead

PR #625, opened 2016-08-05
Initial libpostal integration

trescube:
This is the big PR to incorporate libpostal functionality.  It incorporates several large changes of functionality:

- query/search.js - disables all scoring temporarily, switches to FallbackQuery/GeodisambiguationQuery
- middleware/trimByGranularity.js - choose most granular results when FallbackQuery was used

Because addressIt is still needed for the /autocomplete endpoint, a separate query/texetparser.js was added just for it.  For a discussion on what FallbackQuery and Geodisambiguation, see:

- https://github.com/pelias/query/blob/master/layout/FilteredBooleanQuery.js
- https://github.com/pelias/query/blob/master/layout/GeodisambiguationQuery.js


index.js
```diff
@@ -2,7 +2,7 @@
 var cluster = require('cluster'),
     app = require('./app'),
     port = ( process.env.PORT || 3100 ),
-    multicore = true;
+    multicore = false;
```
missinglink: this should be `true`?

middleware/trimByGranularity.js
```diff
@@ -0,0 +1,69 @@
...
+// FallbackQuery was used.  FallbackQuery is used for inputs like
+// `1090 N Charlotte St, Lancaster, PA` where the address may not exist and
+// we must fall back to trying `Lancaster, PA`.  If the address does exist then
+// FallbackQuery will return results for:
+// - address+city+state
+// - city+state
+// - state
+//
+// Because the address matched, we're not interested in city+state or state, so
+// this component removes that aren't the most granular.
```
missinglink: this component removes [?] that aren't the most granular

package.json
```diff
@@ -54,8 +55,10 @@
     "pelias-logger": "0.0.8",
     "pelias-model": "4.1.0",
     "pelias-query": "8.3.0",
+    "pelias-suggester-pipeline": "2.0.4",
```
missinglink: I recently deprecated this, are you sure it's required?

package.json
```diff
@@ -54,8 +55,10 @@
...
     "pelias-text-analyzer": "1.3.0",
     "stats-lite": "2.0.3",
+    "stats-lite": "1.0.3",
```
missinglink: package `stats-lite` is listed twice

query/search.js
```diff
@@ -118,7 +117,41 @@ function generateQuery( clean ){
     textParser( clean.parsed_text, vs );
   }
 
-  return query.render( vs );
+  var q = getQuery(vs);
+
+  console.log(JSON.stringify(q, null, 2));
```
missinglink: debug

query/search.js
```diff
@@ -118,7 +117,41 @@ function generateQuery( clean ){
...
+  }).length === 1;
+}
+
+function hasQueryOrAddress(vs) {
+  return ['housenumber', 'street', 'query', 'category'].filter(function(layer) {
+    return vs.isset('input:' + layer);
+  }).length > 0;
+}
+
+function getQuery(vs) {
```
missinglink: `function getQuery` defined twice

sanitiser/search.js
```diff
@@ -8,7 +8,7 @@ var sanitizeAll = require('../sanitiser/sanitizeAll'),
       size: require('../sanitiser/_size')(/* use defaults*/),
       layers: require('../sanitiser/_targets')('layers', type_mapping.layer_mapping),
       sources: require('../sanitiser/_targets')('sources', type_mapping.source_mapping),
-      // depends on the layers and sources sanitisers, must be run after them
+      // // depends on the layers and sources sanitisers, must be run after them
```
missinglink: ?

sanitiser/search.js
```diff
@@ -17,10 +17,6 @@ var sanitizeAll = require('../sanitiser/sanitizeAll'),
 
 var sanitize = function(req, cb) { sanitizeAll(req, sanitizers, cb); };
 
-// export sanitize for testing
-module.exports.sanitize = sanitize;
-module.exports.sanitiser_list = sanitizers;
-
```
missinglink: is this no longer required for testing?

index.js
trescube: Ideally yes but waiting on https://github.com/pelias/api/issues/601

sanitiser/search.js
trescube: It's no longer required since the old tests are just re-testing the already-tested individual sanitisers.  The new test proxyquire's all the component sanitisers and verifies that they were called and in the correct order.  See https://github.com/pelias/api/blob/2744244f628627a3e84b23d4e39f83ddce23b4dd/test/unit/sanitiser/search.js#L9

orangejulius:
Looks like the final commit, a merge of a branch on github that you had probably rebased locally (https://github.com/pelias/api/pull/625/commits/2744244f628627a3e84b23d4e39f83ddce23b4dd), both duplicated commit history and added a few unwanted changes (like adding stats-lite as a dependency twice). We can fix it whenever you have a few minutes

vesameskanen:
I did a quick test with this branch, and found out that house number analysis is now working quite well, and we no longer get streets from wrong cities. However, I am quite concerned about the character conversion which libpostal does: ä -> ae etc. This breaks big part of searches. Any ideas how to fix this?

thatdatabaseguy:
@vesameskanen the libpostal models need to do those normalizations at some level so länderstraße, laenderstrasse, and landerstrasse can share statistical parameters.

I'm not sure if Pelias is handling umlaut transliteration now (looks like it mostly does accent stripping, "Muenchen" for instance does not return Munich but "Munchen" does), but it should be possible to use libpostal without breaking any such searches by either indexing the transliterated forms from libpostal, or using the compound analyzer approach detailed [here](http://elasticsearch-users.115913.n3.nabble.com/Folding-German-characters-like-umlauts-td2176078.html) or [here](https://gist.github.com/jprante/4239954). From what I understand, Dutch and the Nordic languages usually drop the umlaut when writing on an ASCII keyboard, etc. with two exceptions in Swedish : ü => y and å => aa, but these are a little less frequently used.

There has been some discussion in the libpostal repo about returning only segments of the original input without any lowercasing or transliteration (or only performing those normalizations on some internal representation) for people looking to display the parser results in some way. It's a little complicated to do that in the current implementation of the transliterator, and not a huge priority for search, but it's on the radar.

vesameskanen:
Hi @thatdatabaseguy, many thanks for your reply and also for the great libpostal tool which seems to parse also finnidh addresses amazingly well.
I was already suspecting that the consiquences of normalization need to be addresses on Pelias side.

vesameskanen:
OK, found out the following related pull requests:

https://github.com/pelias/schema/pull/146
https://github.com/pelias/schema/pull/149

I guess a solution is under way :)

PR #632, opened 2016-08-12
Removed pelias-text-analyzer requirement in test

trescube:
Replaced called to pelias-text-analyzer (which could be libpostal or addressit) with hard-coded addressit responses.


orangejulius:
LGTM. Funny that we were requiring a different module just to generate test data. clearly an improvement :)

dianashk:
yeah, much cleaner this way! :shipit:

PR #633, opened 2016-08-12
exclude `name` for `country` result labels

trescube:
Fixes pelias/wof-admin-lookup#49

### DIFF
--- a/helper/labelGenerator.js
+++ b/helper/labelGenerator.js
@@ -47,8 +47,7 @@ function getSchema(country_a) {
 
 }
 
-// helper function that sets a default label for non-US/CA regions
-// this is a very special case
+// helper function that sets a default label for non-US/CA regions and countries
 function getInitialLabel(record) {
   if (isRegion(record.layer) &&
       isGeonamesOrWhosOnFirst(record.source) &&
--- a/helper/labelGenerator.js
+++ b/helper/labelGenerator.js
@@ -56,8 +56,18 @@ function getInitialLabel(record) {
     return [];
   }
 
+  if (isCountry(record.layer)) {
+    return [];
+  }
+
   return [record.name];
 
+}
+
+// this can go away once geonames is no longer supported
+// https://github.com/pelias/wof-admin-lookup/issues/49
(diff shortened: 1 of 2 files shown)


helper/labelGenerator.js
```diff
@@ -56,10 +56,20 @@ function getInitialLabel(record) {
     return [];
```
orangejulius: the comment above probably has to be updated

test/unit/helper/labelGenerator_default.js
```diff
@@ -175,14 +175,14 @@ module.exports.tests.default_country = function(test, common) {
...
     var doc = {
-      'name': 'country name',
+      'name': 'source country name',
       'layer': 'country',
       'country_a': 'country code',
-      'country': 'country name'
+      'country': 'hierarchy country name'
     };
-    t.equal(generator(doc),'country name');
+    t.equal(generator(doc),'hierarchy country name');
```
orangejulius: I'm kinda concerned that we are taking the hierarchy country, and not the name of the record itself. Not really because of the effect it has, but just because of the precedent it sets (or has already been set and we are now continuing). In any case it may not matter if we decide to get rid of geonames in the near future

trescube: We could use the `name` field but that would generate a label that we wouldn't be able to search for again.  That is, geonames' name for `China` is `The People's Republic of China`.  While libpostal understands it to be a country, we won't find anything since we only import the WOF name `China`.

orangejulius:
Despite my concerns, this is a huge improvement and it will be nice to finally knock out https://github.com/pelias/wof-admin-lookup/issues/49. LGTM

test/unit/helper/labelGenerator_default.js
dianashk: I think being able to search for a record using the label is important so in the interest of preserving that we should keep it this way for now. I imagine a lot of this will change when we start looking at Alt-Names and internationalization, so we can revisit it then and figure out the "right" approach.

dianashk:
:shipit:

PR #634, opened 2016-08-12
Add boundary.country filter to /v1/autocomplete

orangejulius:
It works exactly the same as in `/search`, no surprises here!

Connects https://github.com/pelias/pelias/issues/403


orangejulius:
It works!
screenshot from 2016-08-12 16-35-53

trescube:
:shipit:

dianashk:
LGTM Wow that was quick!

missinglink:
cc/ @campreb

missinglink:
in production now @campreb

eg:
```
/v1/autocomplete?boundary.country=NZ&text=shortland st, auckland
```

PR #636, opened 2016-08-17
Add old calls for fallback

### LINKED ISSUE #602: Fallback to existing behavior when new queries return nothing
The libpostal integration will be iterative as we work out the kinks, but for first release, if neither FallbackQuery nor GeodisambiguationQuery return anything then the API should fallback to current production behavior using FilteredBooleanQuery.

The impetus is that it's entirely possible for either new query module to not return anything.  For example, the input `United States of America` gets parsed by libpostal as:

```
> united states of america

Result:

{
  "country": "united states of america"
}
```

Whereas we only index [United States and USA](https://whosonfirst.mapzen.com/spelunker/id/85633793/#2/52.3/0.3) so the search fails.   

While not optimal, the API should query again using FilteredBooleanQuery to at least return something rather than nothing.

trescube:
This PR adds support for using the current ES querying strategy if the new libpostal integrated strategy doesn't return any results. 

The TL;DR is to call addressit if there are 0 results in `res.data` by introducing a sanitizing layer between ES calls for new and current production querying strategies.  The sanitizing layer and query strategies have early exit conditions.  

I had incorrectly created this PR before based on `master` instead of `initial-libpostal-integration` so I closed and recreated with the correct base.

Fixes #602


routes/v1.js
```diff
@@ -63,7 +64,12 @@ function addRoutes(app, peliasConfig) {
     search: createRouter([
       sanitisers.search.middleware,
       middleware.calcSize(),
-      controllers.search(peliasConfig),
+      // 2nd parameter is `backend` which gets initialized internally
+      // 3rd parameter is which query module to use, use fallback/geodisambiguation
+      //  first, then use original search strategy if first query didn't return anything
+      controllers.search(peliasConfig, undefined, require('../query/search')),
```
dianashk: if you could move the `require` statements from line 70 and 72 into an object called `search_strategy` or something to that effect, it would be more in line with the rest of this file and make it easier to read the code.

controller/search.js
```diff
@@ -17,6 +17,13 @@ function setup( config, backend, query ){
       return next();
     }
 
+    // do not run controller if there are already results
+    // this was added during libpostal integration.  if the libpostal parse/query
+    // doesn't return anything then fallback to old search-engine-y behavior
+    if (res && res.hasOwnProperty('data') && res.data.length > 0) {
+      return next();
+    }
+
```
dianashk: would be great to add some logging here to let us know that we used the original after all, so we can do some analysis on how often this happens and for which queries. that way we can later determine if turning it off is the better way to go. also would be good to set a flag of some sort in the request or result objects so that we can use it in confidence score computation. generally if we used this `original` behavior it's safe to assume that confidence in the results will be lower than usual.

dianashk:
This is looking good. Just a few comments from me before it's ready for primetime.

controller/search.js
orangejulius: I agree some logging would be nice. Note that we can't log queries themselves when users have DNT enabled. There is some code in the [middleware](https://github.com/pelias/api/blob/master/middleware/access_log.js#L32-L34) to help filter out sensitive query params when DNT is enabled that might be helpful.

orangejulius:
LGTM

controller/search.js
trescube: I added the logging [https://github.com/pelias/api/blob/4032578e29dbb6b0637ebd1b2c23969f39acd059/sanitiser/search_fallback.js#L19-L21](here) since that's the easiest place to determine that there actually was a fallback.

dianashk:
:+1:

sanitiser/search_fallback.js
```diff
@@ -0,0 +1,27 @@
...
+  // if res.data already has results then don't call the _text_autocomplete sanitiser
+  // this has been put into place for when the libpostal integration way of querying
+  // ES doesn't return anything and we want to fallback to the old logic
+  if (res && res.hasOwnProperty('data') && res.data.length > 0) {
+    return next();
+  }
+
+  // log the query that caused a fallback since libpostal+new-queries didn't return anything
+  if (req.path === '/v1/search') {
+    logger.info(req.clean.text);
```
orangejulius: We can't unconditionally log `clean.txt` because it would violate the promise of eraser map to not log sensitive information. We can only log it if the user does not have [DNT](https://en.wikipedia.org/wiki/Do_Not_Track) enabled. You can use the `isDNT` [method](https://github.com/pelias/api/blob/master/helper/logging.js#L4) in the logging helper to check for it.

orangejulius: I would change this code to something like


```js
if (req.path === '/v1/search') {
    var queryText = logging.isDNT(req) ? '[text removed]' : req.clean.text;
    logger.info("falling back to original search for query text " + queryText);
}
```

PR #638, opened 2016-08-17
Add provinces to Australian labels

### LINKED ISSUE #615: Add region to AUS labels
Regions should be added to Australian labels for disambiguation.

avulfson17:
Fixes #615

### DIFF
--- a/helper/labelSchema.js
+++ b/helper/labelSchema.js
@@ -13,17 +13,17 @@ module.exports = {
   'USA': {
     'borough': getFirstProperty(['borough']),
     'local': getFirstProperty(['locality', 'localadmin', 'county']),
-    'regional': getUsOrCaOrAusState,
+    'regional': getRegionalValue,
     'country': getUSACountryValue
   },
   'AUS': {
     'local' : getFirstProperty(['locality', 'localadmin']),
-    'regional' : getUsOrCaOrAusState,
+    'regional' : getRegionalValue,
     'country': getFirstProperty(['country'])
   },
   'CAN': {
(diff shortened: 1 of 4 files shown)


helper/labelSchema.js
```diff
@@ -13,12 +13,17 @@ module.exports = {
   'USA': {
     'borough': getFirstProperty(['borough']),
     'local': getFirstProperty(['locality', 'localadmin', 'county']),
-    'regional': getUsOrCaState,
+    'regional': getUsOrCaOrAusState,
```
orangejulius: after how many countries do we rename this to something like `getRegionalValue`? 3 feels like a good time :)

helper/labelSchema.js
```diff
@@ -13,12 +13,17 @@ module.exports = {
...
     'country': getUSACountryValue
   },
+  'AUS': {
+    'local' : getFirstProperty(['locality', 'localadmin']),
+    'regional' : getUsOrCaOrAusState,
```
avulfson17: is there a reason that this function isn't called here? why does it just store the variable name?

orangejulius:
LGTM now!

trescube:
:shipit:

helper/labelSchema.js
orangejulius: There is a reason! It's a bit confusing. The labelGenerator.js module which actually generates the labels loads this module, and [picks the correct country's schema to use](https://github.com/pelias/api/blob/master/helper/labelGenerator.js#L41-L48), and then calls [the function from that schema](https://github.com/pelias/api/blob/master/helper/labelGenerator.js#L12-L13)

PR #644, opened 2016-08-19
added support for `dnt` flag when logging fallback query

trescube:
I didn't include the text "falling back to original search for query text" so that logs are easier to parse.  Lines for log entries are prefixed with `[api:controller:search_fallback]`.


orangejulius:
Oh, right. I forgot there would be a prefix. LGTM

dianashk:
:+1:

PR #649, opened 2016-09-03
Refactor deduper and write additional tests

### LINKED ISSUE #604: Revisit deduper for correct operation
libpostal integration, by way of introducting 2 new query modules, may end up creating result sets that the deduper may not deal with correctly.  Then again, it may operate as expected.  Either way, as part of the https://github.com/pelias/pelias/issues/152, the deduper should be revisited to ensure correct operation.

dianashk:
Refactored the deduper. Here's what got done:

- Break out the diff functions to make testing them easier
- Add tests for the newly separated diff functions
- In the diff function, compare the `layer`, `parent.*_id`, `address_parts.*`, and `name.*`
   - Only compare `zip` if both records have it set, otherwise ignore
   - When comparing the `parent.*_id` properties, skip the check where *===`item.layer`
   - When comparing `name.*` only expect `name.default` to match, but only check additional languages if both items have it set, otherwise ignore

Fixes #604

### DIFF
--- a/helper/diffPlaces.js
+++ b/helper/diffPlaces.js
@@ -1,0 +1,172 @@
+var _ = require('lodash');
+var placeTypes = require('../helper/placeTypes');
+
+/**
+ * Compare the layer properties if they exist.
+ * Returns false if the objects are the same, and throws
+ * an exception with the message 'different' if not.
+ *
+ * @param {object} item1
+ * @param {object} item2
+ * @returns {boolean}
+ * @throws {Error}
+ */
+function isDiffLayer(item1, item2) {
+  if (item1.layer === item2.layer) {
+    return false;
+  }
+
+  throw new Error('different');
+}
+
+/**
+ * Compare the parent.* properties if they exist.
+ * Returns false if the objects are the same, and throws
+ * an exception with the message 'different' if not.
+ *
+ * @param {object} item1
+ * @param {object} item2
+ * @returns {boolean}
+ * @throws {Error}
+ */
+function isDiffParentHierarchy(item1, item2) {
+  // if neither object has parent, assume same
+  if (!item1.hasOwnProperty('parent') && !item2.hasOwnProperty('parent')) {
+    return false;
+  }
+
+  // if both have parent, do the rest of the checking
+  if (item1.hasOwnProperty('parent') && item2.hasOwnProperty('parent')) {
+    placeTypes.forEach(function (placeType) {
+      // don't consider its own id
+      if (placeType === item1.layer) {
+        return;
+      }
+      propMatch(item1.parent, item2.parent, placeType + '_id');
+    });
+    return false;
+  }
+
+  // if one has parent and the other doesn't consider different
+  throw new Error('different');
+}
+
+/**
+ * Compare the name.* properties if they exist.
+ * Returns false if the objects are the same, and throws
+ * an exception with the message 'different' if not.
+ *
+ * @param {object} item1
+ * @param {object} item2
+ * @returns {boolean}
+ * @throws {Error}
+ */
+function isDiffName(item1, item2) {
+  if (item1.hasOwnProperty('name') && item2.hasOwnProperty('name')) {
+    for (var lang in item1.name) {
+      if(item2.name[lang] || lang === 'default') {
+        // do not consider absence of an additional name as a difference
+        propMatch(item1.name, item2.name, lang);
+      }
+    }
+  }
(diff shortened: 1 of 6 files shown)


dianashk:
TBD: I'd like to add a function in the deduper middleware to check which of the two items is prefered when a dupe is found. This would allow us to pick WOF records over Geonames, and OA over OSM, etc.

helper/diffPlaces.js
```diff
@@ -0,0 +1,172 @@
...
+ * Compare the layer properties if they exist.
+ * Returns false if the objects are the same, and throws
+ * an exception with the message 'different' if not.
+ *
+ * @param {object} item1
+ * @param {object} item2
+ * @returns {boolean}
+ * @throws {Error}
+ */
+function isDiffLayer(item1, item2) {
```
missinglink: the method signature of this function is not very intuitive, it's called `isDiffLayer` which I think is a great name if it returns a boolean value.

looking at the calling code it seems as though the return value is not actually used and the functions main purpose is validation.

a bit nit-picky, but would this function name be better as `validateDiffLayer` (which returns nothing and throws on error) **or** `isDiffLayer` (which does not throw and returns bool so the caller can handle the error case)?

helper/diffPlaces.js
```diff
@@ -0,0 +1,172 @@
+var _ = require('lodash');
+var placeTypes = require('../helper/placeTypes');
```
missinglink: nit pick: `'./placeTypes'`?

helper/diffPlaces.js
```diff
@@ -0,0 +1,172 @@
...
+function isDiffLayer(item1, item2) {
+  if (item1.layer === item2.layer) {
+    return false;
+  }
+
+  throw new Error('different');
+}
+
+/**
+ * Compare the parent.* properties if they exist.
```
missinglink: The code only checks `parent.*_id` properties, not **all** `parent.*` props.

> parent.* properties if...
> parent.* property ids if..

helper/diffPlaces.js
```diff
@@ -0,0 +1,172 @@
...
+ *
+ * @param {object} item1
+ * @param {object} item2
+ * @returns {boolean}
+ * @throws {Error}
+ */
+function isDiffName(item1, item2) {
+  if (item1.hasOwnProperty('name') && item2.hasOwnProperty('name')) {
+    for (var lang in item1.name) {
+      if(item2.name[lang] || lang === 'default') {
```
missinglink: nit pick: some parts of the code use `obj.hasOwnProperty('attr')` and some `obj[attr]`

helper/diffPlaces.js
```diff
@@ -0,0 +1,172 @@
...
+
+/**
+ * Remove punctuation and lowercase
+ *
+ * @param {string} str
+ * @returns {string}
+ */
+function normalizeString(str) {
+  if (!str) {
+    return '';
```
missinglink: this line could be the source of bugs, returning an empty string for all falsy values means that those values are always considered equal.

eg. `normalizeString(1) === normalizeString(null) // true`

helper/diffPlaces.js
```diff
@@ -0,0 +1,172 @@
...
+  }
+  return str.toLowerCase().split(/[ ,-]+/).join(' ');
```
missinglink: this code will throw for all non-string types, such as `normalizeString(1)`

test/unit/fixture/dedupe_elasticsearch_results.js
```diff
@@ -1,377 +1,1192 @@
...
+        'United States'
+      ],
+      'macrocounty_a': [],
+      'locality_a': [
+        null
+      ],
+      'county': [
+        'Lancaster County'
+      ],
+      'borough_a': [],
```
missinglink: I thought we got rid of all these empty arrays coming back from ES?

If you're using `pelias/model` to generate them you can use `doc.toESDocument()` to strip empty arrays from the output.

dianashk:
Thanks @missinglink for the helpful feedback. Right on and not at all nitpicky. 👍 I think I've address all your comments.

trescube:
Very helpful description and I think this will solve our deduping issues.  :shipit:

missinglink:
:shipit:

dianashk:
Thanks all for feedback. Merging this and will create a separate issue for the `isPreferred` functionality mentioned in a comment above. 🚀

PR #655, opened 2016-09-07
added special override condition for libpostal query/housenumber mixup

trescube:
libpostal sometimes parses addresses with prefix house numbers in places where the house number is normally postfix incorrectly, for instance:

```
> 1 Grolmanstraße, Berlin, Germany

Result:

{
  "house": "1",
  "road": "grolmanstrasse",
  "state": "berlin",
  "country": "germany"
}
```

as compared to postfix house number:

```
> Grolmanstraße 1, Berlin , Germany

Result:

{
  "road": "grolmanstrasse",
  "house_number": "1",
  "city": "berlin",
  "country": "germany"
}
```

In libpostal parlance, `house` is just a query term, not the house number.  This special case moves the query term to the house number field if there's a street, there's no house number, and the query is parseable as an integer, then use the query as the house number and blank out the query.

### DIFF
--- a/query/text_parser.js
+++ b/query/text_parser.js
@@ -1,4 +1,5 @@
 var logger = require('pelias-logger').get('api');
+var _ = require('lodash');
 
 // all the address parsing logic
 function addParsedVariablesToQueryVariables( parsed_text, vs ){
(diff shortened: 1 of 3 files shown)


query/text_parser.js
```diff
@@ -1,4 +1,5 @@
 var logger = require('pelias-logger').get('api');
+var _ = require('lodash');
```
dianashk: don't see where this is used in the new code

missinglink:
oh that's odd, did you know about this @thatdatabaseguy?

due to all the english speakers in Berlin (like me!?) both these forms are common here; less so for 'real Germany'.

thatdatabaseguy:
@missinglink Anglophiles...amiright? There's a way of handling things like this in the new libpostal [formatting config](https://github.com/openvenues/libpostal/blob/parser-data/resources/formatting/global.yaml). This config allows the parser to be trained on structural edge cases and exceptions to the standard per-country formats in [OpenCage's address-formatting repo] (https://github.com/OpenCageData/address-formatting). It specifies a bunch of swaps to be made in the address formats, randomly with certain probabilities, when building the training addresses. So e.g. one can say: in continental European formats (which, with the exception of France, all write house_number after road), move house_number before road name 1% of the time. Then the next time the training data are built, there will be hundreds of thousands of addresses like "1 Grolmanstraße" including cases where there's a venue name as well. This makes the parser more robust to structural changes.

dianashk:
LGTM

orangejulius:
:shipit:

PR #657, opened 2016-09-08
Add non-override conditions to sanitizeall

trescube:
The libpostal integration adds a fallback sanitize step between query calls that calls addressit.  The problem is that sanitizeAll initializes `req.clean` even if it already exists, so the results of all the sanitizers called before were being dropped.  This PR preserves `req.clean`, `req.errors`, and `req.warnings` if they have already been initialized.  

Also, there were no direct tests for sanitizeAll, it was being tested indirectly thru other tests.


orangejulius:
LGTM

dianashk:
:+1:

PR #660, opened 2016-09-08
disable calling GeodisambiguationQuery

trescube:
This PR disables calling the GeodisambiguationQuery by having query/search.js return `undefined` in the geodisambiguation criteria case.  controller/search.js then skips execution if the query to ES is `undefined`.  **With all the tests!**


orangejulius:
LGTM

dianashk:
:+1:

Issue #652, opened 2016-09-07
Label generation as a separate postprocessing step?


vesameskanen:
Would it make sense to create an explicit postprocessing stage where labels are added to the search results? Now the labels are created as a part of json conversion. Deduper and label generation resolutions do not match, so one often gets identical labels from searches.  I believe that client apps might often want to use Pelias generated labels directly without additional logic in the client side.  In such a case, identical labels are a problem. 

A dedicated module for label generation would allow more logic, where duplicate labels could be expanded on need basis, by adding configured extensions such as the address to a venue, a category description etc.

missinglink:
@trescube spent a bunch of time on label generation

dianashk:
This makes a lot of sense to me. We've struggled with how to make duplicate labels for non-duplicate things show up more clearly. This is an excellent suggestion.

orangejulius:
I like this suggestion as well. Right now our labels are purely a function of the data contained in the record, but there are definitely cases where we want to further disambiguate, like https://github.com/pelias/pelias/issues/317. Building labels with knowledge of other records could help us figure out how much data needs to be added to disambiguate.

missinglink:
I am :+1: for this

PR #661, opened 2016-09-10
fix labels for dependencies for USA

### LINKED ISSUE #628: US dependency labels are weird
Our label generation creates confusing labels for things in US dependencies.  For example, "Pago Pago, American Samoa" is analyzed and queried for as a city/country.  However, the label that the API generates is:  `Pago Pago, Eastern, USA`.  It should probably be `Pago Pago, Eastern, AS` or something that calls out the dependency over, or in addition to, the country.

trescube:
Treats existence of dependency fields as more important than country fields.  Examples:

`San Juan, PR` (abbreviation exists)
`Aasu, AS` (abbreviation exists)
`Altona, United States Virgin Islands` (abbreviation doesn't exist)
`Puerto Rico` (result was `dependency` layer)

This change also excludes `region` if `dependency` fields exist, for example, [San Juan](https://whosonfirst.mapzen.com/spelunker/id/101919427/#11/18.4050/-66.0604) is in the region of [San Juan](https://whosonfirst.mapzen.com/spelunker/id/85687331/#9/18.2001/-66.5836).  Without this exclusion, the label would be `San Juan, PR, PR`.  This exclusion shortens the label to `San Juan, PR`.  

Fixes #628 

While this fixes the source issue, the example in #628 is not fixed because [Pago Pago](https://whosonfirst.mapzen.com/spelunker/id/101734459/#12/-14.2512/-170.7168) has a broken hierarchy.

### DIFF
--- a/helper/labelSchema.js
+++ b/helper/labelSchema.js
@@ -14,7 +14,7 @@ module.exports = {
     'borough': getFirstProperty(['borough']),
     'local': getFirstProperty(['locality', 'localadmin', 'county']),
     'regional': getRegionalValue,
-    'country': getUSACountryValue
+    'country': getUSADependencyOrCountryValue
   },
   'AUS': {
     'local' : getFirstProperty(['locality', 'localadmin']),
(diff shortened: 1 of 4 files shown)


dianashk:
:shipit:

orangejulius:
LGTM

orangejulius:
This PR looks pretty safe. Do we want to merge it now or wait until after the dust is settled from libpostal?

missinglink:
if this change is just for labels, why are there changes to the properties generated in `test/unit/helper/geojsonify.js`?

trescube:
Because dependency has greater priority than country so I had to remove the dependency fields.  This will be fixed with #652 when label generation is injected.

dianashk:
are there corresponding acceptance tests? would be great to add them if not.

trescube:
Turns out there are acceptance tests for all placetypes *except* dependencies.  I'll add.

trescube:
Superseded by pelias/labels#4

PR #666, opened 2016-09-14
limited fallbackQuery usage to analysis with `street`

trescube:
As discussed, this change only allows the new query logic for analyzed text that contains a `street` field.

### DIFF
--- a/query/search.js
+++ b/query/search.js
@@ -120,25 +120,14 @@ function generateQuery( clean ){
 }
 
 function getQuery(vs) {
-  if (isSingleFieldGeoambiguity(vs) && !hasQueryOrAddress(vs)) {
-    // return `undefined` for now until we exorcise the geodisambiguation demons
-    return;
-  } else {
+  if (hasStreet(vs)) {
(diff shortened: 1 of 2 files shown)


trescube:
I also reworked the tests a little to call out the purpose of the object being passed in.

query/search.js
```diff
@@ -120,25 +120,14 @@ function generateQuery( clean ){
 }
 
 function getQuery(vs) {
-  if (isSingleFieldGeoambiguity(vs) && !hasQueryOrAddress(vs)) {
-    // return `undefined` for now until we exorcise the geodisambiguation demons
-    return;
-  } else {
```
orangejulius: For purely readability reasons, it might be nice to leave the else statement with something like

```javascript
} else {
  return undefined; // the old middleware will handle the request if undefined is returned here
}
```

orangejulius: For readability, it might be nice to leave the else, with an explicit `return undefined;` and a reminder comment of why it's happening.

PR #667, opened 2016-09-15
Require Node.js 4.0 or greater

orangejulius:
Getting started on https://github.com/pelias/pelias/issues/404 so we can merge https://github.com/pelias/api/pull/646

Connects https://github.com/pelias/pelias/issues/404

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ cache:
 notifications:
   email: false
 node_js:
-  - 0.12
   - 4
   - 6
 matrix:
--- a/README.md
+++ b/README.md
@@ -16,6 +16,8 @@ See the [Mapzen Search documentation](https://mapzen.com/documentation/search/).
 
 ## Install Dependencies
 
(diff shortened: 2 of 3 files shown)


missinglink:
:shipit: 

I'm :+1: for upgrading to benefit from the new features but we should also be cognizant that some people might be hosting the code in a shared environment where switching might be problematic.

orangejulius:
It's definitely right to not force people to upgrade unnecessarily. Node.js 0.12 will be fully unsupported after [December 31st, 2016](https://github.com/nodejs/LTS#lts_schedule), so we definitely shouldn't support it past then.

I threw out August 31st in pelias/pelias#404 as a proposed date for Pelias dropping support for 0.12, but really we could probably chose any time this year, and am definitely open to choosing another date if we think it's too soon. Otherwise though, let's get off 0.12 and throw a party where we start using ES6 features! :tada:

dianashk:
also LGTM

PR #670, opened 2016-09-16
Add accuracy property to results

dianashk:
Each record in results should have an `accuracy` property. Accuracy can be any of the following:
- `point`
- `interpolated`
- `centroid`

In this release `interpolated` is not handled as we currently don't support that.


middleware/accuracy.js
```diff
@@ -0,0 +1,57 @@
...
+ * Accuracy level should be set for each item in the results.
+ * The level can be any of the following:
+ *  - point
+ *  - interpolated
+ *  - centroid
+ */
+
+var check = require('check-types');
+
+var accuracyLevel_point = 'point';
```
missinglink: variable can't decide if he's a camelCase of a lowercase with underscores :)

trescube:
:shipit:

dianashk:
I rebased this against my `confidence-score-fix` branch so you'll now see all those changes here, but this way there shouldn't be conflicts when we merge this after the confidence score change is merged.

missinglink:
:shipit:

PR #674, opened 2016-09-21
Confidence score fix

dianashk:
The new confidence score computation is intentionally very simple. We only take into account whether or not there was an exact match, which is determined based on parsed_text of request and `layer` property of each individual result. If a fallback scenario was detected, the score is based on the granularity of the resulting layer. Confidence score computation for all other query types will remain unchanged at this time.

- fixes ciao tests for `initial-libpostal-integration` work 💰 
- adds a new module to compute `confidence` score for new libpostal-backed fallback query type
- adds a new `match_type` property for the new fallback queries that indicate whether there was an exact match or a fallback occurred
- adds unit and ciao tests where appropriate

Fixes pelias/pelias#335


trescube:
cleaner!  :shipit:

missinglink:
very nice :+1: :shipit:

PR #680, opened 2016-10-07
Inject label generator

### LINKED ISSUE #652: Label generation as a separate postprocessing step?

trescube:
This PR moves label generation logic out to https://github.com/pelias/labels, removes label generation from geojsonify and into a separate post-processing step with injected label generator function for unit testing.  Fixes #652.


middleware/geocodeJSON.js
```diff
@@ -11,7 +11,7 @@ var _ = require('lodash');
  * @param {string} [basePath]
  * @returns {middleware}
  */
-function setup(peliasConfig, basePath) {
+function setup(peliasConfig, basePath, labelGenerator) {
```
dianashk: I think that `labelGenerator` param is not needed (update corresponding unit test, too)

middleware/assignLabels.js
```diff
@@ -0,0 +1,23 @@
+function setup(labelGenerator) {
```
dianashk: I would prefer setting the default in here to be `pelias-labels` and only override if another label generator has been passed in for testing or whatever. In my opinion, that makes the client code cleaner.

routes/v1.js
```diff
@@ -88,6 +89,7 @@ function addRoutes(app, peliasConfig) {
       postProc.parseBoundingBox(),
       postProc.normalizeParentIds(),
       postProc.geocodeJSON(peliasConfig, base),
+      postProc.assignLabels(require('pelias-labels')),
```
dianashk: I think it makes more sense to generate labels before geojsonifying the results. It feels like the geojsonify step is like a final packaging up of the data, so adding properties afterwards could lead to those property not being packaged properly. Thoughts?

middleware/assignLabels.js
orangejulius: Yeah :+1: to having the default set within this module. Proxyquire seems to work pretty nice for testing, and keeps the interface much cleaner

middleware/assignLabels.js
```diff
@@ -1,6 +1,6 @@
 function setup(labelGenerator) {
   function middleware(req, res, next) {
-    return assignLabel(req, res, next, labelGenerator);
+    return assignLabel(req, res, next, labelGenerator || require('pelias-labels'));
```
orangejulius: Even though I think this setup function will only be called once, so this won't cause `pelias-labels` to be required too many times, it's probably less confusing to keep the require statement at the start of this file.

package.json
```diff
@@ -53,6 +53,7 @@
     "morgan": "1.7.0",
     "pelias-categories": "1.0.0",
     "pelias-config": "2.1.0",
+    "pelias-labels": "^1.0.1",
```
orangejulius: Oh, this should be an exact version so that we can track updates better

helper/geojsonify.js
```diff
@@ -1,18 +1,17 @@
...
 var extent = require('geojson-extent');
-var labelGenerator = require('./labelGenerator');
 var logger = require('pelias-logger').get('api');
 var type_mapping = require('./type_mapping');
 var _ = require('lodash');
 var addDetails = require('./geojsonify_place_details');
 var addMetaData = require('./geojsonify_meta_data');
 
-function geojsonifyPlaces( params, docs ){
+function geojsonifyPlaces( params, docs, labelGenerator ){
```
dianashk: labelGenerator is not used throughout this file anymore as far as i can tell. should be removed from here and lines 14, 37

routes/v1.js
```diff
@@ -87,6 +88,7 @@ function addRoutes(app, peliasConfig) {
       postProc.renamePlacenames(),
       postProc.parseBoundingBox(),
       postProc.normalizeParentIds(),
+      postProc.assignLabels(require('pelias-labels')),
```
dianashk: remove the require from here since we now have the default inside the labels module.

helper/geojsonify.js
trescube: I could've sworn that I removed that.

routes/v1.js
trescube: I know that the purpose of this is so that we have a default in case a label generator isn't passed in, but how do clients actually take advantage of this?  They can certainly pass in their own function for label generation, but doesn't that mean that they have to modify source?  And doesn't that make their upgrade pretty difficult?  Wouldn't it make more sense to make these defaults configuration-driven?

orangejulius: I don't think the intent of breaking out the label generation code into a separate repo is really to allow someone using the Pelias API to easily drop in their own label generation code. Instead it's to keep a really clear line of separation between the label generating code and everything else.

With that in mind it's nice to keep all knowledge of how to set up anything that _does_ require the label code confined to those modules. Right now the routes file needs to know not just to set up the label middleware, but what module to set it up with. I'd argue the label middleware should know how to set itself up.

routes/v1.js
```diff
@@ -119,6 +122,7 @@ function addRoutes(app, peliasConfig) {
       postProc.parseBoundingBox(),
       postProc.normalizeParentIds(),
       postProc.geocodeJSON(peliasConfig, base),
+      postProc.assignLabels(),
```
dianashk: it would be great if assignLabels could go first in all the endpoints, for consistency

package.json
```diff
@@ -53,6 +53,7 @@
     "morgan": "1.7.0",
     "pelias-categories": "1.1.0",
     "pelias-config": "2.3.0",
+    "pelias-labels": "^1.3.0",
```
orangejulius: This has to be an exact version (no caret) so that we can track all the changes in git commits.

PR #683, opened 2016-10-10
removed multicore support

trescube:


orangejulius:
Code looks good. Are we so sure no one wants this/we wont want to use it in the future that we feel the desire to remove it rather than just leave it disabled when libpostal is turned on?

dianashk:
If we don't support using it or recommend people turn it on in default operation I think it's best to remove it and avoid confusion. Those that know about multicore support will not have too much trouble adding it back in if needed.

orangejulius:
Sounds good. The code is definitely simpler this way, and just in general it will lead to less complexity in deployments. :shipit:

PR #684, opened 2016-10-12
hardcode size to 50 for FallbackQuery

trescube:
Fixes pelias/pelias#434. 

Previously, if the user specified a `size` parameter, Pelias would query ES for that many results.  This doesn't work well with the new fallback query model since an `address` layer document may match but scores lower than a fallback `city` layer document.  While the default `size` value is `10`, it was sufficiently high enough to return the `address` layer document.  

We should add an acceptance test or two that checks for this.

### DIFF
--- a/query/search.js
+++ b/query/search.js
@@ -59,7 +59,7 @@ function generateQuery( clean ){
 
   // size
   if( clean.querySize ) {
-    vs.var( 'size', clean.querySize );
+    vs.var( 'size', 100 );
   }
 
   // focus point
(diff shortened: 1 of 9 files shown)


missinglink:
am I correct in saying this would remove the current `size` param for search?

ie. the user will no longer be able to adjust the maximum amount of results returned?

trescube:
This only affects the number of hits requested from ES, `middleware/dedupe.js` handles limiting the ultimate number of results to that specified by the `size` API parameter.

trescube:
Additionally, this only affects the new fallback ES query and not the "revert to old behavior" ES query.

PR #687, opened 2016-10-13
Rock-Paper-Scissors: check which record is preferred when dupe is found

### LINKED ISSUE #650: Determine which of the duped items to keep in search results
Add a function in the deduper middleware to check which of the two items is prefered when a dupe is found. This would allow us to pick WOF records over Geonames, and OA over OSM, etc.

dianashk:
⚫️ 📃 ✂️  We need to apply a few basic rules to determine which of the dupes to keep in the results and which to throw away. So far the rules are as follows:

- whosonfirst trumps geonames (assuming admin areas, although could be venues as well, do we care???)
- openaddresses trumps openstreetmap (assuming addresses)
- openstreetmap trumps whosonfirst (assuming venues)

Fixes #650 
Fixes pelias/pelias#220

### DIFF
--- a/middleware/dedupe.js
+++ b/middleware/dedupe.js
@@ -41,7 +41,7 @@ function dedupeResults(req, res, next) {
           hit: hit.name.default + ' ' + hit.source + ':' + hit._id
         });
         // replace previous dupe item with current hit
-        uniqueResults.splice(dupeIndex, 1, hit);
+        uniqueResults[dupeIndex] = hit;
       }
       // if not preferred over existing, just log and move on
       else {
--- a/middleware/dedupe.js
+++ b/middleware/dedupe.js
@@ -16,11 +16,41 @@ function dedupeResults(req, res, next) {
   var uniqueResults = [];
 
   _.some(res.data, function (hit) {
-    if (uniqueResults.length === 0 || _.every(uniqueResults, isDifferent.bind(null, hit)) ) {
+
+    if (_.isEmpty(uniqueResults)) {
       uniqueResults.push(hit);
     }
     else {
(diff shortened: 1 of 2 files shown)


dianashk:
Yay, thanks! Agreed on the need to improve isDifferent. I've added a [new issue](pelias/pelias#438)/

middleware/dedupe.js
```diff
@@ -16,11 +16,41 @@ function dedupeResults(req, res, next) {
...
+      // since the order in which Elasticsearch returns identical text matches is arbitrary
+      // of course, if the new one is preferred we should replace previous with new
+      else if (isPreferred(uniqueResults[dupeIndex], hit)) {
+        logger.info('[dupe][replacing]', {
+          query: req.clean.text,
+          previous: uniqueResults[dupeIndex].source,
+          hit: hit.name.default + ' ' + hit.source + ':' + hit._id
+        });
+        // replace previous dupe item with current hit
+        uniqueResults.splice(dupeIndex, 1, hit);
```
trescube: Is there a performance reason for using `splice` instead of `uniqueResults[dupeIndex] = hit`?

dianashk: you're right, it could be that simple. somehow it seemed more complicated in my head. will update. thanks!

middleware/dedupe.js
```diff
@@ -41,7 +41,7 @@ function dedupeResults(req, res, next) {
           hit: hit.name.default + ' ' + hit.source + ':' + hit._id
         });
         // replace previous dupe item with current hit
-        uniqueResults.splice(dupeIndex, 1, hit);
+        uniqueResults[dupeIndex] = hit;
```
orangejulius: cleaner, yay!

PR #688, opened 2016-10-13
Increment geocodejson format version number in results

dianashk:
Fixes pelias/pelias#315


trescube:
:shipit:

orangejulius:
Wait, isn't this different than the API version number?

dianashk:
@orangejulius, yes this is incrementing the geocodejson version because we've added a few new params, such as `match_type` and `accuracy` along with the libpostal integration work. We can discuss incrementing the engine version, which we probably should be doing with each new release.

orangejulius:
Ah ok. pelias/pelias#315 mentions the engine version, which one do we want to change?

PR #698, opened 2016-10-18
Update Dockerfile to avoid npm install fail

ambodi:
I was trying to build the image and it actually fails on NPM of Node 0.12 doesn't support some of the packages needed.


orangejulius:
Hey @ambodi,
Good catch, thanks for fixing it.

ambodi:
@orangejulius no worries! Great API btw :)

PR #702, opened 2016-10-19
Update ciao test parsed_text expectations for libpostal

orangejulius:
The format of `parsed_text` has changed with libpostal, but the ciao tests were still expected the addressit format.


orangejulius:
Hm, on the other hand, since `libpostal` isn't yet the default `textAnalyzer` maybe we should keep this as is.

dianashk:
For all intents and purposes `libpostal` is our default. It's the only thing we really focus on testing and maintaining anyway. So I would rather have our tests reflect that reality, even if our config files don't just yet.

PR #705, opened 2016-10-21
enabled city/state only inputs for fallback queries

trescube:
This PR enables city+state inputs for FallbackQuery.  Only inputs that are parsed as city+state are used, any other field identified will skip FallbackQuery.  For example, `Lancaster, PA` will return results but `Lancaster, PA, United States` won't.

### DIFF
--- a/query/search.js
+++ b/query/search.js
@@ -120,7 +120,7 @@ function generateQuery( clean ){
 }
 
 function getQuery(vs) {
-  if (hasStreet(vs) || isCityStateOnly(vs)) {
+  if (hasStreet(vs) || isCityStateOnlyWithOptionalCountry(vs)) {
     return {
       type: 'fallback',
       body: fallbackQuery.render(vs)
@@ -137,7 +137,7 @@ function hasStreet(vs) {
   return vs.isset('input:street');
 }
 
-function isCityStateOnly(vs) {
+function isCityStateOnlyWithOptionalCountry(vs) {
   var isSet = function(layer) {
     return vs.isset('input:' + layer);
   };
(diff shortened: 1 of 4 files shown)


query/search.js
```diff
@@ -1,7 +1,8 @@
 var peliasQuery = require('pelias-query'),
     defaults = require('./search_defaults'),
     textParser = require('./text_parser'),
-    check = require('check-types');
+    check = require('check-types'),
+    _ = require('lodash');
```
orangejulius: Is lodash actually required? `.every` and `.some` are methods on `Array`, so I think it's actually fine.

trescube: d'oh!  i had been using lodash, i'll remove.

dianashk:
The code looks good. We'll need to also update the confidence score computation in [confidencScoreFallback.js](https://github.com/pelias/api/blob/master/middleware/confidenceScoreFallback.js).

trescube:
I modified confidenceScoreFallback for account for city/state/optional-country.  I toyed with the idea of tying fallback scores to the input granularity start point (that is, a fallback-to-state-from-city-input would score differently than a fallback-to-state-from-address-input) but that got all magic numbery.  It's open to discussion.   It does seem a bit harsh to assign a score of 0.3 to city/state inputs that fallback to state.

test/unit/query/search.js
```diff
@@ -270,6 +270,80 @@ module.exports.tests.query = function(test, common) {
...
+      t.equals(query, undefined, 'should have returned undefined');
+
+    });
+
+    t.end();
+
+  });
+
+  test('city and state with at least one other input: field should return undefined', function(t) {
+    ['query', 'category', 'number', 'neighbourhood', 'borough', 'postalcode', 'county'].forEach(function(placeType) {
```
orangejulius: this is not so bad right now but we've seen for loops in unit tests get out of control before. Imagine what 6 months of changes quickly bolted on to this test could do. it does feel repetitive but having this split out into several tests is probably better.

PR #708, opened 2016-10-25
bumped labels version to 1.5.0

trescube:


orangejulius:
we have robots to do this for us :P #707 does the same thing

trescube:
Well, if you want to trust "the robots" then I'm not going to stop you, but I assume you've seen Terminator 2?

trescube:
One day they're handling node dependencies, the next they're enslaving humanity.

PR #712, opened 2016-10-28
Add 5box support

trescube:
Added new query type, `component_geocoding`.  

Connected to pelias/pelias#417


dominikkukacka:
`fallback.street` is not working.

The resulting query shows: `"address_parts.street": "Teststraße 1"`
Where it should be split into `address_parts.street` and `address_parts.number`

trescube:
That's a known issue at this time mainly due to figuring out what the house number portion of the `address` field is.  It's not enough to just remove numbers or the first number or the last number since that requires knowledge of the address formatting scheme for the country (which may not be known).  The benefit of 1box /search is that libpostal splits up the `house number` and `street` fields.  Our current plan to address this (after initial release) is call libpostal with the supplied `address` field and work some magic with it's interpretation.  This approach isn't fool-proof since libpostal sometimes misinterprets the input.  For example, it works great sometimes:  

```
> 30 w 26th st

Result:

{
  "house_number": "30",
  "road": "w 26th st"
}
```

Sometimes it interprets the intended house number as a postal code:

```
> 1338 kobbe ave

Result:

{
  "postcode": "1338",
  "road": "kobbe ave"
}
```

Fortunately it does a pretty good job for the most part even when numbers appear at various parts:

```
> 2831 route 15

Result:

{
  "house_number": "2831",
  "road": "route 15"
}
```

See pelias/api#723

dominikkukacka:
OK, good to know.
Thats exactly how I implemented it now for my use-case.

package.json
```diff
@@ -56,7 +56,7 @@
     "pelias-labels": "1.5.0",
     "pelias-logger": "0.1.0",
     "pelias-model": "4.3.0",
-    "pelias-query": "8.8.0",
+    "pelias-query": "8.9.0",
```
orangejulius: If you rebase now, this will go away :)

PR #715, opened 2016-11-04
Support "county, region" in FallbackQuery with edge case for Hawaii

orangejulius:
To help with the [Maui, Hawaii](https://github.com/pelias/whosonfirst/issues/94#issuecomment-253669462) query issues for WOF venues, this PR enables "county, region" queries in FallbackQuery.

Additionally, because libpostal parses Maui as an island, there is some special handling: the `island` field is moved to `county` if the `region` field is `hawaii`.
```
> maui, hawaii

Result:

{
  "island": "maui",
  "state": "hawaii"
}
```

These changes together allow us to properly return Maui County. No other acceptance-tests are affected!
screenshot from 2016-11-04 15-07-40

Requires https://github.com/pelias/text-analyzer/pull/15 to be merged.

### DIFF
--- a/query/text_parser.js
+++ b/query/text_parser.js
@@ -39,6 +39,12 @@ function addParsedVariablesToQueryVariables( parsed_text, vs ){
     vs.var( 'input:postcode', parsed_text.postalcode );
   }
 
+
+  // island
+  if( parsed_text.hasOwnProperty('island') ){
+    vs.var( 'input:island', parsed_text.island );
+  }
+
   // ==== add parsed matches [admin components] ====
 
   // city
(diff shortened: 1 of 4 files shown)


query/text_parser.js
```diff
@@ -91,4 +102,11 @@ function shouldSetQueryIntoHouseNumber(vs) {
           /^[0-9]+$/.test(vs.var('input:query').toString());
 }
 
+function shouldTreatIslandAsCounty(vs) {
+  return vs.isset('input:island') &&
+    vs.isset('input:region') &&
+    !vs.isset('input:county') &&
+    vs.var('input:region').toString() === 'hawaii';
```
dianashk: this definitely calls for an awesome comment about why it's only `hawaii` at this time and the implications of this one-off fix. also is it safe to assume this will always be all lowercase?

trescube: Ooh, good catch, `maui, hi` and `maui, hawaii` and parsed the same.

trescube:
```
> maui hawaii

Result:

{
  "island": "maui",
  "state": "hawaii"
}

> maui hi

Result:

{
  "island": "maui",
  "state": "hi"
}
```

orangejulius: Yeah, I don't like this PR for a lot of reasons. One off fixes just generally don't feel right, it doesn't handle other Hawaiian islands/counties in any particularly good way. The only things this PR has going for it is that it fixes an acceptance-test and doesn't seem to break anything else.

You're right that we would have to handle `hi`. I did check and libpostal always lowercases.

orangejulius:
Closing this since we're going to look into more generalized ways to solve the Maui Maui problem in the future.

PR #727, opened 2016-11-17
added iso2->iso3 conversion to /search endpoint

trescube:
This change uses the iso3166-1 package to convert an ISO2 country code identified by libpostal into an ISO3.  This is required because we don't currently support ISO2 country codes as altnames so we can't match them and wind up falling back to old production behavior.


trescube:
It's a bit trickier because /autocomplete either already does it or doesn't make a distinction of the country field reliably enough.  For example, addressit identifies `state: DE` from `grolmanstraße 51, berlin, de` while `30 w 26th street, new york, ny, us` is already converted to `country: USA`.

PR #729, opened 2016-11-17
Adjust autocomplete focus boost

orangejulius:
This lowers the autocomplete focus boost from 40 to 15. The idea is
that, because the boost for population is 20, the focus can't possibly
override a popular city, but it can come close.

On our default dev build, no acceptance tests fail but the san francisco
autocomplete test now passes!

It's not clear if this fixes WOF venue issues yet, since the latest
build there failed, so stay tuned :(


orangejulius:
There are more tweaks on deck on my machine that help with WOF venue things, but these changes are also good (and we can always tweak them later). I'd prefer to merge the WOF venue compatible changes along with this change, but since there are some other high priority things to do that may block that, and this is ready to go, we'll just merge it as is.

PR #732, opened 2016-11-18
standardize on non-US-influenced parameter names

trescube:
switched from `city` and `state` to `locality` and `region`, respectively

### DIFF
--- a/sanitizer/_synthesize_analysis.js
+++ b/sanitizer/_synthesize_analysis.js
@@ -1,7 +1,15 @@
 const _ = require('lodash');
 
-const fields = ['address', 'neighbourhood', 'borough', 'city', 'county',
-  'state', 'postalcode', 'country'];
+const fields = {
+  'address': 'address',
+  'neighbourhood': 'neighbourhood',
+  'borough': 'borough',
+  'locality': 'city',
+  'county': 'county',
+  'region': 'state',
+  'postalcode': 'postalcode',
+  'country': 'country'
+};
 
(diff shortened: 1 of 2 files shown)


sanitizer/_synthesize_analysis.js
```diff
@@ -1,7 +1,15 @@
 const _ = require('lodash');
 
-const fields = ['address', 'neighbourhood', 'borough', 'city', 'county',
-  'state', 'postalcode', 'country'];
+const fields = {
+  'address': 'address',
+  'neighbourhood': 'neighbourhood',
+  'borough': 'borough',
+  'locality': 'city',
```
orangejulius: Wait, why do we need this? What part still uses city?

trescube: We adopted the libpostal names for our internal representation:

```
> lancaster, pa

Result:

{
  "city": "lancaster",
  "state": "pa"
}
```

This change is just for outward-facing behavior.  People shouldn't be coding against our internal analysis representation but that should be changed at some point, though it's a bigger effort than what's been done here.

trescube: Here's our internal representation:

```
"query": {
"text": "lancaster, pa",
"parsed_text": {
"city": "lancaster",
"state": "pa"
},
"size": 10,
"private": false,
"querySize": 20
}
```

orangejulius: oh, right. that makes sense.

PR #736, opened 2016-11-23
Use libpostal to parse address field

trescube:
This PR adds support to call text-analyzer (effectively libpostal) to parse the `address` field into `number` and `street` fields.  Libpostal isn't fool-proof for this task, so there are 2 scenarios that this covers:

- `number` can be parsed from `address` field as expected:

```
> 1220 calle de lago

Result:

{
  "house_number": "1220",
  "road": "calle de lago"
}
```

- `number` is identified as `postcode` (`postcode` can be reasonably safely assumed to be the `number` since the caller has already stated that the input is the address portion):

```
> 1338 kobbe ave

Result:

{
  "postcode": "1338",
  "road": "kobbe ave"
}
```

Text analysis is only called if an `address` field was supplied.  If libpostal identifies both a `house_number` *and* `postcode` fields, `house_number` is used for the `number` field.  Regardless of field, if a `number` has been identified, the *first* instance of this value is removed from `address`.  If no `number` can be extracted from the `address` field, the entire `address` field is moved to the `street` field.  

For example:

```
> 15 us route 15

Result:

{
  "house_number": "15",
  "road": "us route 15"
}
```

The result will be that `number` is now `15`, `street` is now `us route 15`, and `address` is removed.  There's no guarantee that this will work in all cases.  

It does not matter what libpostal parses the rest of the `address` field as, the remainder after the `number` value is removed is assigned to `street`.  

For example:

```
> 5 russian hill pl

Result:

{
  "house_number": "5",
  "suburb": "russian hill",
  "road": "pl"
}
```

assigns "russian hill pl" to `street`.  

Fixes pelias/pelias#454
See pelias/query#52

### DIFF
--- a/sanitizer/_synthesize_analysis.js
+++ b/sanitizer/_synthesize_analysis.js
@@ -1,4 +1,5 @@
 const _ = require('lodash');
+const text_analyzer = require('pelias-text-analyzer');
 
 const fields = {
   'address': 'address',
@@ -18,6 +19,15 @@ function normalizeWhitespaceToSingleSpace(val) {
 function isPostalCodeOnly(parsed_text) {
   return Object.keys(parsed_text).length === 1 &&
           parsed_text.hasOwnProperty('postalcode');
+}
+
+function getHouseNumberField(analyzed_address) {
+  for (var field of ['number', 'postalcode']) {
+    if (analyzed_address.hasOwnProperty(field)) {
+      return field;
+    }
+  }
+
 }
 
 function sanitize( raw, clean ){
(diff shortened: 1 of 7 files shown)


trescube:
See https://github.com/openaddresses/openaddresses/issues/2075 for more examples of libpostal parsing of just an address.

trescube:
This also fixes the `/component` issue where `address=30 west 26th st&locality=new york&region=ny` returns `26 West 30th Street`.

sanitizer/_synthesize_analysis.js
```diff
@@ -43,6 +53,24 @@ function sanitize( raw, clean ){
...
 
+  if (clean.parsed_text.hasOwnProperty('address')) {
+    var analyzed_address = text_analyzer.parse(clean.parsed_text.address);
+
+    const house_number_field = getHouseNumberField(analyzed_address);
+
+    if (house_number_field) {
+      clean.parsed_text.number = analyzed_address[house_number_field];
+
+      clean.parsed_text.street = _.trim(_.replace(clean.parsed_text.address, clean.parsed_text.number, ''));
```
dianashk: comment here or separate function would make it more readable.

PR #740, opened 2016-11-29
updated naming from component to structured

trescube:
updated entry point from `beta/component` to `search/structured`


routes/v1.js
```diff
@@ -193,7 +193,7 @@ function addRoutes(app, peliasConfig) {
   app.get ( base + 'autocomplete', routers.autocomplete );
   app.get ( base + 'search',       routers.search );
   app.post( base + 'search',       routers.search );
-  app.get ( base + 'beta/component',    routers.component );
+  app.get ( base + 'search/structured',    routers.structured );
```
dianashk: since the spacing here attempts to line up the `routers` column maybe moving all of them over a bit would be nicer

PR #743, opened 2016-12-02
add status code logging to all the places

trescube:


middleware/500.js
```diff
@@ -7,6 +7,8 @@ function middleware(err, req, res, next) {
   var error = (err && err.message) ? err.message : err;
 
   if( res.statusCode < 400 ){ res.status(500); }
+  logger.info('status code', res.statusCode);
```
trescube: Move this line up to get true status code

trescube:
Closing since we can do all this in [morgan](https://www.npmjs.com/package/morgan)

PR #745, opened 2016-12-02
Git is drunk, or maybe I am

cag:
Git says production is many commits off of master, but it also says that production and master are the same.


cag:
Clearly, I am.

PR #750, opened 2016-12-03
Added a clustered pm2 config

cag:
You know, for this thing: http://pm2.keymetrics.io/

### DIFF
--- a/pm2.config.js
+++ b/pm2.config.js
@@ -1,0 +1,24 @@
+module.exports = {
+  /**
+   * Application configuration section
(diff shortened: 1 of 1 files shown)


orangejulius:
Hey, this is really cool. PM2 looks really good. But I don't think it makes sense to clutter the root directory of our project for a config that only some people will use. Also, its unlikely we'll start using it for Mapzen Search since our ops setup already handles monitoring the API. That means if we merge it, we'll never update it, and it will quickly become out of date, and confuse people. But if we ever need a new process manager we'll look at PM2.

PR #746, opened 2016-12-02
improvements for handling non-scalar request params. fixes #744

missinglink:
improvements for handling non-scalar request params. fixes #744

this PR moves the `_single_scalar_parameters` sanitizer before all others and deletes any non-scalar parameters so they cannot cause issues in other sanitizers/middleware.

added more ciao tests to cover these cases

PR #753, opened 2016-12-05
improvements for handling non-scalar request params. fixes #744

### LINKED ISSUE #744: duplicate request params create Array in expressjs
eg. `/v1/reverse?layers=region&point.lat=47.417&point.lon=10.75&size=3&sources=geonames&sources=gn`

orangejulius:
This is the same code as https://github.com/pelias/api/pull/746, but based off master.


orangejulius:
Fixes https://github.com/pelias/pelias/issues/474

PR #761, opened 2016-12-09
Add parser and query param logging

dianashk:
More logging to be used in graphite and logstash

### DIFF
--- a/controller/search.js
+++ b/controller/search.js
@@ -64,6 +64,9 @@ function setup( config, backend, query ){
         res.meta = meta || {};
         // store the query_type for subsequent middleware
         res.meta.query_type = renderedQuery.type;
+
+        logger.info(`[controller:search] [queryType:${renderedQuery.type}] [es_result_count:` +
+          (res.data && res.data.length ? res.data.length : 0));
       }
       logger.debug('[ES response]', docs);
       next();
--- a/query/autocomplete.js
+++ b/query/autocomplete.js
@@ -1,8 +1,10 @@
+'use strict';
 
-var peliasQuery = require('pelias-query'),
-    defaults = require('./autocomplete_defaults'),
-    textParser = require('./text_parser_addressit'),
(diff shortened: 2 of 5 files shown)


query/autocomplete.js
```diff
@@ -59,23 +61,28 @@ query.filter( peliasQuery.view.boundary_rect );
 **/
 function generateQuery( clean ){
 
-  var vs = new peliasQuery.Vars( defaults );
+  const vs = new peliasQuery.Vars( defaults );
+
+  let logStr = '[query:autocomplete] [parser:addressit] ';
```
orangejulius: minor nitpick: it might make sense to rename this something like `paramLogStr`, since we could be logging lots of things in this file.

controller/search.js
```diff
@@ -64,6 +64,9 @@ function setup( config, backend, query ){
         res.meta = meta || {};
         // store the query_type for subsequent middleware
         res.meta.query_type = renderedQuery.type;
+
+        logger.info(`[controller:search] [queryType:${renderedQuery.type}] [es_result_count:` +
```
orangejulius: oh man, this is gonna be great

query/autocomplete.js
trescube: It can also be a `const` because Javascript.  Apparently things can be appended to a const.

orangejulius: oh. are strings not immutable in Javascript? Interesting

trescube: oh, no, sorry, i interpreted my test code incorrectly.  Ignore me, apparently `+=` operates differently between `const` and `let`:

```
> const s = 'a';
undefined
> s += 'b';
'ab'
> s
'a'
> let t = 'a';
undefined
> t += 'b';
'ab'
> t
'ab'
```

trescube: It looks like it just returns a new string. That's super confusing.

PR #763, opened 2016-12-09
Add query support to structured

trescube:
Adds `query` parameter support to `/v1/search/structured`.  Most everything was in place before, this just adds `query` to analysis synthesis and adjusts tests to query module changes.  

Connected to pelias/pelias#483

### DIFF
--- a/sanitizer/_synthesize_analysis.js
+++ b/sanitizer/_synthesize_analysis.js
@@ -2,6 +2,7 @@ const _ = require('lodash');
 const text_analyzer = require('pelias-text-analyzer');
 
 const fields = {
+  'query': 'query',
   'address': 'address',
   'neighbourhood': 'neighbourhood',
   'borough': 'borough',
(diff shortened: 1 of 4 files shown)


sanitizer/_synthesize_analysis.js
```diff
@@ -2,6 +2,7 @@ const _ = require('lodash');
 const text_analyzer = require('pelias-text-analyzer');
 
 const fields = {
+  'query': 'query',
```
dianashk: Can we call the param something other than query? Maybe poi_name or venue? Query is confusing I'm my opinion.

dianashk:
After discussing in our team meeting, we think changing this to `name` makes sense here. We will continue to also search the categories for results, but the user doesn't need to be directly aware of that. When we officially release the `categories` parameter for search it will be supported in this endpoint as well, so it doesn't seem necessary to include a dedicated `category` parameter for the query as well.

trescube:
Oh, sorry I missed this discussion since I was ooto.

trescube:
After a team meeting on 1/11/2017, it was decided that the parameter name would be `venue`.

PR #767, opened 2016-12-13
Add Mt/St/Ft normalizing

trescube:
WOF names are spelled out as Mount and Fort whereas Saint and Sainte are abbreviated to St and Ste, respectively.  This PR adds a sanitizer that transliterates:

* Mt -> Mount
* Ft -> Fort
* St -> Saint
* Ste -> Sainte

This will ensure that, for example, searching for `Saint Louis, MO` and `St. Louis` return the same results.  

Fixes pelias/schema#157


sanitizer/search.js
```diff
@@ -6,6 +6,7 @@ var sanitizeAll = require('../sanitizer/sanitizeAll'),
       quattroshapes_deprecation: require('../sanitizer/_deprecate_quattroshapes'),
       text: require('../sanitizer/_text'),
       iso2_to_iso3: require('../sanitizer/_iso2_to_iso3'),
+      mount_saint_fort_standardizer: require('../sanitizer/_mount_saint_fort_standardizer'),
```
orangejulius: might we want to expand this in the future so that it's less specific and more of an "abbreviation_standardizer"? It feels a bit too specific of a name

sanitizer/_mount_saint_fort_standardizer.js
```diff
@@ -0,0 +1,48 @@
...
+    // after 3.  fort st louis
+
+    // 1.  remove '.' that could abbreviate ft and mt (makes transliteration regex easier)
+    const periods_removed = _.toLower(clean.parsed_text.city).replace(/\b(mt|ft)\./g, '$1 ');
+
+    // 2.  transliterate 'saint'->'st', etc
+    const transliterated = periods_removed.replace(mountSaintFort, transliterate);
+
+    // 3.  whitespace-normalize by replacing much whitespace with a space and trimming
+    //     duplicate whitespace can be introduced when removing periods
```
orangejulius: I think there's a missing word in here. should it be "and trimming duplicate whitespace _that_ can be introduced when removing periods" (or something similar)

sanitizer/_mount_saint_fort_standardizer.js
```diff
@@ -0,0 +1,48 @@
...
+}
+
+// transliterate ft/mt/saint/sainte to fort/mount/st/ste, respectively
+function sanitize(raw, clean) {
+  // error & warning messages
+  // this function doesn't add any error or warning messages
+  const messages = { errors: [], warnings: [] };
+
+  // only try to transliterate if there is a city in parsed_text
+  if (!_.isEmpty(_.get(clean, 'parsed_text.city'))) {
```
orangejulius: woah `_.get` is kinda nice

PR #775, opened 2017-01-04
added configuration validation + tests

trescube:
# Required
* `api.version` (string)
* `api.indexName` (string)
* `api.host` (string)

# Optional

* `api.legacyUrl` (string)
* `api.accessLog` (string)
* `api.relativeScores` (boolean)
* `api.localization` (object with optional `flipNumberAndStreetCountries` array)
* `esclient.requestTimeout` (integer)

### DIFF
--- a/app.js
+++ b/app.js
@@ -1,11 +1,13 @@
 
 var app = require('express')();
 
-var peliasConfig = require( 'pelias-config' ).generate().api;
+var peliasConfig = require( 'pelias-config' ).generate();
 
+// validate the configuration before attempting to load the app
+require('./src/configValidation').validate(peliasConfig);
 
-if( peliasConfig.accessLog ){
-  app.use( require( './middleware/access_log' ).createAccessLogger( peliasConfig.accessLog ) );
+if( peliasConfig.api.accessLog ){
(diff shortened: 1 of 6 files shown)


trescube:
The changes to app.js doesn't pass around the root object, it passes around `peliasConfig.api` which is effectively the same thing:

* https://github.com/pelias/api/blob/d8cffe4b9bc0c1e9a52c070e60bc456fa3d823a5/app.js#L9
* https://github.com/pelias/api/blob/d8cffe4b9bc0c1e9a52c070e60bc456fa3d823a5/app.js#L10
* https://github.com/pelias/api/blob/d8cffe4b9bc0c1e9a52c070e60bc456fa3d823a5/app.js#L23
* https://github.com/pelias/api/blob/d8cffe4b9bc0c1e9a52c070e60bc456fa3d823a5/app.js#L26

trescube:
It shouldn't matter that geocodeJSON.js requires the config separately, I put the config validation in app.js which throws an error so the app won't even start on config failure.  The config validation also verifies everything that geocodeJSON.js is expecting (just `host` and `version`).

orangejulius:
Ohh, right, I didn't see that you put the `.api` on the lines where the config is passed on. All good then!

PR #782, opened 2017-01-18
Add elasticsearch retries

trescube:
This PR is in service of addressing occasional timeouts when making requests to elasticsearch.  The solution is to retry ES operations that have timed out.  While the elasticsearch docs claim to support `maxRetries`, the ES node module does not consider a request timeout a retryable operation because... reasons?  Anyway, another limitation is that specifying a timeout in the request body which should override the ES client timeout doesn't appear to work.  That leaves an implementation to do the following:

* programmatically support `maxRetries`
* accept the pelias config `requestTimeout` since it can't be overridden

I chose the [retry](https://www.npmjs.com/package/retry) module to implement the solution since it handles the bookkeeping of counting maxRetries, calculates formulaic timeouts, and provides an easy abstraction for retrying requests (`if (operation.retry(err))`).  

For example, if pelias config esclient contains: 
```
{
  "requestTimeout": 10,
  "maxRetries": 4
}
```

then pelias will try each request up to 4 times with a 10ms timeout before failing for good.  

All non-`/v1/place` searches filter thru controller/search.js which is where the retry logic has been added.  `/v1/place` doesn't seem to be a very popular service so I haven't included the retry logic there but don't have a problem doing so if others think it would be good to have.  

This PR also refactors a bit by removing the dependency on geopipes/elasticsearch-backend, moving esclient initialization into routes/v1, and cleaned up routes/v1 a bit.


controller/place.js
```diff
@@ -24,9 +20,7 @@ function setup( config, backend ){
 
     logger.debug( '[ES req]', query );
 
-    service.mget( backend, query, function( err, docs ) {
-      console.log('err:' + err);
```
missinglink: did you intend to remove this error logging?

it seems a bit weird to log every time, since I'm guessing for a vast majority of times it would be `null` anyway

routes/v1.js
```diff
@@ -60,30 +60,30 @@ var postProc = {
  * @param {object} peliasConfig
  */
 function addRoutes(app, peliasConfig) {
+  const esclient = require('elasticsearch').Client(peliasConfig.esclient);
```
missinglink: nit pick: could have this `require` at the top of the file with the other `require` statements?

sanitizer/search_fallback.js
```diff
@@ -19,7 +19,6 @@ module.exports.middleware = function( req, res, next ){
   // log the query that caused a fallback since libpostal+new-queries didn't return anything
   if (req.path === '/v1/search') {
     var queryText = logging.isDNT(req) ? '[text removed]' : req.clean.text;
-    logger.info(queryText);
```
missinglink: did you intend to remove this logging statement?

controller/search.js
```diff
@@ -38,38 +36,68 @@ function setup( config, backend, query ){
...
     }
 
-    // backend command
+    logger.debug( '[ES req]', cmd );
+
+    // options for retry
+    // maxRetries is from the ES client which defaults to 3
+    // factor of 1 means that each retry attempt will esclient requestTimeout
+    const operationOptions = {
+      retries: _.get(esclient, 'transport.maxRetries'),
```
missinglink: it might be best to use a different variable name, a variable named `maxRetries` already exists for `elasticsearch-js` with a different meaning.

controller/place.js
trescube: Yeah, that was intentional, it was cluttering up the logs.

sanitizer/search_fallback.js
trescube: Weird, I don't recall removing that, but I guess it doesn't the code doesn't do much without it.

controller/place.js
```diff
@@ -1,44 +1,82 @@
...
+const retry = require('retry');
 
-  // allow overriding of dependencies
-  backend = backend || require('../src/backend');
+const mgetService = require('../service/mget');
+const logger = require('pelias-logger').get('api');
 
-  function controller( req, res, next ){
+function setup( apiConfig, esclient ){
+  function requestHasErrors(request) {
```
dianashk: I would suggest pulling all the function definitions out of the `setup()` function since they are all independent of the score of that function. The way it is written here they'll be redefined each time the setup function is executed, not to mention keep around the setup function as part of their context, and that's not necessary.

test/unit/mock/backend.js
```diff
@@ -73,23 +59,21 @@ responses['client/mget/fail/1'] = responses['client/search/fail/1'];
...
+        return responses[key].apply( this, arguments );
+      },
+      search: function( cmd, cb ){
+        if( 'function' === typeof cmdCb ){ cmdCb( cmd ); }
+        return responses[key].apply( this, arguments );
       }
     };
   }
-  return backend;
+  return backend();
```
orangejulius: Can this `backend` function be removed if setup returns the object with responses directly?

controller/search.js
```diff
@@ -1,75 +1,118 @@
...
     if (_.isUndefined(renderedQuery)) {
       return next();
     }
 
-    // backend command
-    var cmd = {
-      index: config.indexName,
+    // options for retry
+    // maxRetries is from the API config with default of 3
+    // factor of 1 means that each retry attempt will esclient requestTimeout
```
orangejulius: i think there are some missing words here on line 53

controller/place.js
```diff
@@ -1,44 +1,82 @@
...
+    logger.debug( '[ES req]', cmd );
 
-    service.mget( backend, query, function( err, docs ) {
-      console.log('err:' + err);
+    operation.attempt((currentAttempt) => {
+      mgetService( esclient, cmd, function( err, docs ) {
+        // returns true if the operation should be attempted again
+        // (handles bookkeeping of maxRetries)
+        // only consider for status 408 (request timeout)
+        if (isRequestTimeout(err) && operation.retry(err)) {
```
orangejulius: the interface to this `retry` module is really nice.  good find

controller/place.js
```diff
@@ -1,44 +1,82 @@
...
-    if( req.errors && req.errors.length ){
+function isRequestTimeout(err) {
+  return _.get(err, 'status') === 408;
+}
+
+function setup( apiConfig, esclient ){
+  function controller( req, res, next ){
+    // do not run controller when a request validation error has occurred.
+    if (requestHasErrors(req)){
       return next();
```
orangejulius: lots of great cleanup in this file

Issue #350, opened 2015-10-15
Searching for a postal code should only return matching postalcode records


trescube:
_PREVIOUS TITLE_: Searching for postal code only returns unrelated results

Pretty much any US postal code only request returns street results that are unrelated.

stephenkhess:
Examples:

87801 (in New Mexico) returns addresses with house number 87801:
http://search.mapzen.com/v1/search?api_key=search-xDYkNp8&text=87801

19010 (west of Philadelphia, PA) returns a few addresses within the 19010 postal code but the rest are house numbers
http://search.mapzen.com/v1/search?api_key=search-xDYkNp8&text=19010

17356 (Red Lion, PA) returns a combination of addresses in 17356 and 17356 house numbers
http://search.mapzen.com/v1/search?api_key=search-xDYkNp8&text=17356

missinglink:
this is due to the very top part of the query being a `'must match'` on the `name` field.

as its currently set up we don't surface any results where the postcode tokens aren't also contain in the name; which is clearly not ideal ;)

there are probably a bunch of ways we can fix this, one would be to change the query to a postcode-only query when the input parser is happy that we only want to search by postcode.

eg. when the parser says:
```
"parsed_text": {
  "postalcode": 87801,
  "regions": []
}
```

.. we could do a super simply query such as:
```
{
   "query": {
      "match": {
         "address.postcode": {
            "query": "87801"
         }
      }
   }
}
```

riordan:
Given that we don't have postal code boundaries (kinda sorta, OSM is iffy about this, and we don't even import them), we will need some sort of expected behavior when that postal code is searched for.

Should we return a bunch of points within that postal code or draw a box around that postal code and return that box's centroid? Are there other options until we have suficient data for this?

riordan:
More on the general issue for postal codes being tracked in: pelias/pelias #201

dianashk:
Done! 💃  🕺

PR #796, opened 2017-02-08
Search by postalcode

dianashk:
⚠️ _This change only updates search, not autocomplete or structured. Future and separate PRs should be submitted for that._

- Allow users to search for postalcodes alone, [lilke so](http://pelias.github.io/compare/#/v1/search%3Ftext=90210)
- This will not have any affect on other types of search queries where postalcode is one of many tokens

Connected to #350


middleware/renamePlacenames.js
```diff
@@ -1,12 +1,14 @@
...
-var ADDRESS_PROPS = {
-  'number': 'housenumber',
-  'zip': 'postalcode',
-  'street': 'street'
-};
+const PARENT_PROPS = require('../helper/placeTypes');
+
+const ADDRESS_PROPS = [
+  { name: 'number', newName: 'housenumber' },
+  { name: 'zip',    newName: 'postalcode', transform: (value) => { return [value]; } },
```
trescube: yeah, shorter function syntax!

PR #810, opened 2017-03-06
added support for coarse reverse geocoding

trescube:
several things are in this commit:

- coarse reverse controller / pip service
- inject "conditional execute" predicate into search service to shortcut execution conditions
- added coarse reverse controller to precede standard reverse controller
- lots of tests!


controller/coarse_reverse.js
```diff
@@ -0,0 +1,129 @@
+const logger = require('pelias-logger').get('coarse_reverse');
```
dianashk: Doesn't this need 'use strict'?

controller/predicates/is_coarse_reverse.js
```diff
@@ -0,0 +1,7 @@
+const _ = require('lodash');
+
+module.exports = (req, res) => {
+  // returns true if layers is undefined, empty, or contains 'address' or 'venue'
+  return !_.isEmpty(req.clean.layers) &&
+          _.intersection(req.clean.layers, ['address', 'venue']).length === 0;
```
dianashk: Should street layer be in this list?

controller/predicates/is_pip_service_enabled.js
```diff
@@ -0,0 +1,9 @@
+const _ = require('lodash');
+
+module.exports = (uri) => {
+  // this predicate relies upon the fact that the schema has already validated
+  //  that api.pipService is a URI-formatted string
+  return (request, response) => {
+    return uri !== undefined;
+  };
```
dianashk: Use the lodash isSet function or remove lodash from the top

dianashk: isSet is the wrong function, sorry. It checks that the value is of type Set, which I always forget! but there is _.isUndefined()

controller/coarse_reverse.js
trescube: `const` doesn't, `let` does.

routes/v1.js
```diff
@@ -80,9 +111,9 @@ function addRoutes(app, peliasConfig) {
       middleware.calcSize(),
       // 3rd parameter is which query module to use, use fallback/geodisambiguation
       //  first, then use original search strategy if first query didn't return anything
-      controllers.search(peliasConfig.api, esclient, queries.libpostal),
+      controllers.search(peliasConfig.api, esclient, queries.libpostal, not(hasDataOrErrors)),
```
trescube: Good point, I'll fix it.

routes/v1.js
```diff
@@ -1,6 +1,10 @@
 var Router = require('express').Router;
 var elasticsearch = require('elasticsearch');
 
+const all = require('predicates').all;
+const any = require('predicates').any;
+const not = require('predicates').not;
```
missinglink: yea this is cool

PR #816, opened 2017-03-14
Add debug:force-libpostal query param

dianashk:
🚧  DO NOT MERGE 🚧 

Allow a debug param to force query logic to use libpostal and to not fallback to addressit even when no results have been returned.

This should be updated after @trescube's PR is merged introducing middleware predicates. It will be the perfect place to do checks like this.


query/search.js
```diff
@@ -160,6 +162,10 @@ function getQuery(vs) {
 
 }
 
+function forceLibpostal(params) {
+  return _.isSet(params, 'force_libpostal');
```
trescube: I don't think this is doing what you're expecting.  `isSet` just returns whether the lone parameter is actually a `Set`.  You probably want `_.get(params, 'force_libpostal', false)`.

trescube: You already knew this apparently!  Just a placeholder then.

dianashk: Yup, missed changing it here. Thanks for catching it. Gets me every time!

trescube: It's a very misleading method name.  I'll submit a PR to change the name to `isTotesASet`.

orangejulius:
Is this ready? it would still be really cool to have

missinglink:
I am listed as a reviewer, is this still in review?

dianashk:
Not ready. I put it on hold a while back and haven't been able to come back to it. Sorry.

PR #819, opened 2017-03-17
negotiate HTTP locales for incoming browser requests

missinglink:
this middleware is responsible for negotiating HTTP locales for incoming
  browser requests by reading the querystring param 'lang' or 'Accept-Language' request header.

  the preferred language will then be available on the $req object:
  eg. for '?lang=fr' or 'Accept-Language: fr':
  ```javascript
  console.log( req.language );

  {
    name: 'French',
    type: 'living',
    scope: 'individual',
    iso6393: 'fra',
    iso6392B: 'fre',
    iso6392T: 'fra',
    iso6391: 'fr',
    defaulted: false
  }
  ```

---

  BCP47 language tags can contain three parts:
    1. A language subtag (en, zh).
    2. A script subtag (Hant, Latn).
    3. A region subtag (US, CN).

  at time of writing we will only be concerned with 1. (the language subtag) with
  the intention of being compatible with the language standard of whosonfirst data.

  whosonfirst data is in ISO 639-3 format so we will need to configure the library
  to support all ISO 639-1 (2 char) codes and convert them to 639-1 (3-char) codes.

  see: https://github.com/whosonfirst/whosonfirst-names

---

information about the language is available in the response object header:

```javascript
$ curl -s -H'Accept-Language: fr-CH, fr;q=0.9, en;q=0.8, de;q=0.7, *;q=0.5' "http://localhost:3100/v1/search"

{
  "geocoding": {
    "version": "0.2",
    "attribution": "http://pelias.mapzen.com/v1/attribution",
    "query": {
      "size": 10,
      "private": false,
      "lang": {
        "name": "French",
        "iso6391": "fr",
        "iso6393": "fra",
        "defaulted": false
      },
      "querySize": 20
    },
    ...
}
```


middleware/requestLanguage.js
```diff
@@ -0,0 +1,77 @@
...
+iso6393.filter( i => !!i.iso6391 ).forEach( i => language[ i.iso6391 ] = i );
+
+// a pre-processed locale list of language subtags we support (all of them).
+const allLocales = new locale.Locales( Object.keys( language ) );
+
+// return the middleware
+module.exports = function middleware( req, res, next ){
+
+  // input language, either from query param or header
+  var input = ( req.query && req.query.lang ) || ( req.headers && req.headers['accept-language'] );
```
trescube: What happens if `lang` is explicit and invalid but `accept-language` is valid?

middleware/requestLanguage.js
```diff
@@ -0,0 +1,77 @@
...
+
+  // parse request & choose best locale
+  var locales = new locale.Locales( input || '' );
+  var best = locales.best( allLocales );
+
+  // set $req.language property
+  req.language = language[ best.language ] || language.en;
```
trescube: We should probably return at least a warning that though the parameter/header was specified, we couldn't honor it due to it being an invalid value.

missinglink:
updated code to address @trescube feedback:

- in the case of either the query param or header (or both) being invalid, a warning is emitted
- in the case where the query param is invalid but the header is valid, use the header language

also:

- add logging in the form of:
```
info: [api] [lang] 'en' via 'default'
info: [api] [lang] 'fr' via 'querystring'
info: [api] [lang] 'fr' via 'header'
```

PR #825, opened 2017-03-24
Add dependency check to pre-commit hook and travis script

dianashk:


package.json
```diff
@@ -68,11 +69,17 @@
...
+    "pelias-interpolation": "*"
+  },
   "devDependencies": {
     "ciao": "^1.0.0",
+    "depcheck": "^0.6.7",
+    "dependency-check": "^2.8.0",
     "difflet": "^1.0.1",
     "istanbul": "^0.4.2",
     "jshint": "^2.5.6",
+    "npm-check": "^5.4.0",
```
orangejulius: There are 3 modules listed here, are we using all 3?

package.json
```diff
@@ -68,11 +69,17 @@
     "superagent": "^3.2.1",
     "through2": "^2.0.3"
   },
+  "optionalDependencies": {
+    "pelias-interpolation": "*"
```
orangejulius: does it make sense to put a version number here? I'm guessing no, since I wonder if Greenkeeper could even handle keeping it up to date.

dianashk: yeah i wasn't sure how to handle this since it doesn't actually exist in npm. we would need to either remove it or publish the module to npm so this works.

package.json
dianashk: yup, got excited and forgot to remove these. thanks for catching it.

dianashk:
I'm not sure why on travis it flags outdated modules but on my machine it didn't say anything about them at all. Looking into it now.

dianashk:
So it appears that removing the travis setting to grab npm modules from cache fixes the dependency checker update warnings. I think that using cache here is actually masking potential issues, even if it does make running our tests faster. 

Thoughts?

package.json
orangejulius: we probably should publish that module, since we publish everything else. it would make this nice and clean too. can we do that first and then update this PR?

.travis.yml
```diff
@@ -22,7 +19,7 @@ addons:
     packages:
       - g++-4.8
 before_install:
-  - npm i -g npm@^2.0.0
+  - npm i -g npm@^3.0.0
```
orangejulius: just recording that we're going to upgrade to NPM 3 in production to follow this

package.json
dianashk: yup. will do.

dianashk:
closing this in favor of https://github.com/pelias/api/pull/826

PR #832, opened 2017-04-03
Change natural language text based off requested language

missinglink:
this middleware is responsible for modifying the response body in order to change the natural language of place names.

```
/v1/search?size=1&sources=wof&lang=es&text=cape town, south africa

1)	Ciudad del Cabo, Sudáfrica
```
```
/v1/search?size=1&sources=wof&lang=ja&text=cape town, south africa

1)	ケープタウン, 南アフリカ
```
```
/v1/search?size=1&sources=wof&lang=en&text=cape town, south africa

1)	Cape Town, South Africa
```

the user can request a language as per https://github.com/pelias/api/pull/819, then when the response document is generated, we query the new 'language service' to get all the translations of those places names (administrative areas only).

if possible, the names are changed and the user is returned a document which has been translated to their own language as best as we can, based on the data we have from our providers.

configuration of the 'language service' is almost identical to the interpolation service, by default, and in the situation of a network disruption the service is disabled and the strings are left unmodified.

to configure the service you must add a new new stanza to your `~/pelias.json` config as such:

```
  "language": {
    "client": {
      "adapter": "http",
      "host": "http://localhost:6100"
    }
  },
```

this will enable the service and queries will be sent to the HTTP address specified.

the service interface has been initially designed to work with https://github.com/pelias/placeholder, so in order to get it set up you'll need to install and run that docker as a service, the minimum steps are:

```
# download data

mkdir -p /data/pelias/placeholder

curl -s http://missinglink.geo.s3.amazonaws.com/fts.sqlite3.gz | gzip -d > /data/pelias/placeholder/fts.sqlite3
```

```
# run service

git clone git@github.com:pelias/placeholder.git && cd placeholder

docker-compose up -d
```

the service should now be running, you can test it here http://localhost:6100/parser/findbyid?ids=101750367

note: the service can be configured to read it's data from different directories and run on different ports, see the `placeholder` repo for more configuration options.

### DIFF
--- a/middleware/changeLanguage.js
+++ b/middleware/changeLanguage.js
@@ -1,0 +1,172 @@
+
+var logger = require( 'pelias-logger' ).get( 'api' );
+var service = require('../service/language');
+
+/**
+example response from language web service:
+{
+  "101748479": {
+    "wofid": 101748479,
+    "placetype": "locality",
+    "iso": "DE",
+    "area": 0.031614,
+    "lineage": {
+      "continent_id": 102191581,
+      "country_id": 85633111,
+      "county_id": 102063261,
+      "locality_id": 101748479,
+      "macrocounty_id": 404227567,
+      "region_id": 85682571
+    },
+    "rowid": 90425,
+    "names": {
+      "default": "München",
+      "eng": "Munich"
+    }
+  },
+}
+**/
+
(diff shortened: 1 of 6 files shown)


missinglink:
jpn

PR #834, opened 2017-04-04
Sort interpolated results to show address first

dianashk:
Sometimes multiple streets are sent to the interpolation service for evaluation and only one comes back with an address hit. In that case, if that successful hit was not the first street, it will remain in its position in the list, even though that's clearly the correct result and should be at the top of the list.

This change sorts the interpolated results to ensure that any address records show up above all others.

### DIFF
--- a/middleware/interpolate.js
+++ b/middleware/interpolate.js
@@ -43,6 +43,13 @@ function setup() {
         res.data = results;
       }
 
+      // sort the results to ensure that addresses show up higher than street centroids      
+      res.data = res.data.sort((a, b) => {
+        if (a.layer === 'address' && b.layer !== 'address') { return -1; }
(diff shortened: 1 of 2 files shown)


dianashk:
I ran the TriMet test data against prod and then again against dev with this branch of API deployed, and here are the results.

* tests were created with `priorityThresh=1` and  `distanceThresh=50m`*

#### production

**Aggregate test results**
        * Pass: 2777
        * Improvements: 0
        * Fail: 0
        * Placeholders: 0
        * Regressions: 1290
        * Took 302078ms
        * Test success rate 68.28%

#### sort-interpolated-results

**Aggregate test results**
        * Pass: 2799
        * Improvements: 0
        * Fail: 0
        * Placeholders: 0
        * Regressions: 1268
        * Took 475025ms
        * Test success rate 68.82%

#### 22 tests improved!

PR #841, opened 2017-04-06
500 errors returned when no language service is found

dianashk:
Getting this error when language service is not configured:

```
500 cb is not a function
```

This is happening because copy/paste. :)


missinglink:
d'oh!

PR #846, opened 2017-04-13
Remove unused reverse fixture

orangejulius:
This must have been missed when we removed boundary.circle a long time
ago.


orangejulius:
The tests for this will pass once https://github.com/pelias/api/pull/844 is merged

PR #852, opened 2017-04-19
pass DNT to pip service and sanitize error messages

trescube:
passes `dnt` header to PIP service when requested at API by caller.  Also sanitizes identifiable parameters from error messages.

### DIFF
--- a/service/pointinpolygon.js
+++ b/service/pointinpolygon.js
@@ -2,14 +2,51 @@ const logger = require( 'pelias-logger' ).get( 'pointinpolygon' );
 const request = require('request');
 const _ = require('lodash');
 
+function sanitizeUrl(requestUrl) {
+  return requestUrl.replace(/^(.+)\/.+\/.+$/, (match, base) => {
+    return `${base}/[removed]/[removed]`;
+  });
+}
+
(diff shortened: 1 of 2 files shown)


service/pointinpolygon.js
```diff
@@ -2,14 +2,51 @@ const logger = require( 'pelias-logger' ).get( 'pointinpolygon' );
 const request = require('request');
 const _ = require('lodash');
 
+function sanitizeUrl(requestUrl) {
+  return requestUrl.replace(/^(.+)\/.+\/.+$/, (match, base) => {
+    return `${base}/[removed]/[removed]`;
+  });
+}
```
missinglink: no tests for `sanitizeUrl()`, regex's always have edge cases?

trescube: There are tests [here](https://github.com/pelias/api/pull/852/files#diff-f1db4ccb690ad5f0b8838fc6aece9abbR213) and [here](https://github.com/pelias/api/pull/852/files#diff-f1db4ccb690ad5f0b8838fc6aece9abbR298).

missinglink: oh, missed those sorry

PR #884, opened 2017-05-10
Remove gid properties from any interpolated results

dianashk:
Connected to pelias/pelias#537

This is the first step towards resolving the interpolated gid issue. Currently the gid values are misleading because they can not be used in the `/place` endpoint, which technically is their only purpose for existing. We need to hide them until we can implement a solution that allows the reconstruction of these records via the `/place` endpoint.

@msmollin: heads up on the gid properties going away for some but not all results. let us know if this has any negative implications on mobile. ☎️ 

@louh: once this is in staging we should test the demo app and make sure it doesn't break anything.

### DIFF
--- a/helper/geojsonify_meta_data.js
+++ b/helper/geojsonify_meta_data.js
@@ -8,7 +8,14 @@ var Document = require('pelias-model').Document;
  */
 function addMetaData(src, dst) {
   dst.id = src._id;
-  dst.gid = makeGid(src);
+
+  // at this time the gid's of interpolated addresses are not usable in the place endpoint
+  // to avoid confusion they will be removed altogether for now and replaced in the future when we have determined
+  // the right format for these types of results
+  if (src.source !== 'mixed') {
+    dst.gid = makeGid(src);
+  }
+
   dst.layer = lookupLayer(src);
   dst.source = lookupSource(src);
   dst.source_id = lookupSourceId(src);
--- a/test/unit/helper/geojsonify.js
+++ b/test/unit/helper/geojsonify.js
@@ -459,6 +459,53 @@ module.exports.tests.categories = function (test, common) {
   });
 };
 
+
+module.exports.tests.interpolated_gids = function (test, common) {
+  test('do not set the gid for interpolated results', function (t) {
+    var input = [
+      {
+        '_id': 'polyline:10439991',
+        'source': 'mixed',
+        'layer': 'address',
+        'center_point': {
+          'lon': -73.953698,
+          'lat': 40.757667
+        },
+        'name': {
+          'default': '123 Main Street'
+        },
+        'source_id': 'polyline:10439991'
+      }
+    ];
+
+    var expected = {
+      'type': 'FeatureCollection',
+      'bbox': [-73.953698, 40.757667, -73.953698, 40.757667],
+      'features': [
+        {
+          'type': 'Feature',
+          'properties': {
+            'id': 'polyline:10439991',
+            'layer': 'address',
+            'source': 'mixed',
+            'source_id': 'polyline:10439991',
+            'name': '123 Main Street'
+          },
+          'geometry': {
+            'type': 'Point',
+            'coordinates': [ -73.953698, 40.757667 ]
+          }
+        }
+      ]
+    };
+
+    var json = geojsonify( {}, input );
+
+    t.deepEqual(json, expected, 'gid is not defined');
+    t.end();
+  });
+};
+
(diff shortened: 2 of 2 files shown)


msmollin:
Thanks for the heads up. We'll have to plan releases around this, as currently both clients expect GID to be part of the `Feature` returned. We have an Android SDK release tentatively scheduled for next week barring any issues testing 0.6.2 of Tangram-es. We can probably work this into that although I'll have to confirm with @sarahlensing. The iOS SDK release is a few days to a week behind that depending on how quickly we can get thru the backlog.

How quickly are you planning on rolling this to prod?

missinglink:
copied from: https://github.com/pelias/pelias-doc/pull/185

---

- The current behavior is that we generate a `gid` value for *all* records.
- This is misleading as those `gid` values created for interpolated results cannot be used to retrieve a stored record using the `/v1/place` endpoint.
- The confusion is that, for *all other* `gid` values you are able to retrieve a stored document using the `gid`, this is an exception to that rule.
- I would like to start a discussion around removing those `gid` values from the results, but this is yet to be decided as the cons are that it breaks consistency.

The behavior I would like to clear up is that, there is a logical assumption that when a `gid` is provided that it can be used with the `/v1/place` endpoint, since the interpolation feature this is no longer true.

This can be cleared up a few different ways:
- Better documentation.
- Removing the `gid` value from interpolated results.
- Somehow supporting interpolated `gid` values in the `/v1/place` endpoint.
- Emitting a better warning or error message when using these gid values with `/v1/place`

---

:+1: This sounds like a good interim measure.

helper/geojsonify_meta_data.js
```diff
@@ -8,7 +8,14 @@ var Document = require('pelias-model').Document;
  */
 function addMetaData(src, dst) {
   dst.id = src._id;
-  dst.gid = makeGid(src);
+
+  // at this time the gid's of interpolated addresses are not usable in the place endpoint
+  // to avoid confusion they will be removed altogether for now and replaced in the future when we have determined
+  // the right format for these types of results
+  if (src.source !== 'mixed') {
```
missinglink: the `source` returned by the interpolation service can be one of `mixed, openstreetmap, openaddresses`.

it is currently possible, due to builds being performed at different times, that an ID returned via the interpolation service (for OSM or OA) is not retrievable via the `/v1/place` endpoint.

see: https://github.com/pelias/api/blob/master/middleware/interpolate.js#L108-L111

I suspect this is more of an exception than the norm (more common for missing unit numbers than missing house numbers), if you think it might be a problem then we can consider removing the `gid` for all results coming back from the interpolation service by targeting `hit.match_type = 'interpolated'`.

see: https://github.com/pelias/api/blob/master/middleware/interpolate.js#L102

missinglink: interestingly the middleware code has no logic to handle the case of `TIGER` data, instead it is marked as `mixed`, this is not technically correct but it's satisfactory because it would also be impossible to retrieve this content via `/v1/place`.

msmollin:
A couple more thoughts came out of discussion on this - making this field optional will most assuredly break Tizen's client of Pelias. I believe @kevinkreiser wrote that originally? 

Any chance we can version the API?

kevinkreiser:
is this relevant to the discussion: https://github.com/mapzen/tizen-maps-plugin-mapzen/blob/c22c22f61fd9686748f84bfe54cdfae54f97f712/src/mapzen/mapzen_jsonparser.cpp#L305

dianashk:
Thanks @msmollin and @kevinkreiser for your input. We certainly don't want to break any clients. Since we eventually want to put back the gid for interpolated results, it seems a bit heavy handed to version the API. We will discuss alternatives. Perhaps we could just add a warning instead of error so people can still call the place endpoint with the gid and not be confused about the behavior.

msmollin:
So @sarahlensing and I went over the ios and android SDKs with a comb and it turns out what I thought originally about this breaking those SDKs is actually incorrect! The Android SDK sets the gid field to an empty string by default and expects that by default so that's OK, and the iOS SDK doesn't ever reference the `gid` field in the response object so we're good there. 

Sorry for the confusion.

orangejulius:
If I recall, the other day we discussed this and despite the fact that no Mapzen clients would be broken, we were hesitant to remove `gid`s from responses because they would break our response format in any clients that _were_ hardcoded to expect a `gid`.

Was there anything more? If not, it sounds like we want to close this PR.

PR #885, opened 2017-05-12
Merge multiple lineage into single record with parent arrays

dianashk:
Multiple hierarchies should be treated the same way they are currently treated in the wof importer. All the values are merged into a single record where each parent property is an array of values.


controller/placeholder.js
```diff
@@ -129,6 +125,12 @@ function setup(placeholderService, should_execute) {
           results.filter(layersFilter).map(synthesizeDocs))
```
trescube: You won't need `_.flatten` here anymore since `synthesizeDocs` now only ever returns a single results.

controller/placeholder.js
```diff
@@ -129,6 +125,12 @@ function setup(placeholderService, should_execute) {
...
         .filter(countryFilter);
 
+        const messageParts = [
+          '[controller:placeholder]',
+          `[result_count:${_.get(res, 'data', []).length}]`
```
trescube: Since `data` is a direct child of `res` instead of a nested chain of objects, `_.defaultTo` can be used instead of `_.get`.

trescube:
I'm going to merge since @dianashk is on vacation.

PR #890, opened 2017-05-15
Prefer openaddresses results with zip code

### LINKED ISSUE #872: Duplicate result inconsistency
I have two nearly identical searches, but one uses `focus.point`:

[/v1/autocomplete?text=284%20Broadway,%20Cambridge&layers=address&boundary.country=us](http://pelias.github.io/compare/#/v1/autocomplete%3Flayers=address&boundary.country=us&text=284%20Broadway,%20Cambridge)

[/v1/autocomplete?text=284%20Broadway,%20Cambridge&layers=address&boundary.country=us&focus.point.lat=42.3601&focus.point.lon=-71.0589](http://pelias.github.io/compare/#/v1/autocomplete%3Flayers=address&boundary.country=us&focus.point.lat=42.3601&focus.point.lon=-71.0589&text=284%20Broadway,%20Cambridge)

The first query returns `us/ma/statewide:43976057cecaadf8`, and the second returns `us/ma/city_of_cambridge:9d29920532149cc0`. The latter doesn't have a `postalcode`.

I can't quite figure out why `focus.point` ends up preferring the second version – but it would be cool if the dedupe middleware could pick a winner not just by data source, but also by record completeness.

---

If this is something you think is worth tackling, I'd be happy to throw together a PR. But I might need a bit of guidance.

kevincennis:
Resolves #872 

If duplicate results exist and both are from `openaddresses`, prefer the one with a `zip` when applicable.

---

Note: It's not entirely clear to me if I only need to worry about `address_parts.zip` or if I should also be looking for `address_parts.postalcode`.

`helper/diffPlaces.js` seems to only care about `zip` – but presumably non-US data wouldn't have that property. I'm just not 100% clear on what that data can/will look like.

### DIFF
--- a/middleware/dedupe.js
+++ b/middleware/dedupe.js
@@ -65,6 +65,14 @@ function dedupeResults(req, res, next) {
 function isPreferred(existing, candidateReplacement) {
   // NOTE: we are assuming here that the layer for both records is the same
 
+  var isOA = _.flow(_.property('source'), _.eq.bind(null, 'openaddresses'));
+  var hasZip = _.bind(_.has, null, _.bind.placeholder, 'address_parts.zip');
+
+  // https://github.com/pelias/api/issues/872
+  if (isOA(existing) && isOA(candidateReplacement)) {
+    return hasZip(candidateReplacement) && !hasZip(existing);
+  }
+
   //bind the trumps function to the data items to keep the rest of the function clean
   var trumpsFunc = trumps.bind(null, existing, candidateReplacement);
 
--- a/test/unit/middleware/dedupe.js
+++ b/test/unit/middleware/dedupe.js
@@ -186,6 +186,42 @@ module.exports.tests.trump = function(test, common) {
       t.end();
     });
   });
+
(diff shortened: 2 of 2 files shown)


middleware/dedupe.js
```diff
@@ -65,6 +65,14 @@ function dedupeResults(req, res, next) {
 function isPreferred(existing, candidateReplacement) {
   // NOTE: we are assuming here that the layer for both records is the same
 
+  var isOA = _.flow(_.property('source'), _.eq.bind(null, 'openaddresses'));
```
dianashk: We ❤️ and use lodash here quite a bit, but this is a master class in its application!! 🙇🏻🙇🏻‍♀️

dianashk:
To answer your question about checking the `postalcode` property in addition to zip: that's not necessary. At this point in the flow we only have the zip property you're checking. 
It gets renamed a few steps later in [this middleware](https://github.com/pelias/api/blob/master/middleware/renamePlacenames.js#L9).

PR #899, opened 2017-06-02
Add coarse reverse fallback

trescube:
Fixes pelias/pelias#585

If non-coarse reverse returns nothing, call coarse reverse using all layers.


test/unit/controller/coarse_reverse.js
```diff
@@ -2,6 +2,7 @@
 
 const setup = require('../../../controller/coarse_reverse');
 const proxyquire =  require('proxyquire').noCallThru();
+const _  = require('lodash');
```
dianashk: even in the tests! 😮

PR #900, opened 2017-06-03
only query ES for non-coarse layers on non-coarse reverse requests

trescube:
Currently, /reverse queries ES for all explicit layers or all implied layers for non-coarse requests.  This causes an issue when coarse results can be mixed in with non-coarse results.  For example, in this request, a coarse result (geonames localadmin) is returned 1st in combination with addresses:

http://pelias.github.io/compare/#/v1/reverse%3Fpoint.lat=39.898609&point.lon=-76.607735

In this request, a WOF locality is returned 2nd along with a number of addresses:

http://pelias.github.io/compare/#/v1/reverse%3Flayers=venue,locality&point.lat=39.898609&point.lon=-76.607735

To continue moving towards querying for addresses, venues, and streets, then falling back to admin areas, this PR modifies reverse ES queries as follows:

- if no `layers` parameter is specified, ES is only queried for `address`, `street`, and `venue`
- if the `layers` parameter is specified, ES is only queried for the intersection of the `layers` array and `['venue', 'address', 'street']`

This will cause ES to only be queried for non-coarse results.  If no results are returned, the PiP service will be called.  

This also has the side benefit of removing geonames entirely from the /reverse endpoint.  

This may also eliminate the need to have two coarse calls since they would have the same effect when the non-coarse ES request is either explicitly not called or returns 0 results.

### DIFF
--- a/test/unit/fixture/reverse_with_layer_filtering_non_coarse_subset.js
+++ b/test/unit/fixture/reverse_with_layer_filtering_non_coarse_subset.js
@@ -3,22 +3,24 @@ var vs = require('../../../query/reverse_defaults');
 module.exports = {
   'query': {
     'bool': {
-      'filter': [{
-        'geo_distance': {
-          'distance': '500km',
-          'distance_type': 'plane',
-          'optimize_bbox': 'indexed',
-          'center_point': {
-            'lat': 29.49136,
-            'lon': -82.50622
+      'filter': [
+        {
+          'geo_distance': {
+            'distance': '500km',
+            'distance_type': 'plane',
+            'optimize_bbox': 'indexed',
+            'center_point': {
+              'lat': 29.49136,
+              'lon': -82.50622
+            }
+          }
+        },
+        {
+          'terms': {
+            'layer': ['venue', 'street']
           }
         }
-      },
-      {
-        'terms': {
-          'layer': ['venue', 'address', 'street']
-        }
-      }]
+      ]
     }
   },
   'sort': [
--- a/query/reverse.js
+++ b/query/reverse.js
@@ -3,6 +3,7 @@
 const peliasQuery = require('pelias-query');
 const defaults = require('./reverse_defaults');
 const check = require('check-types');
+const _ = require('lodash');
 const logger = require('pelias-logger').get('api');
 
 //------------------------------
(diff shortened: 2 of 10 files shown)


trescube:
See pelias/acceptance-tests#389

test/unit/fixture/reverse_with_layer_filtering_non_coarse_subset.js
```diff
@@ -0,0 +1,41 @@
+var vs = require('../../../query/reverse_defaults');
+
+module.exports = {
+  'query': {
+    'bool': {
+      'filter': [
+        {
+          'geo_distance': {
+            'distance': '500km',
```
orangejulius: Our previous experiences with slow reverse queries suggest that anything that can query over such a large distance has the potential to be very slow. Just to be sure, this 500km is only here because of the `boundary.radius` set in the test, right? Even so, we might want to lower it. Even setting the example of a 500km radius search for all venues and streets could cause us to write non-performant code someday.

trescube: The plan is to remove the initial coarse reverse (we discussed this in standup) since non-coarse reverse now is limited to just non-coarse `layers` scenearios.

orangejulius: Yeah I like that plan (I mentioned it in my overall comment on the review). Basically I'm asking two things:

1.) Can you confirm that the 500km here comes from line 141 in `test/unit/query/reverse.js` and
2.) Can we remove that line or lower the value?

trescube: I can't speak to why the test says `500km` but it makes sense to update to something much smaller.  This is just a unit test, they don't actually query ES.  

This sanitizer can probably be cleaned up since `boundary:circle:radius:coarse` doesn't mean anything anymore:  

https://github.com/pelias/api/blob/master/sanitizer/_geo_reverse.js#L33-L37

orangejulius: Oh, yes i guess it's actually line 161 that came from 141, which is probably a very old line that dates back to the days when we actually did query in a 500km radius. I think it's important to remove but I'll do it in a separate PR so not to distract from here.

PR #904, opened 2017-06-06
Cleanup reverse artifacts

trescube:
Updates:

- add warning/errors when `geonames` or `gn` is specified in `sources`
- remove support for `boundary.circle.*` for coarse /reverse requests

### DIFF
--- a/controller/coarse_reverse.js
+++ b/controller/coarse_reverse.js
@@ -103,6 +103,11 @@ function setup(service, should_execute) {
       return next();
     }
 
+    // return a warning to the caller that boundary.circle.radius will be ignored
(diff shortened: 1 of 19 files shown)


trescube:
The /reverse endpoint hasn't supported boundary.circle for coarse requests since mid-April when the pip service was released.  All this PR does is remove the impression that it does by explicitly not honoring the parameter.

dianashk:
Can we add a warning in the case of boundary.circle.radius being specified but not used? Like if the query is coarse or if we ended up falling back to pip after not finding anything in ES.

trescube:
Sure, I can do that.

missinglink:
:kiss:

PR #906, opened 2017-06-08
added note about placeholder root

trescube:
only turned out to be small documentation update


trescube:
Closed in favor of #907

PR #918, opened 2017-06-30
set query_type for placeholder to enable confidencescores

### LINKED ISSUE #917: Queries going through the Placeholder API do not have confidence scores
It looks as if no search queries that go through the Placeholder API have a confidence score. For example, this query for [london, uk](http://pelias.github.io/compare/#/v1/search%3Ftext=london,%20uk):
screenshot from 2017-06-30 11-32-40

This means that Pelias is breaking the contract of what parameters we promise to return and **anyone who hardcoded an expectation that confidence score will always be there is having problems**, so we should fix it pretty soon.

Queries via autocomplete or with a full address appear to be fine.

For example [30 W 26th St, New York, NY](http://pelias.github.io/compare/#/v1/search%3Ftext=30%20W%2026th%20st,%20new%20york,%20ny)
screenshot from 2017-06-30 11-34-10

_Thanks to @louisgillies for originally reporting this in https://gitter.im/pelias/pelias_

trescube:
Fixes #917

### DIFF
--- a/controller/placeholder.js
+++ b/controller/placeholder.js
@@ -235,7 +235,10 @@ function setup(placeholderService, should_execute) {
         // convert results to ES docs
         // boundary.country filter must happen after synthesis since multiple
         //  lineages may produce different country docs
-        res.meta = {};
+        res.meta = {
+          query_type: 'fallback'
+        };
+
         res.data = results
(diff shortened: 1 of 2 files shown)


trescube:
not sure

Issue #762, opened 2016-12-09
Deprecated queries removed in ES5


cag:
I've discovered that while most of the functionality of the API server works just fine on ES5, some of the queries to ES on the API server now have to be rewritten, as they were on a deprecation schedule. The one I've seen so far is the [filtered](https://www.elastic.co/guide/en/elasticsearch/reference/2.3/query-dsl-filtered-query.html) query, which has been deprecated in favor of bool. Just a heads up.

orangejulius:
Good to know. I'm surprised we haven't caught that already since it's already deprecated in 2.3. The good news is we can fix it without breaking 2.3 compatibility, which we should definitely do.

PR #923, opened 2017-07-06
fix(package): update pelias-query to version 8.16.1

orangejulius:
pelias-query 8.16.1 includes working Elasticsearch 5 queries, and my additional commit updates our test fixtures to use them.

Fixes https://github.com/pelias/api/issues/762
Fixes #844
Connects https://github.com/pelias/pelias/issues/461


orangejulius:
Yeah, getting rid of the fixtures would have made this a no-op

PR #925, opened 2017-07-08
fix(package): update pelias-model to version 4.10.1

trescube:
Closes #924


trescube:
I'm merging anyway since it was greenkeepers' anyway.

PR #928, opened 2017-07-10
Configure for Location Bias - set focus.point.lat and focus.point.lon

tigerlily-he:
### Overview
Adds the ability to set a default `focus.point.lon` and `focus.point.lat`
Fixes https://github.com/pelias/pelias/issues/569

### Things to Note
- new configuration parameters:
      - `api.defaultParameters['focus.point.lon']`: default longitude coordinate
      - `api.defaultParameters['focus.point.lat']`: default latitude coordinate
- Default focus point will only be applied if there isn't already a `focus.point.lat` and `focus.point.lon` specified in raw. This way, a call to the API with a different lat/lon can still override the default coordinates in the config 
- The location bias sanitizer is used for `search`, `structured_geocoding` and `autocomplete` only.

### Other Changes
- Added tests to schema test to check valid location bias lat/lon inputs 
- Moved `sanitizers` list and function inside `modules.export.middleware` in order to pass the parameter `_api_pelias_config` to the location bias sanitizer
-  In `tests/unit/autocomplete.js`, verify that all sanitizers for autocomplete were called as expected with proxyquire  instead of checking just the list of sanitizers.

### DIFF
--- a/test/unit/sanitizer/_location_bias.js
+++ b/test/unit/sanitizer/_location_bias.js
@@ -59,6 +59,42 @@ module.exports.tests.setLocationBias = function(test, common) {
       t.end();
   });
 
+  test('raw has focus.point.lon already', t => {
+    const defaultParameters = {
+      'focus.point.lon': 12.2121212,
+      'focus.point.lat': 12.2121212
+    };
+    const locationBias = setup(defaultParameters);
+    const raw = {
+      'focus.point.lon': 43.4343434
+    };
+    const expected = {
+      'focus.point.lon': 43.4343434
+    };
+
+    locationBias(raw, undefined);
+      t.deepEqual(raw, expected, 'should be unmodified' );
+      t.end();
+  });
+
+  test('raw has focus.point.lat already', t => {
+    const defaultParameters = {
+      'focus.point.lon': 12.2121212,
+      'focus.point.lat': 12.2121212
+    };
+    const locationBias = setup(defaultParameters);
+    const raw = {
+      'focus.point.lat': 34.3434343
+    };
+    const expected = {
+      'focus.point.lat': 34.3434343
+    };
+
+    locationBias(raw, undefined);
+      t.deepEqual(raw, expected, 'should be unmodified' );
+      t.end();
+  });
+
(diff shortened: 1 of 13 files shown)


dianashk:
Haven't reviewed the code yet, but 🥇 🏆 for that PR description!!!

test/unit/sanitizer/_location_bias.js
```diff
@@ -0,0 +1,109 @@
...
+    const raw = {};
+    const expected = {};
+
+    locationBias(raw, undefined);
+      t.deepEqual(raw, expected, 'should be unmodified' );
+      t.end();
+  });
+
+  test('raw has focus.point.lon already', t => {
+    const defaultParameters = {
```
orangejulius: I was wondering about the case where there is a default set but the parameters are not valid. It's awesome that it works exactly like before and even more awesome that its covered in the tests.

tigerlily-he: In the schema test (`test/unit/schema.js`) there is a check that non-numeric default parameter input in the config will throw an error.

In the case that the user calls the API with invalid parameters (ex. /v1/search?focus.point.lon=string&text=30 w 26th street), there is a warning.

orangejulius: I saw those too. I was thinking of the case where the default parameters are specified correctly, but the URL parameters are not. I was really happy to see that the tests cover this case, and the actual behavior (return an error just like if there were no default params specified) is great too.

tigerlily-he:
This is the first instance where query default values can be changed through `pelias.json`.  The default focus point will always be specific to the user (ie. there is no universally applicable focus point). By exposing this setting, we intentionally allow users to alter the focus point. Putting it in the config file is more accessible to non-developers.

Another place where default parameters can live in is another config json file. As a separate piece, the default parameters won't clutter `pelias.json`, whose responsibility is to connect services for the API.

Developers can also fork the API and change/add defaults as Peter suggested. This would keep everything together and there won't be a need to worry about hierarchy for overriding settings if the same property is set in multiple places.

Issue #279, opened 2015-09-18
Should warn if any unexpected parameters are found in query.


missinglink:
# error when specifying singular of 'sources' or 'layers'

I frequently make the mistake of setting 'source' instead 'sources'; same for 'layer', the API responds without a warnings or errors which causes me much confusion until I catch my typo.

We should at very least `warn`, but ideally `error` in this case so we `fail fast, fail early`

dianashk:
Should warn if any unexpected parameters are found in query.

orangejulius:
This issue has cropped up several times in the past few months, and it really feels like we should prioritize it for the near future. I believe we should first warn, and then perhaps a few weeks later error on any unknown parameters.

## How do unexpected parameters cause problems

In addition to the general case @missinglink mentioned where parameters are not quite correct, there are two additional problems specifically for Mapzen Search.

If a user intends to send along the `api_key` parameter, but instead uses `apikey`, `apiKey` or another similar but unknown parameter, they will fall back to the keyless rate limit, which is currently very low. It's pretty hard to trace down this issue.

Another problem we've seen is users sending along parameters that don't actually exist, but were presumably passed along from one of their internal APIs. For example someone might send along a "session_id" or "user_id" parameter. This doesn't change the results, but it does cause them to lose any benefits of caching. It increases the load on Mapzen Search and also prevents that user from getting "free" requests that were already cached.

## How should we implement this?

Right now there's no central place in any of our code that actually knows what all the parameters the API accepts are. We obviously have lists in our documentation, but that's really the only place.

A big concern about the implementation is that if the list of "acceptable" parameters gets out of sync with what parameters are actually allowed, bad things will happen. It would also be annoying and error-prone to have to maintain this list somewhere. Therefore it feels like the implementation for parameter verification _must_ involve building a way for the API to know what parameters are possibly used.

I wonder if we could build this into our sanitizers. Right now, sanitizers are passed the entire `raw` query object. If they were made so that they had to register ahead of time which parameters they used, then the code that currently runs all the sanitizers would have the full list of parameters, and could use that to validate all the params.

tigerlily-he:
Will unexpected properties in `raw` get sanitized and passed onto `clean`? (ie. `clean['layer'] = raw['layer']`) 

If so, would it be sufficient to check that every property in `raw` is also in `clean`? Any property from `raw` that is not found in `clean` would be an unexpected parameter. This way, there is no need to create a list of acceptable parameters to check against. This sanitizer can run after all other sanitizers.

PR #931, opened 2017-07-13
Fix warnings for boundary.circle.radius

### LINKED ISSUE #915: Remove coarse reverse boundary.circle.radius warning
Currently all reverse requests that are either explicitly coarse or end up falling back to coarse return a warning of `boundary.circle.radius is not applicable for coarse reverse`.  Due to how the sanitizers work, `boundary.circle.radius` is set to a default if not supplied so by the time the coarse reverse controller runs, it doesn't know that value was explicitly supplied by the user or is a default.  Figure out a way to only add the warning `boundary.circle.radius` was explicitly passed by the user.

tigerlily-he:
Coarse reverse searches should no longer warn about `boundary.circle.radius` not being applicable. Fixes #915. This warning used to occur because if the property was undefined, the default value (1 km) would get applied in the `_geo_reverse` sanitizer. By the time the controller sees the value, it cannot distinguish between a value set by the user or the sanitizer (default). 

~~This PR also fixes pelias/pelias#518. `boundary.rect` parameter does not make sense for nearby and reverse. There could be a conflict if the `point.lat` and `point.lon` is outside the boundary box region. The documentation does not say `boundary.rect` is a supported parameter, so we should just give a warning.~~

### DIFF
--- a/query/reverse.js
+++ b/query/reverse.js
@@ -75,6 +75,18 @@ function generateQuery( clean ){
     logStr += '[param:boundary_circle] ';
   }
 
+  // for coarse reverse when boundary circle radius is undefined
+  if( check.number(clean['boundary.circle.lat']) &&
+      check.number(clean['boundary.circle.lon']) &&
+      check.undefined(clean['boundary.circle.radius']) ){
+    vs.set({
+      'boundary:circle:lat': clean['boundary.circle.lat'],
+      'boundary:circle:lon': clean['boundary.circle.lon'],
+      'boundary:circle:radius': defaults['boundary:circle:radius']
+    });
+    logStr += '[param:boundary_circle] ';
+  }
+
   // boundary country
   if( check.string(clean['boundary.country']) ){
     vs.set({
--- a/query/reverse.js
+++ b/query/reverse.js
@@ -65,25 +65,24 @@ function generateQuery( clean ){
   // where point.lan/point.lon are provided in the
   // absense of boundary.circle.lat/boundary.circle.lon
   if( check.number(clean['boundary.circle.lat']) &&
-      check.number(clean['boundary.circle.lon']) &&
-      check.number(clean['boundary.circle.radius']) ){
-    vs.set({
-      'boundary:circle:lat': clean['boundary.circle.lat'],
-      'boundary:circle:lon': clean['boundary.circle.lon'],
-      'boundary:circle:radius': clean['boundary.circle.radius'] + 'km'
-    });
-    logStr += '[param:boundary_circle] ';
-  }
+      check.number(clean['boundary.circle.lon']) ){
 
-  // for coarse reverse when boundary circle radius is undefined
-  if( check.number(clean['boundary.circle.lat']) &&
-      check.number(clean['boundary.circle.lon']) &&
-      check.undefined(clean['boundary.circle.radius']) ){
-    vs.set({
-      'boundary:circle:lat': clean['boundary.circle.lat'],
-      'boundary:circle:lon': clean['boundary.circle.lon'],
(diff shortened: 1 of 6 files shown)


orangejulius:
So...story time. Long ago we had code somewhat similar to this, where we essentially overwrote raw/clean params, and thus lost knowledge of whether or not we did that or the user actually sent the param later in the code. (I actually went and found it [here](https://github.com/pelias/api/blob/2.2.1/sanitiser/_layers.js) and [here](https://github.com/pelias/api/blob/2.2.1/controller/search.js#L25)) We were eventually able to remove it, but it was fairly confusing and breaks the assumption that raw/clean all come from user entered parameters.

Is it possible instead to set the default in query/reverse if there isn't anything in clean (maybe somewhere around [here](https://github.com/pelias/api/blob/master/query/reverse.js#L84)?), and then keep raw/clean only for user input?

tigerlily-he:
@orangejulius I agree with your concern. Adding a clean property without a raw property even is so messy and ugly. It'll be best to remove the code that overwrites `raw['boundary.circle.radius']` in the first place.

tigerlily-he:
I think what you are saying about catching all unexpected parameters makes a lot more sense. It's much cleaner than warning for specific parameters. I can delete the second commit and work on pelias/api#279, which would resolve pelias/pelias#518.

orangejulius:
Great, it definitely seemed to make more sense to me, I just wanted to make sure I wasn't missing something specific to the `boundary.rect` parameter. Handling all the parameters at once separately should be easier for everyone :)

tigerlily-he:
### Update
Check for undefined `clean['boundary.circle.radius']` in `query/reverse.js` instead of trying to overwrite raw in `sanitizer/_geo_reverse.js`. 

The boundary circle radius will be undefined for coarse reverse, so just set it to the default `'1km' `from `query/reverse_defaults.js`. This value is not actually used during coarse reverse.

The check for user input of boundary circle radius in `controller/coarse_reverse.js` will work as intended (ie. it will give a warning if the user provides boundary circle radius BUT will not complain that there is a default set by the API).

Tests were updated or deleted to reflect new behavior.

query/reverse.js
```diff
@@ -75,6 +75,18 @@ function generateQuery( clean ){
     logStr += '[param:boundary_circle] ';
   }
 
+  // for coarse reverse when boundary circle radius is undefined
+  if( check.number(clean['boundary.circle.lat']) &&
```
dianashk: Any reason not to combine this block with the one above it? Checking for lat/lon and then inside the if checking radius and setting the value accordingly?

query/reverse.js
```diff
@@ -65,13 +65,24 @@ function generateQuery( clean ){
...
-    });
+      check.number(clean['boundary.circle.lon']) ){
+
+        vs.set({
+          'boundary:circle:lat': clean['boundary.circle.lat'],
+          'boundary:circle:lon': clean['boundary.circle.lon']
+        });
+
+        if (check.undefined(clean['boundary.circle.radius'])){
+          // for coarse reverse when boundary circle radius is undefined
```
orangejulius: this could also be the case if it's a non-coarse reverse query but the query didn't specify a boundary.circle.radius, right?

tigerlily-he: Boundary.circle.lat and boundary.circle.lon are [currently unsupported](http://pelias.github.io/compare/#/v1/reverse%3Fboundary.circle.lat=40&boundary.circle.lon=-74&point.lat=40.1&point.lon=-74.1) anyway. There shouldn't be a non-coarse reverse query that has the point coordinates without a radius.

PR #936, opened 2017-07-26
Convert language and interpolation to microservices

trescube:
Converts the following to use [pelias-microservice-wrapper](https://www.npmjs.com/package/pelias-microservice-wrapper):

- `middleware/interpolate.js`
- `middleware/changeLanguage.js`

Introduces [proxyquire](https://www.npmjs.com/package/proxyquire) into testing for these services.  I tried to touch the actual result-handling code as little as possible since it works great as-is, though I removed unreachable conditions and some other minor changes.  

Summary of changes:

- convert `routes/v1.js` to use service configurations and inject `should_execute` predicates into `middleware/interpolate.js` and `middleware/changeLanguage.js`
- predicates (all including tests):
  - `controller/predicates/has_parsed_text_properties.js`: returns true if `req.clean.parsed_test` has `any`/`all` properties
  - `has_request_parameter.js`: return true if the supplied parameter is in req.clean
- added `language` and `interpolation` services into `schema.js`
- grouped service definition tests for schema
- added ServiceConfiguration implementations for language and interpolation services
- removed `service/interpolation.js` and `service/language.js`


middleware/interpolate.js
```diff
@@ -21,132 +20,96 @@ example response from interpolation web service:
...
+
+        // -- metatdata --
+        source_result.layer = 'address';
+        source_result.match_type = 'interpolated';
+
+        // -- name --
+        source_result.name.default = `${interpolation_result.properties.number} ${source_result.name.default}`;
+
+        // -- source --
+        if (interpolation_result.properties.source === 'OSM') {
```
orangejulius: mappings between the various permutations of our source names exist in https://github.com/pelias/api/blob/master/helper/type_mapping.js#L26 to keep all that knowledge in one place. It appears to not have been updated in a long time, for example to support polylines, but it feels like we should use it here.

middleware/interpolate.js
```diff
@@ -64,13 +65,13 @@ function setup(service, should_execute) {
...
-        } else {
-          source_result.source = 'mixed';
-        }
+        // lookup the lowercased source, defaulting to 'mixed' when not found
+        // the source mapping is a jagged string->array, so default to 'mixed' as an array
+        //  to ensure that subscript works
+        source_result.source = _.defaultTo(
+          source_mapping[_.toLower(interpolation_result.properties.source)],
+          ['mixed']
+        )[0];
```
orangejulius: there's a lodash for everything!

middleware/changeLanguage.js
```diff
@@ -28,84 +27,34 @@ example response from language web service:
...
-      if( err ){
-        logger.error( '[language] [error]', err );
-      } else {
-        updateDocs( req, res, translations );
+    service(req, res, (err, translations) => {
+      // if there's an error, log it and bail
+      if (err) {
+        logger.info(`[middleware:language][error]`);
+        logger.error(err);
+        return next();
```
dianashk: Is this a good place to pass the error to next () like we talked about with lily this morning? Not urgent but since we were on the subject.

middleware/changeLanguage.js
```diff
@@ -136,17 +85,14 @@ function updateDocs( req, res, translations ){
...
 
-        // skip invalid records
-        if( !translations[id].hasOwnProperty( 'names' ) ){ continue; }
-
         // requested language is not available
         if (_.isEmpty(_.get(translations[id].names, requestLanguage, [] ))) {
-          logger.debug( '[language] [debug]', 'missing translation', requestLanguage, id );
+          logger.debug( `[language] [debug] missing translation ${requestLanguage} ${id}` );
           continue;
         }
```
dianashk: This file is so much better with all that code stripped out! Yay!!

routes/v1.js
```diff
@@ -97,6 +101,14 @@ function addRoutes(app, peliasConfig) {
   const placeholderService = serviceWrapper(placeholderConfiguration);
   const isPlaceholderServiceEnabled = _.constant(placeholderConfiguration.isEnabled());
 
+  const changeLanguageConfiguration = new Language(_.defaultTo(peliasConfig.api.services.language, {}));
```
dianashk: What's the reasoning for not combining placeholder and language services into one?

service/configurations/Language.js
```diff
@@ -0,0 +1,34 @@
...
+  getParameters(req, res) {
+    // find all the values for all keys with names that end with '_id'
+    const ids = _.get(res, 'data', []).reduce((acc, doc) => {
+      Array.prototype.push.apply(acc, _.values(_.pickBy(doc.parent, (v, k) => _.endsWith(k, '_id') ) ) );
+      return acc;
+    }, []);
+
+    return {
+      // arrays will be nested, so flatten first, then uniqify, and finally join elements with comma
+      ids: _.uniq(_.flattenDeep(ids)).join(',')
```
dianashk: Great explanation and cool new word uniqify

routes/v1.js
trescube: I'm not sure what you mean.  Each has a different URL path and different ways of building URLs.

dianashk: Just that they are pointing to the same service. But I guess the URLs and results are different enough to warrant separate services. We also talked about combining the Pelias config parameters into one service instead of having both language and placeholder.

trescube: Oh, sure, I can do that

middleware/interpolate.js
```diff
@@ -21,132 +21,96 @@ example response from interpolation web service:
...
+        if (!_.has(interpolation_result, 'properties')) {
+          logger.debug(`[interpolation] [miss] ${req.clean.parsed_text}`);
+          return;
+        }
+
+        // the interpolation service returned a valid result, debug log for posterity
+        // note: we now merge those values with the existing 'street' record
+        logger.debug(`[interpolation] [hit] ${req.clean.parsed_text} ${JSON.stringify(interpolation_result)}`);
+
+        // -- metatdata --
```
missinglink: nit: typo

PR #942, opened 2017-07-28
Check for unexpected parameters in query input

### LINKED ISSUE #279: Should warn if any unexpected parameters are found in query.

tigerlily-he:
# Overview
WARN user if an unexpected parameter is found in the query. No more guesswork as to whether the parameter name is "source" or "source**s**", "layer" or "layer**s**"! Fixes #279 
screen shot 2017-07-28 at 4 44 40 pm

## Sanitizer Structural Changes
- All sanitizers export a setup function (whether named or anonymous) that returns an object with the main sanitize function. 
```javascript
// Anonymous
module.exports = () => ({
  sanitize: _sanitize,
  expected: _expected
});

// Named setup
function _setup(param1, param2) {
...
    return {
        sanitize: function _sanitize( raw, clean ) { ...}
        expected: function _expected() {...}
    };
}

modules.export = _setup;
```
- Sanitizers that validate query parameters directly have a new function `_expected` that returns an array of expected parameters. (Ex. `sanitizers/_text.js` returns `[{ name: 'text' }]`) These parameters are combined into one big list called `goodParameters`. `sanitizeAll.checkParameters()` validates whether every parameter passed by the user exists in goodParameters. A warning will get passed into req if an unexpected parameter is encountered.
- The main sanitizer function `sanitizer/sanitizeAll.js` exports three functions: `sanitize()`, `checkParameters()`, and `runAllChecks()`. The last one runs the other two synchronously. A callback is no longer passed into `sanitizeAll.js`.
- Tests that use proxyquire have their mock dependencies modified to the new format. 

## Consistency refactoring
- Required sanitizer modules in tests are called `sanitizer`
- Setup function is called every time a sanitizer module is required. (Ex. `sources_and_layers: require('../sanitizer/_sources_and_layers')()`). Uses `sanitizer.sanitize(raw, clean)` instead of `sanitizer(raw, clean)`
- the sanitize function in each of the sanitizers (excluding `sanitizeAll`) only takes in `raw, clean` as arguments. 
- All sanitizer wrapper tests use proxyquire to test whether the sanitizers were called. The effect of the sanitizers are tested in individual sanitizer tests.

### DIFF
--- a/sanitizer/_flag_bool.js
+++ b/sanitizer/_flag_bool.js
@@ -41,7 +41,7 @@ function _setup( paramName, defaultValue ) {
     }, // end of _sanitize function
 
     expected: function _expected(){
-      return [{ name: 'private'}];
+      return [{ name: opts.paramName}];
     } // end of _expected function
   }; // end of return object
 } // end of _setup function
--- a/sanitizer/_flag_bool.js
+++ b/sanitizer/_flag_bool.js
@@ -5,38 +5,46 @@ var _ = require('lodash');
  *
  * @param {string} paramName name of parameter being sanitized
  * @param {boolean} defaultValue value to set variable to if none specified
- * @returns {Function}
+ * @returns {Object} object containing functions
  */
-function setup( paramName, defaultValue ) {
-  return function( raw, clean ){
-    return sanitize( raw, clean, {
-      paramName: paramName,
-      defaultValue: defaultValue
-    });
+function _setup( paramName, defaultValue ) {
+  /**
+   * {object} opts
+  */
+
+  const opts = {
+    paramName: paramName,
+    defaultValue: defaultValue
   };
-}
 
-/**
- * Validate inputs, convert types and apply defaults
- *
- * @param {object} raw
- * @param {object} clean
- * @param {object} opts
- * @returns {{errors: Array, warnings: Array}}
- */
-function sanitize( raw, clean, opts ){
+  return {
+    /**
+     * Validate inputs, convert types and apply defaults
+     *
+     * @param {object} raw
+     * @param {object} clean
+     * @returns {{errors: Array, warnings: Array}}
+     */
 
-  // error & warning messages`1
-  var messages = { errors: [], warnings: [] };
+    sanitize: function _sanitize( raw, clean){
 
-  if( !_.isUndefined( raw[opts.paramName] ) ){
-    clean[opts.paramName] = isTruthy( raw[opts.paramName] );
-  }
-  else {
-    clean[opts.paramName] = opts.defaultValue;
-  }
-  return messages;
-}
+      // error & warning messages`1
+      var messages = { errors: [], warnings: [] };
+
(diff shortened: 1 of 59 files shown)


sanitizer/sanitizeAll.js
```diff
@@ -41,8 +41,8 @@ function checkParameters(req, sanitizers, cb) {
 
     // checks if there is a function that returns valid params
     if (typeof sanitizers[s].expected === 'function'){
-      /** func returns {array} ex: [{ name: 'text' }, { name: 'parsed_text' }] */
-      for (let t in sanitizers[s].expected()) {
+      /** func returns {array} ex: [{ name: 'text' }] */
+      for (var t in sanitizers[s].expected()) {
```
orangejulius: if you are having to change `let` to `var` to make the travis tests work it's probably because `'use strict';` is not at the top of the file. Unfortunately Node.js 4 (but not Node.js 6) requires this in order to enable some ES6 features. Definitely use `let`

test/unit/sanitizer/sanitizeAll.js
```diff
@@ -140,11 +147,97 @@ module.exports.tests.all = function(test, common) {
...
-    sanitizeAll(req, sanitizers, function(){
-      t.deepEquals(req, expected_req);
-      t.end();
-    });
+    sanitizeAll.runAllChecks(req, sanitizers);
+    t.deepEquals(req, expected_req);
+    t.end();
+  });
+
+  test('unexpected parameters should throw warning', function(t) {
```
dianashk: Using the word throw is misleading because it usually implies am exception is getting thrown and needs to be caught. Could lead to misunderstanding of the code

helper/debug.js
```diff
@@ -0,0 +1,36 @@
+'use strict';
```
dianashk: the debug class shouldn't be part of this commit, right?

tigerlily-he: good catch! I accidentally included it when I rebased. I've deleted the file from the commit.

sanitizer/_flag_bool.js
```diff
@@ -5,38 +5,46 @@ var _ = require('lodash');
...
+        clean[opts.paramName] = isTruthy( raw[opts.paramName] );
+      }
+      else {
+        clean[opts.paramName] = opts.defaultValue;
+      }
+      return messages;
+    }, // end of _sanitize function
+
+    expected: function _expected(){
+      return [{ name: 'private'}];
```
orangejulius: 'private' needs to be changed to come from `opts.paramName`, right?

tigerlily-he: Yep you're right. Fixed :)

PR #944, opened 2017-08-03
Debug Mode for Debugging 

tigerlily-he:
The debug mode can be activated by passing `debug=true` into the query. This PR sets up the framework for enabling this option. 

## How to Enable Debug Mode
In the API call, set `debug=` to any of the following: `true, 'TRUE', '1', 1, 'tRuE'`. By default this flag is turned off. Future plans include passing in an array to specify which debug messages to show (ex. `[controller:search, middleware:distance]`) and enabling debug mode from pelias-config

### Initialize the class 
```javascript
const Debug = require('../helper/debug');
const debugLog = new Debug('controller:search'); // type:name. defaults to 'unnamed module'
```

### Add a debug message
To push a message, pass `req` as the first argument and the message as the second
```javascript
debugLog.push(req, 'myMsg'); 
/* adds an object to req. The second argument can be string, number, array, or object.
req = {
    debug: [
        { "controller:search" : "myMsg"}
    ]
}
*/
```
### Start and Stop Timer
```javascript
debugLog.beginTimer(req, 'Msg 1'); // Records Date.now(). Message is optional
/*
req = {
    debug: [
        { "controller:search" : "Timer Began: Msg 1"}
    ]
}
*/
debugLog.stopTimer(req, 'Msg 2'); // calculates time difference
/*
req = {
    debug: [
        { "controller:search" : "Timer Began: Msg 1"},
        { "controller:search" : "Timer Stopped: 12 ms: Msg 2"}
    ]
}
*/
```
Relates to pelias/pelias#207


controller/search.js
```diff
@@ -28,6 +30,7 @@ function setup( apiConfig, esclient, query, should_execute ){
 
     // if there's no query to call ES with, skip the service
     if (_.isUndefined(renderedQuery)) {
+      debugLog.stopTimer(req, 'No query to call ES with. Skipping');
```
dianashk: Great idea adding the message to the stop timer function

orangejulius: yeah, this is pretty cool

sanitizer/search_fallback.js
```diff
@@ -1,29 +0,0 @@
-var sanitizeAll = require('../sanitizer/sanitizeAll'),
```
orangejulius: Was this file not used anywhere? If so, and it's not too much work, feel free to submit a separate PR to remove it. Many small PRs > one bigger one.

tigerlily-he: I included it by accident in #942. It should have been deleted at the point of PR 941 with placeholder integration.

dianashk: yeah I agree with @orangejulius here. it seems like a big and unrelated change to make as part of the debugging work. would be best in a separate PR.

controller/search.js
missinglink: @tigerlily-he what's happens if you have two timers running concurrently? If the message is optional then running two service requests concurrently could look like this:

```bash
01s Timer Began
01s Timer Began
02s Timer Stopped
99s Timer Stopped
```

and you wouldn't know which was which?

It could be as simple as just discouraging people from leaving the message blank.

another option is to store each timer as a variable, such as:

```javascript
var timer = debugLog.beginTimer(req, 'Msg 1');

... timely code

timer.end();
```

helper/debug.js
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  push(req, debugMsg){
+    if (!_.isEmpty(req.clean) && req.clean.enableDebug){
+      req.debug = req.debug || [];
+      // remove the extra space character
+      req.debug.push({[this.name]: debugMsg});
+    //  req.debug.push(`[${this.name}] ${debugMsg}`);
+    }
+  }
```
missinglink: It would be great if there was 0 performance penalty when debugging was disabled.

As it is, there is a need to perform 'work' on the message to compose and format the error string, this is also true of objects and may result in timely code and external function calls which end up being needless if debugging is enabled.

for example:
```javascript
debug.push( req, foo.getName() + " timer took " + ( baz / Math.round( bar ) ) )
```

In this case ^ the production environment will pay the cost of the function call `getName()` plus the math functions when in the end, none of it is returned to the user.

A nice clean way of minimizing the performance penalty when debugging is turned off, is to only allow users to pass **functions** to `push()`

for example:

```javascript
debug.push( req, function(){
   // do some work, but only if debugging is enabled
});
```

or if you decide to pass `req` in the constructor:

```javascript
debug.push( function( req ){
   return foo.getName() + " timer took " + ( baz / Math.round( bar ) )
});
```

there would be a minor cost associated with creating the function, but the time taken to compose and format the message would only be paid when debugging was enabled, not when it's disabled (in production).

another thing to consider is how timers are used, if another debug message is defined while a timer is running, then the cost of the debugging code will be included in the timer, which will make the timers far less intuitive when tracking down slow code paths.

helper/debug.js
```diff
@@ -0,0 +1,43 @@
+'use strict';
+const _ = require('lodash');
+
+class Debug {
+  constructor(moduleName){
+    this.name = moduleName || 'unnamed module';
+  }
```
dianashk: can't do that because the `req` object is different for every call to push. the constructor happens at startup and the same object is used for all debugging in a given module.

missinglink: I see, I guess if there are lots of calls in the same file that a developer could simply:

```javascript
const debug = new Debug( 'foo' );
const push = debug.push.bind( debug, req );
```

missinglink: I guess I'm feeling a little uneasy about this being a class, there's not really any reason for it to be so?

I think of classes as being a `'structure which contains state'` and that methods are used to `'operate on and modify internal state'`, in this case the state is being stored on the `req` object and so the functions are better off to be either:

1. next to the data (ie: as methods of `$req`)
2. as pure functions (ie. in a module somewhere, exposed as functions)

PR #946, opened 2017-08-07
More debug logs to controllers

tigerlily-he:
The debugger module, `helper/debug.js` accepts a function as an argument of the `push` function. This change allows for minimizing performance penalties when debug mode is turned off. The function that generates the message is defined but not executed. 

All controller predicates return two properties: `reply`: the true/false response and `stack_trace`: the portion of the stack trace that calls a controller. The debugger prints messages like this:
```javascript
{
"controller:predicates:is_request_sources_only_whosonfirst": {
    "reply": true,
    "stack_trace": "at controller (/Users/lilyhe/api/controller/libpostal.js:10:10)"
    }
},
```

The debug timer structure has changed to accommodate multiple timers. Users can save off the value from `beginTimer()` and pass the variable into `stopTimer()`. Different timers can be referenced by different `initialTime`.

Addresses @missinglink 's comments https://github.com/pelias/api/pull/944#discussion_r131636817

Connects to pelias/pelias#207. Follows up #944

### DIFF
--- a/controller/place.js
+++ b/controller/place.js
@@ -5,6 +5,8 @@ const retry = require('retry');
 
 const mgetService = require('../service/mget');
 const logger = require('pelias-logger').get('api');
+const Debug = require('../helper/debug');
+const debugLog = new Debug('controller:place');
 
 function requestHasErrors(request) {
   return _.get(request, 'errors', []).length > 0;
@@ -20,7 +22,7 @@ function setup( apiConfig, esclient ){
     if (requestHasErrors(req)){
       return next();
     }
-
+    const initialTime = debugLog.beginTimer(req);
     // options for retry
     // maxRetries is from the API config with default of 3
     // factor of 1 means that each retry attempt will esclient requestTimeout
@@ -42,6 +44,7 @@ function setup( apiConfig, esclient ){
     });
 
     logger.debug( '[ES req]', cmd );
+    debugLog.push(req, {ES_req: cmd});
 
     operation.attempt((currentAttempt) => {
       mgetService( esclient, cmd, function( err, docs ) {
@@ -50,6 +53,7 @@ function setup( apiConfig, esclient ){
         // only consider for status 408 (request timeout)
         if (isRequestTimeout(err) && operation.retry(err)) {
           logger.info(`request timed out on attempt ${currentAttempt}, retrying`);
+          debugLog.stopTimer(req, initialTime, 'request timed out, retrying');
           return;
         }
 
@@ -81,6 +85,7 @@ function setup( apiConfig, esclient ){
 
         next();
       });
+      debugLog.stopTimer(req, initialTime);
     });
 
   }
--- a/controller/place.js
+++ b/controller/place.js
@@ -22,7 +22,7 @@ function setup( apiConfig, esclient ){
     if (requestHasErrors(req)){
       return next();
     }
-    const initialTime = debugLog.beginTimer(req);
+
     // options for retry
     // maxRetries is from the API config with default of 3
     // factor of 1 means that each retry attempt will esclient requestTimeout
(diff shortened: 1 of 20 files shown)


tigerlily-he:
Needs to be rebased to include pelias#949 that removes the `is_service_enabled` predicate. 

The `has_parsed_text_properties` predicate needs debug log statements.

dianashk:
Hi @tigerlily-he! We miss you. 👋

tigerlily-he:
Ready to be merged. I'll leave adding more debug statements up to the team.

controller/place.js
```diff
@@ -50,6 +53,7 @@ function setup( apiConfig, esclient ){
         // only consider for status 408 (request timeout)
         if (isRequestTimeout(err) && operation.retry(err)) {
           logger.info(`request timed out on attempt ${currentAttempt}, retrying`);
+          debugLog.stopTimer(req, initialTime, 'request timed out, retrying');
```
trescube: Wouldn't this just keep logging time deltas based on when the first request was made rather than the time that the subsequent attempts were made?

trescube: That is, if `initialTime` is set to `0`, the first failed request takes 10ms and the second successful request takes 13ms, then 2 debug log messages would be output: `took 10ms` and `took 23 ms` even though the 2nd only took 13ms.

controller/search.js
```diff
@@ -110,7 +113,7 @@ function setup( apiConfig, esclient, query, should_execute ){
         logger.debug('[ES response]', docs);
         next();
       });
-      debugLog.stopTimer(req);
+      debugLog.stopTimer(req, initialTime);
```
trescube: Same comments as with controllers/place, `initialTime` never changes and therefore reports incorrect times.

controller/search_with_ids.js
```diff
@@ -95,14 +101,19 @@ function setup( apiConfig, esclient, query, should_execute ){
...
+            }});
 
           }
 
         }
         logger.debug('[ES response]', docs);
         next();
       });
-
+      debugLog.stopTimer(req, initialTime);
```
trescube: `initialTime` is only set once, and therefore reports incorrect times.

controller/place.js
tigerlily-he: You're right, all the durations would be based off the very first attempt's start time. A fix is to call a new `beginTimer()` immediately after stopping the timer during a failed request. 

```
debugLog.stopTimer(req, initialTime, 'request timed out, retrying');
initialTime = debugLog.beginTimer(req, `Attempt ${currentAttempt +1}`);
```
Would this mean overwriting the `initialTimer` constant (changing to `let`)? The call to stop timer will not know which is the "true" start time to calculate delta.

trescube: Try changing the scope of `initialTime` so that it can be set per attempt and still use `const`.

tigerlily-he: I tried to create an array to store all the start time values. The stopTimer will take the last element of the array, which will always be the most recent query attempt start time. 

```
const timerStarts = new Array();
timerStarts.push(debugLog.beginTimer(req));
[...]
debugLog.stopTimer(req, _.last(timerStarts), 'request timed out, retrying');
timerStarts.push(debugLog.beginTimer(req, `Attempt ${currentAttempt +1}`));
```
However the linter requires `array literal notation []`. If the array is constant, how can I push more elements to it? I chose an array over an object to preserve the order.

trescube: How about just making `initialTime` scoped to the `operation.attempt((currentAttempt) => { ... }` function on line 49?

trescube:
Thanks @tigerlily-he!

tigerlily-he:
Yay! I'll merge it. The Circle CI tests are failing because Node v0.12.2 is used.

Issue #973, opened 2017-09-05
Postalcode Search Issues


rabidllama:
When searching for postalcode and country (a postalcode should be unique to a country) results are not as expected:

1) For normal searching, providing the country results in an addressit search with multiple records being returned from outside of the country.
2) In structured searching, providing another field along with a postalcode results in the postalcode information being completely ignored 

In both cases, if there is a matching postalcode within the country provided, only that record should be returned.

dianashk:
Hey @rabidllama, we believe this is fixed in production now. Please have a look and let us know if this issue can be closed. Cheers!

PR #974, opened 2017-09-05
Added postalcode rules for node-postal call and fallbacks

rabidllama:
Updated searching to take into account postcodes in structured search and to use the fallback scoring if postcode is passed with a country.

In relation to issue #973


rabidllama:
Just added some tests (and added a couple of postcode parameters to some other tests) which hopefully will cover it :)

PR #977, opened 2017-09-05
treat undefined text-analyzer response as no house number available

### LINKED ISSUE #976: _synthesize_analysis sanitizer doesn't handle undefined response well
In the structured request:  

http://pelias.github.io/compare/#/v1/search/structured%3Faddress=2021%20E.%20Washington%20St.,%203A&country=United%20States&layers=address&locality=Colton&postalcode=92324&region=CA&size=1

Libpostal returns 2 `house_number`s from the `address` field instead of a `house_number` and `unit`:

```
> 2021 E. Washington St., 3A

Result:

{
  "house_number": "2021",
  "road": "e. washington st.",
  "house_number": "3a"
}
```

We have logic in pelias-text-analyzer to treat a libpostal response with duplicate fields as invalid and to return `undefined`.  The _synthesize_analyzer sanitizer doesn't handle `undefined` as a return value.  

At some point, the pelias-text-analyzer should be changed to handle this special case.

trescube:
Fixes #976

### DIFF
--- a/sanitizer/_synthesize_analysis.js
+++ b/sanitizer/_synthesize_analysis.js
@@ -27,12 +27,8 @@ function isPostalCodeOnly(parsed_text) {
 // so because we're treating the entire field as a street address, it's safe
 // to assume that an identified postcode is actually a house number.
 function getHouseNumberField(analyzed_address) {
-  for (var field of ['number', 'postalcode']) {
-    if (analyzed_address.hasOwnProperty(field)) {
-      return field;
-    }
-  }
-
+  // return the first field available in the libpostal response, undefined if none
+  return _.find(['number', 'postalcode'], _.partial(_.has, analyzed_address));
 }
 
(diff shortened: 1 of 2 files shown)


sanitizer/_synthesize_analysis.js
```diff
@@ -27,12 +27,8 @@ function isPostalCodeOnly(parsed_text) {
...
 // to assume that an identified postcode is actually a house number.
 function getHouseNumberField(analyzed_address) {
-  for (var field of ['number', 'postalcode']) {
-    if (analyzed_address.hasOwnProperty(field)) {
-      return field;
-    }
-  }
-
+  // return the first field available in the libpostal response, undefined if none
+  return _.find(['number', 'postalcode'], _.partial(_.has, analyzed_address));
```
orangejulius: so lodashy :stuck_out_tongue_closed_eyes:

Issue #971, opened 2017-09-01
Contradicting HTTP header responses?


sbaxter:
I noticed the [CORS headers are hard-coded](https://github.com/pelias/api/blob/01baa47a83ae1e32a7163ff5eeb600195fc2d37f/middleware/cors.js#L3-L6), thus producing the following response headers:
```
< HTTP/1.1 200 OK
< Access-Control-Allow-Credentials: true
< Access-Control-Allow-Headers: X-Requested-With,content-type
< Access-Control-Allow-Methods: GET, OPTIONS
< Access-Control-Allow-Origin: *
...
```

One of our devs filed the following report:
> You cannot specify both `Allow-Credentials=true` and use a wildcard. When Allow-Credentials is true, the Allow-Origin must echo the Origin header value from the request (generally provided it matches a whitelist). However, if the data is truly wide open (which it currently is because the curl request succeeded without sending cookies or certs), you can just remove Allow-Credentials or set it to false and keep Allow-Origin as a wildcard.

I believe the developer is referring to [this portion](https://www.w3.org/TR/cors/#resource-sharing-check-0) of the spec .

Is there any way we could configure the headers ourselves from the server config file?

trescube:
I don’t think there’s a way to configure this from pelias.json, but that’s a good idea for people looking to run their own pelias instances. We'll discuss internally next week when everyone is back in the office.

darkyelox:
For now what you can do is to edit the file in middleware/cors.js for setting the headers as you want, then restart the api process.

orangejulius:
Hi @sbaxter,
Thanks for reporting this. Looks like your developer is right, and the MDN page on CORS [agrees](https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS#Credentialed_requests_and_wildcards). I looked through the history of our code and there's nothing that points to us needing that header specifically, and I can't think of any reason we would need it. Certainly not for the pelias API itself which doesn't deal with authentication intentionally. We'll remove it.

orangejulius:
Ok, https://github.com/pelias/api/pull/982 will remove this header. @sbaxter out of curiosity do you know what the effect of having these conflicting headers was? Since we never saw it, I'd love to know what to look for in the future.

PR #982, opened 2017-09-08
Remove Access-Control-Allow-Credentials header

orangejulius:
This header is only relevant if dealing with authentication via cookies or other methods in HTTP requests. The Pelias API intentionally doesn't deal with authentication at all, assuming anyone who wants authentication will deal with it using a service placed between Pelias
and end users.

Additionally the CORS spec [does not allow](https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS#Credentialed_requests_and_wildcards) specifying this header while setting a wildcard for `Access-Control-Allow-Origin`, so it can cause problems in some cases.

Fixes https://github.com/pelias/api/issues/971


missinglink:
:+1: I don't recall where I copied those lines from originally, probably stackoverflow or some other semi-authoritative source :)

care should be taken to ensure we don't break compatibility with common front-end javascript http clients, there are a couple examples available as jsfiddles in https://mapzen.com/documentation/search/use-cors/

PR #1001, opened 2017-09-22
added empire as supported explicit and as part of coarse

trescube:
allows empire placetype records to be returned when `layers=empire` and `layers=coarse`.

### DIFF
--- a/helper/type_mapping.js
+++ b/helper/type_mapping.js
@@ -49,7 +49,7 @@ var LAYERS_BY_SOURCE = {
  openaddresses: [ 'address' ],
  geonames: [ 'country','macroregion', 'region', 'county','localadmin',
   'locality','borough', 'neighbourhood', 'venue' ],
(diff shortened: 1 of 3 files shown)


orangejulius:
Do we actually want to support `layers=empire`, or only being able to return empire in the hierarchy? Anyways it looks like there _might_ be a few more places where we'd have to add `empire` such as

https://github.com/pelias/api/blob/master/controller/coarse_reverse.js#L18
https://github.com/pelias/api/blob/master/test/unit/controller/predicates/is_coarse_reverse.js#L7

trescube:
We probably should in /reverse requests since we support all the other placetypes.  

I'm working on the coarse_reverse controller now, we don't even need to filter on placetype since the PiP service will take care of that.  I'll add the predicate.

trescube:
Closed in favor of #1002

PR #1013, opened 2017-10-02
wrap interpolation service call to intercept errors

trescube:
This PR introduces a function that wraps the microservice-wrapper that the interpolation middleware layer uses to intercept errors so that the entire request doesn't fail if any interpolation attempt produces an error.

### DIFF
--- a/test/unit/middleware/interpolate.js
+++ b/test/unit/middleware/interpolate.js
@@ -38,129 +38,6 @@ module.exports.tests.early_exit_conditions = (test, common) => {
 
 };
 
-module.exports.tests.error_conditions = (test, common) => {
-  test('service error string should log and not modify any results', t => {
-    t.plan(2);
-
-    const service = (req, res, callback) => {
-      callback('this is an error', {
-        properties: {
-          number: 17,
-          source: 'OSM',
-          source_id: 'openstreetmap source id',
-          lat: 12.121212,
-          lon: 21.212121
-        }
-      });
-
-    };
-
-    const logger = require('pelias-mock-logger')();
-
-    const controller = proxyquire('../../../middleware/interpolate', {
-      'pelias-logger': logger
-    })(service, () => true);
-
-    const req = { a: 1 };
-    const res = {
-      data: [
-        {
-          id: 1,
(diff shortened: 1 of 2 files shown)


test/unit/middleware/interpolate.js
```diff
@@ -219,7 +96,7 @@ module.exports.tests.success_conditions = (test, common) => {
 
   });
 
-  test('interpolated results should be mapped in', t => {
+  test('only \'street\' layer results should attempt interpolation', t => {
```
orangejulius: nice clarification!

test/unit/middleware/interpolate.js
```diff
@@ -398,6 +276,171 @@ module.exports.tests.success_conditions = (test, common) => {
...
+            lat: 12.121212,
+            lon: 21.212121
+          }
+        });
+
+      } else if (res.id === 2) {
+        callback('id 2 produced an error string', {});
+
+      } else if (res.id === 3) {
+        callback({ message: 'id 3 produced an error object' }, {});
```
orangejulius: nice! testing different possible error cases is good :)

dianashk:
@orangejulius, good question about how to acceptance-test this. one way would be use a very common street such as `Main Street` and verify that at least a couple of interpolated results are returned. There's guaranteed to be one that isn't matched which would trip up the old code.

orangejulius:
@dianashk  Well that's the thing. Previously, the interpolation server would return a 400 error in that case. But now it returns 200. So there are no errors. there aren't always results, but there also aren't errors. As far as I know the only way to get an error response back from the interpolation service now is to either send it invalid parameters or to have it broken somehow...

trescube:
It wouldn't be as good as an acceptance test but one way to smoke test it would be to lower the timeout to 250ms.  It's been shown in the past that we've had to go with a value higher than the 250ms default.  Having a lower timeout would show that there are some failures (but hopefully not all), those this is not reproducible.

trescube:
Setting the timeout to 75ms allows an occasional timeout error to slip thru and uninterpolated addresses are still returned.

Issue #888, opened 2017-05-12
Consider replacing `extend` module with lodash extend


orangejulius:
In a few places in the code, we use the [extend](https://www.npmjs.com/package/extend) module. However, lodash also [supports extend](http://stackoverflow.com/questions/19965844/lodash-difference-between-extend-assign-and-merge#19966511). We can get rid of one NPM dependency if the one from lodash works in the same way!

PR #1015, opened 2017-10-03
Remove extend dependency

trescube:
Removes the [extend](https://www.npmjs.com/package/extend) dependency, opting for [Lodash extend](https://lodash.com/docs/4.17.4#assignIn).


orangejulius:
Fixes https://github.com/pelias/api/issues/888!!! :tada:  amazing!

PR #1021, opened 2017-10-10
switched structured query tests away from fixtures

trescube:


trescube:
Thanks!  I'll keep plugging away at the others.

PR #1023, opened 2017-10-11
WIP: Query for venues on admin only

trescube:


controller/predicates/is_layer_requested.js
```diff
@@ -0,0 +1,19 @@
+const _ = require('lodash');
```
dianashk: would it make sense for this predicate to look like `has_results_at_layer`? with functions for `any` and `all`? i'd imagine we would eventually want to use it that way and I kinda like the consistency across these where possible. thoughts?

query/venues.js
```diff
@@ -0,0 +1,117 @@
...
+    return _.has(clean.parsed_text, layer);
+  });
+
+  if (mostGranularLayer) {
+    vs.var('input:query', clean.parsed_text[mostGranularLayer]);
+  }
+
+
+  // focus point
+  if( check.number(clean['focus.point.lat']) &&
```
dianashk: Probably outside the scope of this PR, however, each time we make a copy of existing code I have an internal rule: _ if this is the 3rd copy, it's time to move it to a utility function_.
We have the `focus.point` and `boundary.rect` handling in each of our query files: search, autocomplete, reverse, structured, search_original, and now this one. Would be great if we handled them similar to the existing `query\text_parser.js` style, by grouping them into a utility file.

orangejulius:
This looks good to me and seems to be working pretty well on dev. What's left before we can merge?

trescube:
Nothing left to do!  Going to attempt to merge now (or soon, depending on how bad the rebase is).

controller/search_with_appending_results.js
```diff
@@ -0,0 +1,127 @@
...
+        // (handles bookkeeping of maxRetries)
+        // only consider for status 408 (request timeout)
+        if (isRequestTimeout(err) && operation.retry(err)) {
+          logger.info(`request timed out on attempt ${currentAttempt}, retrying`);
+          debugLog.stopTimer(req, initialTime, 'request timed out, retrying');
+          return;
+        }
+
+        // if execution has gotten this far then one of three things happened:
+        // - the request didn't time out
```
orangejulius: Should this read 'the request timed out'?

trescube: It's correct as-is, the code block above this comment handles timeouts, so the code after handles "no timeout" and "there were too many timeouts so we've given up trying".

orangejulius: Ah ok. so maybe the comment in that section on line 64 should read something like `return immediately if no further action should be taken because of a timeout or the retry limit being hit`, and then this comment can say something like `handl all errors (besides timeouts and retry limit hits which are both handled above)`. Does that sound right?

orangejulius: The wording can obviously change, but what I'm getting at is that the comment reads like any one of the 3 items in the list is handled by the code below the comment, which is not true

trescube: I'm trying to update the comments now but fighting with npm-check.

orangejulius: god speed

trescube: Updated!  Let me know how that looks.

PR #1031, opened 2017-10-16
use host from HOST environmental variable

### LINKED ISSUE #771: Enable configuration of listen address
If running behind a reverse proxy / load balancer on a separate machine, there is no need to only allow listening to localhost, thus it would make sense if this could be configurable.

trescube:
Fixes #771


orangejulius:
Nice

PR #1050, opened 2017-10-28
Verify min and max longitude and latitude.

pbackx:
If min and max are switched this leads to a fairly strange error message. See pelias (not api!) issue https://github.com/pelias/pelias/issues/656

I accidentally got here because I was searching for that exact Elasticsearch error message, so I thought, why not fix this and get to know the Pelias project a bit in the mean time :smile:


orangejulius:
Ok merged! Sorry about the noise, I made a minor tweak to your commit so that it points to the issue in the pelias/pelias repo rather than the unrelated one with the same number in this repo, and then merged it in a way where this PR shows up as closed instead of merged. It should all be sorted out now :)

PR #1056, opened 2017-11-09
Setting env variable host in dockerfile to make app listen on every available network interface

sweco-semhul:
Since this change, https://github.com/pelias/api/commit/cb8f4017e55c2b7c6647d23e655e0e59d7af8da5#diff-168726dbe96b3ce427e7fedce31bb0bc I hade a hard time running Pelias API in docker setup. Could not reach the api from outside docker anymore. This small change will fix that.

Took me some time to find this problem so hopefully getting this fixed could help others.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,9 @@ LABEL maintainer="pelias@mapzen.com"
 
 EXPOSE 3100
 
+# Make sure host is set to enable exposure of app outside of docker container
+ENV HOST 0.0.0.0
+
 # Where the app is built and run inside the docker fs
 ENV WORK=/opt/pelias
 


orangejulius:
Hey @sweco-semhul,
Thanks for taking the time to track this down. We discovered it also, and it wasn't an intended change. We should have fixed it in https://github.com/pelias/api/pull/1043, can you verify that with an up-to date master this PR is no longer needed? If we broke something else we definitely want to know :)

sweco-semhul:
Oh my bad. should have searched the issues before requesting this.

Yes updating to master solves this it will be set to undefined instead and that seems to work.

PR #1058, opened 2017-11-11
Pelias intersections

AlexLoyko:
Dear Pelias community,

At this point, I have tried my best and think that my pelias-intersections implementation is ready to take a look at. I have tried to take care of all the concerns and suggestions that came up when I visited your office, but if i missed something just let me know.

It's been a pleasure.


controller/search.js
```diff
@@ -1,6 +1,7 @@
 'use strict';
 
 const _ = require('lodash');
+const util = require('util');
```
orangejulius: Don't forget to clean up require statements that are no longer used :)

controller/search.js
```diff
@@ -21,6 +22,7 @@ function setup( apiConfig, esclient, query, should_execute ){
     const debugLog = new Debug('controller:search');
 
     let cleanOutput = _.cloneDeep(req.clean);
+
```
orangejulius: Same for whitespace. Ideally there'd be no extra whitespace changes not part of the actual code changes. Not a very big deal of course :P

helper/intersectionsParsing.js
```diff
@@ -0,0 +1,134 @@
...
+        str2 = wordsToSentence(words, delimiterIndex+1, words.length);
+    } else {
+        throw 'Missing streets in the intersection';
+    }
+    return { street1: str1, street2: str2 };
+};
+
+// intended for intersections only
+// this function turns '77' into '77th', '3' into '3rd', etc
+function addOrdinality(arr) {
```
orangejulius: This function won't handle 11th, 12th, or 13th. There's a pretty elegant implementation at https://github.com/dcousens/ordinal/blob/master/indicator.js. Its up to you if you want to improve this code, use that implementation, or add a requirement on the ordinal package from NPM. Any works for us.

However if you do decide to modify this implementation I would suggest looking at [Array.map()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) which is useful when you are iterating over an array and modifying the values within that array.

helper/intersectionsParsing.js
```diff
@@ -0,0 +1,134 @@
...
+        w : 'West',
+        s : 'South',
+        n : 'North',
+        ne: 'Northeast',
+        nw: 'Northwest',
+        se: 'Southeast',
+        sw: 'Southwest'
+    };
+
+    for (let i = 0; i < arr.length; i++) {
```
orangejulius: This is another case where `Array.map` would really help. instead of modifying the incoming array, returning a new array with appropriate changes made is generally easier to read and make correct.

query/search_intersections.js
```diff
@@ -0,0 +1,227 @@
...
+                              { 'match': { 'layer': 'intersection'} },
+                              { 'match': { 'address_parts.street1': clean.parsed_text.street2} },
+                              { 'match': { 'address_parts.street2': clean.parsed_text.street1} }
+                          ]
+                      }
+                  }
+              ]
+          },
+          'sort': [sort]
+      }
```
orangejulius: We probably want to add additional filters to this query so that it matches the rest of our queries. Layers, sources, etc. @trescube will know more and also can advise if we want to handle any sort of fallback (to street, city, etc) logic here.

routes/v1.js
```diff
@@ -127,7 +131,20 @@ function addRoutes(app, peliasConfig) {
...
+    not(hasRequestErrors)
+  );
+
+  const intersectionQueryShouldExecute = all (
+    isIntersectionLayer,
+    not(hasRequestErrors)
+  );
+
   const libpostalShouldExecute = all(
+    //not(isIntersectionLayer),
```
orangejulius: If this and the following few commented out lines aren't needed please remove them.

test/unit/controller/predicates/is_intersection_layer.js
```diff
@@ -0,0 +1,77 @@
+'use strict';
```
orangejulius: This file is full of good tests and is pretty comprehensive. Very nice :)

test/unit/runIntersectionsTests.js
```diff
@@ -0,0 +1,18 @@
...
+var common = {
+    // a visual deep diff rendered using console.error()
+    diff: function( actual, expected ){
+        console.error( diff.compare( actual, expected ) );
+    }
+};
+
+var tests = [
+    require('./helper/intersectionsParsing'),
+    require('./controller/predicates/is_intersection_layer')
```
orangejulius: There's no reason for these tests to be separate from our main tests (I'm sure it was easier initially when testing, so no worries). Please move these two lines into `test/unit/run.js` and remove this file.

util/arrayHelper.js
```diff
@@ -0,0 +1,75 @@
...
+    for(let i = 0; i < arr.length; i++) {
+        if(arr[i] === '') {
+            arr.splice(i, 1);
+            i--;
+        }
+    }
+    return arr;
+};
+
+// intended for intersections only
```
orangejulius: A lot of this code looks duplicated. If its only used in one place it would be fine to remove from this file and keep in the code higher in this PR.

helper/intersectionsParsing.js
```diff
@@ -0,0 +1,134 @@
...
+            return 1;
+        }
+    }
+
+    return 0;
+}
+
+function wordsToSentence(arr, start, end) {
+    var sentence = '';
+    for (let i = start; i < end; i++) {
```
orangejulius: Take a look at [Array.join()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join). It allows you to combine elements of an array with a configurable separation character, so that you don't have to write your own logic to add a space in between all words but not at the end. In fact you can probably get rid of this helper function and use `join()` perhaps in combination with `slice()`

helper/intersectionsParsing.js
```diff
@@ -0,0 +1,134 @@
...
+        // changes '6' to '6th', etc
+        words = addOrdinality(words);
+        // only treat input as intersection if contains '&' or 'and'
+        const delimiter = _.includes(text, '&') ? '&' : 'and';
+        const delimiterIndex = words.indexOf(delimiter);
+
+        str1 = wordsToSentence(words, 0, delimiterIndex);
+        str2 = wordsToSentence(words, delimiterIndex+1, words.length);
+    } else {
+        throw 'Missing streets in the intersection';
```
orangejulius: I'm not super familiar with our API middleware code anymore, but what part of the code would catch this exception? My guess is that it would be pretty common for this code to fail to find a valid intersection in the input, so throwing an exception in those cases doesn't feel right. But again, I don't know this code super well anymore, so maybe its ok.

query/search_intersections.js
```diff
@@ -0,0 +1,227 @@
+/* eslint-disable */
```
orangejulius: What doesn't eslint like in this file? Ideally we remove this line and fix the thing it complains about.

AlexLoyko:
@orangejulius Hi Julian! Thank you for your prompt feedback. I am glad it is coming along. Sorry for the sloppiness. I'll clean it up shortly and recommit. Thank you!

test/unit/controller/predicates/is_intersection_layer.js
AlexLoyko: @orangejulius Thanks!

test/unit/runIntersectionsTests.js
AlexLoyko: @orangejulius Taken care of!

controller/search.js
AlexLoyko: @orangejulius Done!

controller/search.js
AlexLoyko: @orangejulius Done!

helper/intersectionsParsing.js
AlexLoyko: @orangejulius You are right. Since I changed the approach to fallback if the intersection query returns nothing, the exception no longer makes sense.

AlexLoyko: Done!

helper/intersectionsParsing.js
AlexLoyko: @orangejulius Oops:) Implemented using the proposed solution. Very elegant.

helper/intersectionsParsing.js
AlexLoyko: @orangejulius Done!

query/search_intersections.js
AlexLoyko: @orangejulius Right. I have some suggestions, too. Would love to talk.

util/arrayHelper.js
AlexLoyko: @orangejulius That was my misstep. Pure duplicate of helper/intersectionsParsing.js. Deleted

AlexLoyko:
@orangejulius I have taken care of the comments you have made:) Waiting on your future comments.

orangejulius:
Okay this is looking great! A few bits of cleanup can really make a PR nicer :) As far as I can tell the only thing left to discuss on this PR is the query. @trescube we will definitely need you to chime in here.

Beyond that, lets talk about what else is required for intersections? How is the importer coming along. Are there changes required to pelias/model or pelias/schema?

Also, @AlexLoyko would you mind if I rebased your commits to clean up the git history and pushed to your branch?

AlexLoyko:
@orangejulius Great. When should I come in?

The importer is pretty much the one you helped with a a few modifications. You can find it here https://github.com/PeliasIntersections/pelias-intersections. Also, the modifications needed to model and schema are here: https://github.com/PeliasIntersections/schema and here https://github.com/PeliasIntersections/model.

Regarding the rebasing, no objections. Do whatever you think is better.

Thanks! Let me know the possible times to come in.

jperelli:
Hi @AlexLoyko @orangejulius! awesome work! Are there any follow-ups on this one? I want to help on this:
 1. should I make another PR with the conflicts solved? Or can it be solved here?
 2. What are the next steps in order to merge this? / Who needs to approve+merge it?

orangejulius:
We ended up implementing intersection support in https://github.com/pelias/api/pull/1287. Thanks for the initial prototype! :)

PR #1060, opened 2017-11-14
Microserviceify libpostal

### LINKED ISSUE #1030: microserviceify libpostal
Currently the API communicates with libpostal locally which presents 3 issues:

- the API takes extra time to start
- the API cannot be run in clustered mode as each would require its own libpostal expensive instance
- treats libpostal as a special case whereas PiP, interpolation, and placeholder are configured and called as microservices

The ideal scenario would be to move libpostal to an HTTP interface elsewhere and treat it as a microservice call.

trescube:
Fixed #1030 

This PR:

- removes the pelias-text-analyzer dependency
- moves conversion logic from pelias-text-analyzer to controller/libpostal.js
- adds controller/structured_libpostal.js that calls libpostal for help splitting up `address` into `number` and `street` (moved from sanitizers/_synthesize_analysis.js)
- adds libpostal service to schema
- adds service/configurations/Libpostal.js that accepts a function that extracts a property from `req` (required to satisfy calling libpostal with values from different places)
- fixed some typos in test/unit/schema.js


orangejulius:
I just tested this out locally and it works great. One thing I noticed is there are two log entries on startup for the libpostal service. What are the implications of that?
image

trescube:
That's an unfortunate side effect of having 2 libpostal configurations initialized:  1 for standard libpostal that uses `req.clean.text` and the other for structured which uses `req.clean.parsed_text.address`.

trescube:
A service instance needs a configuration and there are two configurations.  You might be able to make it "smarter" in `getParameters` but that's burying business logic deep inside which would get confusing.

orangejulius:
It sounds like we don't have the right abstraction for the interface to libpostal then. Can the parameter to the libpostal service take exactly one parameter, which contains the actual value to parse, and then wherever is calling it contains the knowledge of which property on the req object to use?

trescube:
That might work for now but there are advantages to passing in the entire `req` object.  For instance if other things are needed later.  This is why the `propertyExtractor` is passed to the constructor.

orangejulius:
Personally having used the microservice-wrapper module heavily for the first time the other day I found that it was a really great interface _except_ for the fact that its currently somewhat assumed that the entire `req` and `res` objects will be passed in. I think that's far too much context for that module to be given, even if it only uses a little bit of it. Even without the side effects in this PR I  would love to see us change that.

If and when we need to send more context to microservice-wrapper, we can add params for the specific values we need to send in.

orangejulius:
It also doesn't have to be now. I'm actually totally ok merging this as is :)

PR #1068, opened 2017-11-29
Output "category" field in all responses

amatissart:
`category` field is currently output only when a valid `categories` list is passed in request params.

This PR suggests to always include this field if it's defined.
Is there any drawback to returning categories in all responses ?

### DIFF
--- a/helper/geojsonify_place_details.js
+++ b/helper/geojsonify_place_details.js
@@ -57,7 +57,7 @@ const DETAILS_PROPS = [
   { name: 'marinearea_a',      type: 'string' },
   { name: 'bounding_box',      type: 'default' },
   { name: 'label',             type: 'string' },
-  { name: 'category',          type: 'array',     condition: checkCategoryParam }
+  { name: 'category',          type: 'array' }
 ];
 
 // returns true IFF source a country_gid property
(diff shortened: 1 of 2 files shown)


dianashk:
Hi @amatissart! 👋 Happy holidays! 🎉 🎉 
Thank you for taking an interest in the Pelias project. We're always excited to see contributions.
The category functionality is in `beta` at this time because the category names (taxonomy) has not been finalized yet. We are purposely not exposing the categories in the general user experience to avoid anyone relying on the current taxonomy. Once it changes it could cause potentially breaking changes for anyone that hardcodes the names.

With that said, if exposing the categories was behind a config flag that gets processed at startup time, we'd be happy to merge it. The way it stands in this PR would mean that categories are always exposed in production and we just aren't ready to go that route.

Are you hoping to get this functionality into Mapzen Search's hosted instance or are you hosting your own instance of Pelias and need this to be turned on there?

Thanks again for your time.

amatissart:
Thanks for the clarification.
Yes we'd like to host our own instance, so using a config flag would be helpful indeed.
Would you have a hint on how to achieve this, so that I can update this PR ?

orangejulius:
This was done in https://github.com/pelias/api/pull/1273!

PR #1070, opened 2017-12-04
dedup OSM/OA address

antoine-de:
linked to https://github.com/pelias/pelias/issues/541
and maybe to https://github.com/pelias/pelias/issues/541 but I cannot reproduce it.

and to the fact that there are lots of dupplicates in france like:
[20 rue hector malot paris](https://mapzen.com/search/explorer/?query=searchtext=20%20rue%20hector%20malot%20paris) that returns 1 result from open address and 1 result from OSM

now we first check which source has a zipcode and if both have, we prefer OA over OSM


orangejulius:
Hey @antoine-de thanks for the quick PR! This looks great to me. Thanks for adding a test! I can't remember if there was a reason why we chose only to deduplicate OA addresses, and not OSM. My thought is it was an oversight and we would gladly dedupe either source. But I'd like to know if @dianashk or @missinglink or @trescube remember a reason why the code did that.

Assuming there isn't a reason this looks good to merge!

orangejulius:
I looked into this and the original issue behind the code in question is https://github.com/pelias/api/issues/872. The resulting code in https://github.com/pelias/api/pull/890 was written to only affect OpenAddresses records, but only because the issue in that case was for deduping between two OpenAddresses records. However I don't think there was any intention or need to limit to only OpenAddresses, and not OpenStreetMap.

orangejulius:
Hey @antoine-de,
I finally got around to testing this, and it works great. I rebased and merged it over in #1093. Thanks for the PR!

PR #1082, opened 2018-01-27
Add MIT license

Echelon9:
**pelias-api** never previously had a `LICENSE`, although it appears clear from contextual information that the intention was for pelias-api to be MIT licensed (e.g. metadata in `package.json` to this effect).

Copy over MIT license as formatted in the [pelias/pelias](https://github.com/pelias/pelias/blob/master/LICENSE) meta repository.


orangejulius:
Nice catch

PR #1095, opened 2018-03-10
fix: Avoid querying PIP service for non-coarse reverse

orangejulius:
Queries that specified only non-coarse layers (address or venue) and had no results returned from Elasticsearch would trigger a request to the PIP service.

The PIP service does not contain any addresses or venues so this query will never return anything, and only waste time.

@trescube @missinglink any idea how to test this?


orangejulius:
Huh that's weird, because the acceptance tests definitely show a difference

orangejulius:
I moved the language service change out of this PR and into #1098, since the address reverse calling the PIP service fix is definitely good.

PR #1097, opened 2018-03-12
Reduce minimum result size of reverse Elasticsearch query

orangejulius:
Currently, we always tell Elasticsearch to send at least 20 results back, so that a query with no `size` param, which defaults to 10, will have enough records to pretty much guarantee that even after some of them are deduped, there will still be at least 10 unique records to send to the client.

One downside to this is that parsing Elasticsearch responses is quite expensive, since it's a fairly large JSON response. The situation is most severe when a query parameter such as `size=1` is specified.

This PR adds the ability to modify the minimum result size on a per-endpoint basis, and reduces the minimum query size for `/v1/reverse` to 2.

Nothing will change for `size=10` queries, but for `size=1` queries, only 2 results will be returned from Elasticsearch, which should drastically reduce the amount of time spent parsing JSON.


orangejulius:
Tests added!

PR #1098, opened 2018-03-13
Only call language service when language not defaulted

orangejulius:
The default language is ~English~, which is also what is stored in Elasticsearch. Thus, if the user did not specify a language via headers or query parameter, there is no reason to call Placeholder.

Some Placeholder responses for language translation can be 30KB, and all that JSON takes considerable time to parse

This was originally part of #1095 

Edit: the default language is not English in WOF/Placeholder.

### DIFF
--- a/routes/v1.js
+++ b/routes/v1.js
@@ -226,6 +226,10 @@ function addRoutes(app, peliasConfig) {
     isAddressItParse
   );
 
+  const languageWasSpecified = function(req, res) {
+    return !req.clean.lang.defaulted;
+  };
+
   // get language adjustments if:
   // - there's a response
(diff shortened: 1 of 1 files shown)


orangejulius:
In retrospect, we may not want to merge this PR. Instead, we can disable _storing_, but not indexing, of the admin name fields in Elasticsearch. This should drastically cut down on the amount of data stored in Elasticsearch (all records in the USA store a duplicate copy of the string "United States", for example).

For simple get requests Placeholder is quite fast, and if needed we can cache commonly used WOF records in the API (although that brings complications), so the tradeoff of an extra request to a fast service to possibly improve the efficiency of Elasticsearch by quite a bit should be worth it.

PR #1102, opened 2018-03-22
support aliases for name fields

missinglink:
this is the twin PR for https://github.com/pelias/model/pull/90, check that out for some background on the feature.

In the PR linked above, I have enabled the ability for `name.*` fields to be stored as either a scalar string or as an array of strings.

This PR adds support for name aliases in the `pelias/api` codebase, adding a bunch of checks to ensure that middleware/helper comparisons are done on the types they are expecting (ie. strings).

In order to avoid code duplication, I split off the `getStringValue()` and `getArrayValue()` functions in to their own helper file.

In doing so, and adding tests, I wasn't satisfied with how that function was behaving, I made a change to how it functions:

I added support for integers. the original code rejected integers because they fail `_.isEmpty` cc/ @trescube was this intended? [I had to change this test](https://github.com/pelias/api/compare/name_aliases?expand=1#diff-4693562c4d2429d296796a015f8a8d1fR111), I couldn't figure out why we would want to replace numbers with empty strings, do you recall?

I'm also not a big fan of `t.equal(field.getStringValue({foo: 'bar'}), '[object Object]');` but I didn't want to change too much, and that seems to be working?

The rest of the PR is just adding test cases for various functions which would be affected by the change.

### DIFF
--- a/helper/fieldValue.js
+++ b/helper/fieldValue.js
@@ -17,7 +17,7 @@ function getStringValue(property) {
     return property;
   }
 
-  // array value, take first item in array (at this time only used for admin values)
+  // array value, take first item in array (at this time only used for admin & name values)
   if (_.isArray(property)) {
     return property[0];
   }
--- a/helper/fieldValue.js
+++ b/helper/fieldValue.js
@@ -1,0 +1,47 @@
+'use strict';
+
+const _ = require('lodash');
+
+function getStringValue(property) {
+  // numeric value, cast to string
+  if (_.isNumber(property)) {
+    return _.toString(property);
+  }
+
+  // isEmpty check works for all types of values: strings, arrays, objects
+  if (_.isEmpty(property)) {
+    return '';
+  }
+
+  if (_.isString(property)) {
+    return property;
+  }
+
+  // array value, take first item in array (at this time only used for admin values)
+  if (_.isArray(property)) {
+    return property[0];
+  }
+
+  return _.toString(property);
+}
+
+function getArrayValue(property) {
+  // numeric value, cast to array
+  if (_.isNumber(property)) {
+    return [property];
+  }
+
+  // isEmpty check works for all types of values: strings, arrays, objects
+  if (_.isEmpty(property)) {
+    return [];
+  }
+
+  if (_.isArray(property)) {
+    return property;
+  }
+
+  return [property];
+}
+
+module.exports.getStringValue = getStringValue;
+module.exports.getArrayValue = getArrayValue;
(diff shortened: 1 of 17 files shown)


test/unit/middleware/dedupe.js
```diff
@@ -223,41 +223,71 @@ module.exports.tests.trump = function(test, common) {
...
-        'address_parts': {}
-      },
-      {
-        'name': { 'default': '100 Main St' },
-        'source': 'openstreetmap',
-        'source_id': '654321',
-        'layer': 'address',
-        'address_parts': {
-          'zip': '54321'
+  test('osm with zip trumps openaddresses without zip', function (t) {
```
missinglink: this function wasn't modified, I just fixed the indentation

test/unit/middleware/dedupe.js
```diff
@@ -223,41 +223,71 @@ module.exports.tests.trump = function(test, common) {
...
+
+    var expectedCount = 1;
+    dedupe(req, res, function () {
+      t.equal(res.data.length, expectedCount, 'results have fewer items than before');
+      t.deepEqual(res.data[0].source_id, '654321', 'openstreetmap result with zip won');
+      t.end();
+    });
+  });
+
+  test('works with name aliases', function (t) {
```
missinglink: this is a new function

helper/fieldValue.js
```diff
@@ -0,0 +1,47 @@
...
+  // isEmpty check works for all types of values: strings, arrays, objects
+  if (_.isEmpty(property)) {
+    return '';
+  }
+
+  if (_.isString(property)) {
+    return property;
+  }
+
+  // array value, take first item in array (at this time only used for admin values)
```
missinglink: out of date comment

test/unit/helper/fieldValue.js
```diff
@@ -0,0 +1,41 @@
...
+module.exports.tests.getStringValue = function(test) {
+  test('getStringValue', function(t) {
+    t.equal(field.getStringValue([]), '', 'empty array');
+    t.equal(field.getStringValue(''), '', 'empty string');
+    t.equal(field.getStringValue('foo'), 'foo', 'string');
+    t.equal(field.getStringValue(['foo','bar']), 'foo', 'array');
+    t.equal(field.getStringValue(['']), '', 'array with empty string');
+    t.equal(field.getStringValue(-0), '-0', 'number');
+    t.equal(field.getStringValue(+0), '0', 'number');
+    t.equal(field.getStringValue({foo: 'bar'}), '[object Object]', '_.toString');
```
orangejulius: add a comment that this is not necessarily desired behavior, just how it used to work

PR #1107, opened 2018-03-24
Fix extremely verbose, repeated error output

orangejulius:
In some error cases, a warning is repeated many many times. It turns out there is code checking against elasticsearch error codes, and warning _each_ time it fails to match aganst one of them.

Many times, the error that is being compared is not from elasticsearch, and in any case, the [elasticsearch-exceptions](https://www.npmjs.com/package/elasticsearch-exceptions) module is 3 years old. We should rewrite most of this code and stop using that module.

For now, this at least reduces log noise.


missinglink:
@orangejulius GTG but requires a rebase

PR #1109, opened 2018-03-25
fix(package): update elasticsearch to version 14.2.1

orangejulius:
Closes #1062


orangejulius:
I tested this on geocode.earth staging real quick. Seems to work fine

PR #1128, opened 2018-04-30
[DO NOT MERGE] Disable dfs_query_then_fetch

orangejulius:
This query type dramatically slows down queries by requiring extra round-trip and computation on each Elasticsearch shard. It was enabled a [long time ago](https://github.com/pelias/api/pull/67/commits/ff9e9973fcdbd40b5b1cbec7d54be4e4d8422300) and is generally not recommended to be run on large indices.

From some cursory testing, it appears that one of our structured endpoint acceptance-tests returns different results with this setting changed, but the results are not obviously wrong. Structured queries also currently use a fairly complex query that was written before we had the [Placeholder](https://github.com/pelias/placeholder) service to make querying for records contained within a given administrative area much easier.

### DIFF
--- a/controller/search.js
+++ b/controller/search.js
@@ -49,7 +49,6 @@ function setup( apiConfig, esclient, query, should_execute ){
     // elasticsearch command
     const cmd = {
       index: apiConfig.indexName,
-      searchType: 'dfs_query_then_fetch',
       body: renderedQuery.body
     };
 
--- a/test/unit/controller/search.js
+++ b/test/unit/controller/search.js
@@ -34,7 +34,6 @@ module.exports.tests.success = function(test, common) {
         t.equal(esclient, 'this is the esclient');
         t.deepEqual(cmd, {
           index: 'indexName value',
-          searchType: 'dfs_query_then_fetch',
           body: 'this is the query body'
         });
 
@@ -96,7 +95,6 @@ module.exports.tests.success = function(test, common) {
         t.equal(esclient, 'this is the esclient');
         t.deepEqual(cmd, {
           index: 'indexName value',
-          searchType: 'dfs_query_then_fetch',
           body: 'this is the query body'
         });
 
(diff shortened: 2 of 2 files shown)


missinglink:
My understanding is that, if this setting is disabled then the sorting if different depending on which server received the request.

This means that the tests start to intermittently fail and quality (consistency) suffers.

I would be very careful disabling this.

Although I would love to disable/reduce the effects of TF/IDF, which would make this not such a big issue.

orangejulius:
I don't believe it will depend on which server recieved the request, but it _will_ vary more depending on how records ended up distributed across shards. More shards and fewer records make things worse. But at least across builds results will be the same.

I don't think we can disable this until we reduce the number of shards we use/recommend for full planet builds. 24 is too high, maybe we can go down to 12 or even 6. Likewise, some of our queries can probably be rewritten to avoid some issues.

In the long term though, we can definitely disable this.

missinglink:
Interesting, I forget exactly how it works :)

I thought that, with this setting turns off, that the request would be satisfied by all shards but the TF/IDF scoring would be based off the values local to the coordinating node.

If that's how it works (I would need to confirm this is correct) then it would in fact produce different sorting depending on which node received the request.

orangejulius:
The best documentation on this issue appears to be a [blog post from 2013](https://www.elastic.co/blog/understanding-query-then-fetch-vs-dfs-query-then-fetch), but I believe it's still correct. There is also the [search type docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-search-type.html) but it's fairly minimal.

Anyways, I believe we should still consider at least allowing this to be disabled. For full planet build it offers a nice latency improvement, and with hundreds of millions of documents it should cause minimal result impact.

For smaller builds, using a single elasticsearch shard should be messaged as a best practice. So that probably means https://github.com/pelias/config/pull/60 should be reverted. In particular, most of my comments [here](https://github.com/pelias/config/pull/60#issuecomment-307802941) are very wrong. Especially now that we have [disabled fielddata](https://github.com/pelias/schema/pull/302/files), there is even more incentive to use a single shard.

orangejulius:
I've done some testing and this PR which merely removes the parameter is not the way forward.

Full report is written up in https://github.com/pelias/api/issues/1335

PR #1131, opened 2018-05-07
add API filter targets to config

missinglink:
This PR allows developers to configure which `sources` and `layers` (and their aliases) which the API accepts, via `pelias-config`.

There is a future plan to auto-discover the `sources` and `layers` (collectively referred to as `"the targets"`) from elasticsearch. For now we are holding off on this extension until we consider how best to implement that.
An example implementation is included in the source, with a callback interface available for when we decide to do that.

Developers can edit the `api.targets` section of the `~/pelias.json` file in order to change the default values for `sources`, `targets` and their aliases.

You can find the [default values here](https://github.com/pelias/config/blob/master/config/defaults.json), most users will not need to modify these defaults, advanced users may modify the following options:

### layers_by_source

A list of all layers per source. The keys of this object will be used as the list of available `layers` and the values will be used as the available `sources`, any new layers or sources you wish to add should appear here.

```javascript
"layers_by_source": {
  "openstreetmap": [ "address", "venue", "street" ],
  "openaddresses": [ "address" ],
  "geonames": [
    "country", "macroregion", "region", "county", "localadmin", "locality", "borough", 
    "neighbourhood", "venue"
  ],
  "whosonfirst": [
    "continent", "empire", "country", "dependency", "macroregion", "region", "locality",
    "localadmin", "macrocounty", "county", "macrohood", "borough", "neighbourhood", 
    "microhood", "disputed", "venue", "postalcode", "continent", "ocean", "marinearea"
  ]
}
```

### source_aliases

A list of alternate names for sources. These 'aliases' are a convenient way to provide a short alias for a more verbose source name. An alias may refers to one or more sources.

```javascript
"source_aliases": {
  "osm": [ "openstreetmap" ],
  "oa":  [ "openaddresses" ],
  "gn":  [ "geonames" ],
  "wof": [ "whosonfirst" ]
}
```

### layer_aliases

A list of alternate names for layers. These 'aliases' are a convenient way to provide a short alias for a more verbose layer name. An alias may refers to one or more layers.

```javascript
"layer_aliases": {
  "coarse": [
    "continent", "empire", "country", "dependency", "macroregion", "region", "locality", 
    "localadmin", "macrocounty", "county", "macrohood", "borough", "neighbourhood", 
    "microhood", "disputed", "postalcode", "continent", "ocean", "marinearea"
  ]
}
```

~~config PR should be merged first: https://github.com/pelias/config/pull/83 and then the `package.json` should be updated accordingly.~~


orangejulius:
Connects https://github.com/OpenTransitTools/trimet-mod-pelias/issues/9

test/unit/helper/TypeMapping.js
```diff
@@ -0,0 +1,208 @@
...
+  test('setter setLayerAliases', function(t) {
+    var tm = new TypeMapping();
+    t.deepEqual(tm.layer_aliases, {});
+    tm.setLayerAliases({ foo: ['foo', 'bar'] });
+    t.deepEqual(tm.layer_aliases, { foo: ['foo', 'bar'] });
+    t.end();
+  });
+};
+
+module.exports.tests.generateDynamicMappings = function(test) {
```
orangejulius: This shadows an Elasticsearch term which means something pretty different, maybe we can come up with a different name?

helper/TypeMapping.js
```diff
@@ -0,0 +1,146 @@
...
+  this.layers_by_source = lbs;
+};
+
+// layer alias setter
+TypeMapping.prototype.setLayerAliases = function( aliases ){
+  this.layer_aliases = aliases;
+};
+
+// generate dynamic mappings after setters have been run
+TypeMapping.prototype.generateDynamicMappings = function(){
```
orangejulius: This shadows an Elasticsearch term which means something pretty different, maybe we can come up with a different name?

missinglink: I'm happy to change it, please suggest an alternative.

orangejulius:
We should add this documentation to the API Readme so that we can update it over time as we add/remove features. There's a section for the custom boost code now, so it makes sense this would be there as well.

Alternatively, we could put all documentation regarding customizing Pelias in pelias/documentation

PR #1150, opened 2018-06-01
fix(interpolation): Ensure proper sorting of streets with interpolated addresses

orangejulius:
In cases where several conditions are met, it is possible for results to be returned from the API that are not sorted as they were intended.

These conditions are:

* over 10 results total were returned from Elasticsearch
* the interpolation middleware was called
* not all street results end up with possible interpolated address matches, and some of those streets come before other interpolated address records, necessitating a re-sorting of the results in the interpolation middleware

In these cases, the ordering of streets as defined by Elasticsearch, such as by linguistic match or distance from a focus point, will no longer be respected in the results.

This is because Node.js's `Array.prototype.sort` uses an [*un*stable QuickSort for arrays of size 11 or greater](https://github.com/nodejs/node/blob/master/deps/v8/src/js/array.js#L670).

The solution is to switch to a sorting algorithm that is always stable.
This ensures that whatever ordering was specified in the Elasticsearch queries is observed, without any of that logic having to be duplicated, and then possibly conflict.

### DIFF
--- a/middleware/interpolate.js
+++ b/middleware/interpolate.js
@@ -2,6 +2,7 @@ const async = require('async');
 const logger = require( 'pelias-logger' ).get( 'api' );
 const source_mapping = require('../helper/type_mapping').source_mapping;
 const _ = require('lodash');
+const stable = require('stable');
 
 /**
 example response from interpolation web service:
@@ -116,7 +117,7 @@ function setup(service, should_execute) {
 
       // sort the results to ensure that addresses show up higher than street centroids
       if (_.has(res, 'data')) {
-        res.data.sort((a, b) => {
+        res.data = stable(res.data, (a, b) => {
           if (a.layer === 'address' && b.layer !== 'address') { return -1; }
           if (a.layer !== 'address' && b.layer === 'address') { return 1; }
           return 0;
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
     "pelias-sorting": "1.1.0",
     "predicates": "^2.0.0",
     "retry": "^0.10.1",
+    "stable": "^0.1.8",
     "stats-lite": "^2.0.4",
     "through2": "^2.0.3"
   },
--- a/test/unit/middleware/interpolate.js
+++ b/test/unit/middleware/interpolate.js
@@ -277,6 +277,81 @@ module.exports.tests.success_conditions = (test, common) => {
 
   });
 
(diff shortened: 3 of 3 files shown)


missinglink:
@orangejulius I feel uneasy about merging a custom-written sorting algorithm which isn't accompanied by a robust test suite.

The advantage of using a third-party library is that it's (hopefully!) covered by a bunch of tests and 'battle-tested' by the community.

An easy solution would be to use a third-party library like lodash.

If you favour your implementation because it is functionally better than the lodash implementation I would suggest:

- Move the code out of 'middleware' and into an area where the shared code lives, such as 'helper' or a 'util' directory.
- Expose the interface using `module.exports` so that it can be used outside `interpolation` and can be tested.
- Write some tests covering a range of failure scenarios, possibly copying the cases from `lodash`.

The implementation from stackoverflow also mutates the `Array.prototype`. This is generally considered a bad practice, although used carefully it can be fairly powerful, I would strongly discourage its use in the Pelias codebase.

If we were to allow extension to native objects in this fashion, it would be unintuitive for debugging purposed to do it inside functions such as `middleware/interpolate.js` when it has a global effect across all arrays in the process.

Functionally the PR is great, nice fix for a nasty unexpected problem!

orangejulius:
@missinglink I originally wanted to use Lodash as well. It turns out that Lodash _does_ have a [stable sort function](https://lodash.com/docs/4.17.10#sortBy), but you can't pass a custom comparison function, so it's not possible to use it here.

I was initially hesitant to use `Array.prototype` here as well, but my tests and my understanding of the isolation of Node.js modules suggest that it won't apply globally. Anyways, I am not married to that calling format at all, although it does fit nicely with the previous code.

As for the sort method itself, we could use something like the [stable](https://www.npmjs.com/package/stable) NPM module, which looks reasonable. This is one of those cases where for some reason I actually feel hesitant to add an external dependency for something as low level as a stable sort. Which direction are you leaning towards? External module or helper method?

missinglink:
I'm 99.99% certain that modules share the same global scope and so modifying it in one module also alters it in another.

I could be wrong but it's easy enough to test.

In Go there is a Proverb:

> A little bit of copying is better than a little dependency

For the most part I agree with that for Go, node is a different beast and seems to have pretty much evolved to go against that paradym :P

I would lean towards using a well tested external library.

orangejulius:
@missinglink this PR now uses the [stable](http://npmjs.com/stable) NPM package. It does indeed have [tests](https://github.com/Two-Screen/stable/blob/master/src/test.js).

PR #1165, opened 2018-06-20
Fix for Pelias/Pelias#670

tiranno:
**The problem:**
From https://github.com/pelias/pelias/issues/670: Performing a search query for a known admin area specifying the `sources=geonames` parameter returns Who's on First results anyways.

**The cause:**
The `placeholderGeodisambiguationShouldExecute` was evaluating to true any time that WhosOnFirst was the only source OR whenever Libpostal returns an admin area. This means that any data source other than WoF that also contains admin areas would be caught in Placeholder - and Placeeholder only does searches on the WoF dataset.

**The solution:**
Simply removing the check for `isAdminOnlyAnalysis` and using only a check to see if WoF is the only source requested means that admin areas will skip Placeholder and be handled by the default searches.

**Testing:**
See https://github.com/pelias/acceptance-tests/pull/471 for tests that can be used to validate that Geonames or WhosOnFirst is returned when it is the only source requested.


orangejulius:
This looks like it will do the trick. I'll test it out later today and merge :) Thanks for the PR

tiranno:
Maybe I'm mistaken but isn't Placeholder just a faster way of searching on WoF data? If we don't specify a source then we would want all available sources returned; thus, we should skip Placeholder, no?

orangejulius:
It's more than just faster :)  It's a significantly smarter and more capable way of searching administrative data, and should be preferred where possible. Currently, it only supports WOF data, but WOF also includes most data from Geonames and many other administrative data sources now.

Because of the quality and breadth of WOF data, and the improvements inherent to Placeholder, we're okay with the fact that using Placeholder wherever possible (except when someone explicitly specifies `sources=geonames`) basically limits admin results to records from WOF.

Long term, we'll also make Placeholder support data from more administrative areas.

In the medium term, WOF also stores concordances to other data sets, so for example the WOF record [Nashville, TN in WOF](https://spelunker.whosonfirst.org/id/101723183/)  has a concordance to Geonames record [4644585](http://geonames.org/4644585/). So there's a lot we could do to search with the consistent and well defined structure of WOF records, but also help people correlate those results to whatever other dataset they like.

orangejulius:
Yeah I think that sounds right. Hopefully there aren't more cases. I'll try to do some searching to see if there might be other conditions.

tiranno:
Part Duex:

So I think I've touched every base. Added some predicates and their respective unit tests. Refined the acceptance tests. Testing the API locally on Portland looks great. Let me know if you find any other problems.

controller/predicates/is_request_sources_includes_whosonfirst.js
```diff
@@ -0,0 +1,17 @@
+const _ = require('lodash');
+const Debug = require('../../helper/debug');
+const debugLog = new Debug('controller:predicates:is_request_sources_includes_whosonfirst');
+const stackTraceLine = require('../../helper/stackTraceLine');
+
+// returns true IFF 'whosonfirst' is included in the requested sources
+module.exports = (req, res) => {
+  const is_request_sources_includes_whosonfirst = _.includes(
```
trescube: Pelias switched to node 6 at a minimum so `array.includes` is now a function

controller/predicates/is_request_sources_undefined.js
```diff
@@ -0,0 +1,17 @@
+const _ = require('lodash');
+const Debug = require('../../helper/debug');
+const debugLog = new Debug('controller:predicates:is_request_sources_undefined');
+const stackTraceLine = require('../../helper/stackTraceLine');
+
+// returns true IFF there are no requested sources
+module.exports = (req, res) => {
+  const is_request_sources_undefined = _.isEqual(
```
trescube: It would read more cleanly to do `_.isEmpty(_.get(req, 'clean.sources'))`

orangejulius:
This looks good and thank you @trescube for ensuring maximum code elegance always :)

I should have time to test this out later today

orangejulius:
@tiranno I also made you a contributor in the Pelias org. You can now make branches in any of the Pelias repos now, so feel free to do that :)

PR #1168, opened 2018-06-20
Swagger

gibranparvez:
image

### Swagger 2.0 documentation for Pelias API.

Using [swagger-jsdoc](https://www.npmjs.com/package/swagger-jsdoc) and [express-swaggerize-ui](https://www.npmjs.com/package/express-swaggerize-ui), Swagger documentation is now generated and served upon `npm start`, compiling API information commented inline on `routes/v1.js`.

**URLs**:
* Viewing Swagger UI: `<pelias-api>/api-docs`
* Viewing Swagger JSON document: `<pelias-api>/api-docs.json`

### DIFF
--- a/app.js
+++ b/app.js
@@ -1,7 +1,7 @@
-const app = require('express')();
-const swaggerJSDoc = require('swagger-jsdoc');
-const swaggerUi = require('express-swaggerize-ui');    
-const peliasConfig = require( 'pelias-config' ).generate(require('./schema'));
+var app = require('express')();
+var swaggerJSDoc = require('swagger-jsdoc');
+var swaggerUi = require('express-swaggerize-ui');    
+var peliasConfig = require( 'pelias-config' ).generate(require('./schema'));
 
 if( peliasConfig.api.accessLog ){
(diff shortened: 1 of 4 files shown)


missinglink:
wow looks cool, we will need a bit of time to test and review it!

@gibranparvez would it be possible to change the links from the `venicegeo` to something within the `pelias` org, maybe https://github.com/pelias/documentation if the docs exists there?

gibranparvez:
@missinglink That sounds good. I've seen that traditionally the comment blocks are put inline, but in our case, it is indeed overly verbose and I'll move it.

Any reference to VeniceGeo in this PR is something I've overlooked when I cherry-picked changes we've already merged into our fork and I will remove them shortly- my mistake.

gibranparvez:
@missinglink Would it instead be more appropriate to simply attach the separate json file? Currently it's being generated, but if this data is going to largely be static and not recorded inline, generating it from an external file (not v1.js) would pretty much be the same thing as simply having the static swagger.json file served. Thoughts? @orangejulius

stefanocudini:
please, this PR could it be united?

I would also need the swagger file .. even if it is not complete with all the parameters of the endpoints, but I think it is very useful to everyone in many projects.

many thanks!

Issue #1169, opened 2018-06-26
input text wrapped in quotes


missinglink:
via email:

...I found that results are different for queries with quotes around it and no quotes. For instance:

- /v1/search?text="142 CASA BELLA LN, SAINT AUGUSTINE, FL 32086"
- /v1/search?text=142 CASA BELLA LN, SAINT AUGUSTINE, FL 32086

In the first case, results are parsed with AddressIt and the returned JSON contains the true exact address.

In the second (without quotes), the query is parsed with libpostal but the result falls back to the city of Saint Augustine.

Is there any reason for this change? I'd love to understand what is driving this behaviour when using the service.

I also found this to work:

- /v1/search/structured?address=142+CASA+BELLA+LN&locality=AUGUSTINE&region=fl&postalcode=32086&country=United+States

Many thanks in advance for your help.

PR #1170, opened 2018-06-26
trim whitespace and quotation marks from a range of natural languages

missinglink:
trim whitespace and quotation marks from a range of natural languages

closes: https://github.com/pelias/api/issues/1169


orangejulius:
Looks good, will do a quick acceptance test run and see if it breaks anything unexpected.

Can you add a unit test confirming it _does not_ trim quotes in the middle of the input string?

PR #1173, opened 2018-07-18
Configurable boosts for sources and layers

orangejulius:
This PR enables setting custom boosts that apply to specific sources and layers.

The changes only affect both the `autocomplete` endpoint and the `search` endpoint when using `addressit`-style queries, which are the ones most likely to return custom data. The libpostal-based `search` queries are address specific, and not likely to benefit from custom source/layer boosts.

## Functionality

Allows setting a section like the following in `pelias.json`:

```javascript
{
   "api": {
     "customBoosts":
       "layer": {
         "stop": 5
       },
       "source": {
         "transit": 2
       }
   }
}
```

This would boost any record in the `stop` layer by 5, and any record from the `transit` source by 2. This means any record that was in the `stop` layer AND the `transit` source would be boosted even more.

In my testing, boost between 5 and 20 are probably all that's required. Boosts from 0 to 1 would effectively be a negative boost, although I haven't tested that extensively.

Connects https://github.com/OpenTransitTools/trimet-mod-pelias/issues/7


missinglink:
Also worth noting that this is an advanced configuration option that should only be used by expert users experienced with Pelias.

There are some valid cases where business logic requires certain types of documents to score higher in the result set, but balancing this boost in relation to the linguistic matching score and focus point scoring is a very difficult thing to so.

batteries not included.

orangejulius:
I pushed https://github.com/pelias/api/pull/1173/commits/973d83501819fc62cd867f98002e21831084578c which uses `proxyquire` to avoid the possibility of unit tests failures if a user runs the tests with customizations in their `pelias.json`. I think we have to do one more once-over of this functionality as a whole, and then we are finally good to merge!

orangejulius:
@missinglink this PR now includes a README section with info and guidelines on boosting. Please take a look and see if it makes sense. Once the README looks good, we can merge!

Issue #1179, opened 2018-08-01
Issues with git precommit-hook, NPM


orangejulius:
For a long time we have noticed that our git [precommit-hook](https://www.npmjs.com/package/precommit-hook) setup has some issues. It is intended to ensure, among other things, that NPM packages are correct and up to date before allowing any commits. This prevents, for example, tests passing only with an old, or new, or missing NPM package.

In general, it works quite well, however sometimes it breaks when things are clearly right (for example a fresh `rm -rf node_modules; npm install` has just been run).

This seems to be related to things like optional dependencies of very deeply nested packages, in particular [tape](https://www.npmjs.com/package/tape)

Some known fixes:
- run `npm install` twice (yes, seriously)
- change the version of `tape` to 0.2.2 (see https://github.com/pelias/api/pull/1178)

This has been going on for years, across NPM versions, and is really annoying. I've taken a look at it a few times but never figured out the exact cause.

worace:
I don't really have any answers yet but I put together a simple bash script to test this out on a few different versions (relies on having NVM set up):

```
#!/usr/bin/env/bash

versions="4.9.1 6.14.3 8.11.3 9.11.2 10.8.0"

cd ~/code/pelias/api

export NVM_DIR="$HOME/.nvm"
. $NVM_DIR/nvm.sh

for v in $versions; do
  echo "** NVM Version $v **"
  nvm install $v &> /dev/null
  nvm use $v

  rm -rf node_modules
  npm install &> /dev/null

  if npm ls > /dev/null; then
    echo "** Succeeds on $v **"
  else
    echo "** Fails on $v **"
  fi
done
```

For me it seems to consistently fail on `6.14.3` alone. Would be interesting to see if that is reproducible for others. I may be able to try it on another machine as well. Here is my output:

```
** NVM Version 4.9.1 **
Now using node v4.9.1 (npm v2.15.11)
** Succeeds on 4.9.1 **
** NVM Version 6.14.3 **
Now using node v6.14.3 (npm v3.10.10)
npm ERR! missing: tape@~0.2.2, required by zlib-browserify@0.0.3
** Fails on 6.14.3 **
** NVM Version 8.11.3 **
Now using node v8.11.3 (npm v5.6.0)
** Succeeds on 8.11.3 **
** NVM Version 9.11.2 **
Now using node v9.11.2 (npm v5.6.0)
** Succeeds on 9.11.2 **
** NVM Version 10.8.0 **
[...]

PR #1181, opened 2018-08-07
Migrate to internal iso3166 definitions

worace:
@orangejulius I made a first pass at moving the ISO3166 definitions into the project based on your comment in https://github.com/pelias/api/issues/1179#issuecomment-410856756.

I copied the data definitions from [this project](https://github.com/ecrmnn/iso-3166-1/blob/master/src/iso-3166.js) and did a little re-structuring of them in a node repl to generate the reverse mapping we need for ISO 3 -> ISO 2.

However I was not totally sure about a couple things:

1) Where to put the actual data definitions. For now I stuck them in `helper/iso3166.js` along with the functions, but they could easily go into a separate file if that is tidier. I'm not sure if there are established conventions for this.
2) How many fields for each country are worth keeping around. For now I just stuck in everything I got from that other project, which looks like `{'country':'Afghanistan','alpha2':'AF','alpha3':'AFG','numeric':'004'}`. But I think technically all we really need is one `ISO2 -> ISO3` mapping and one `ISO3 -> ISO2` mapping. So we could  squeeze those constant definitions down even further if we don't anticipate using the extra fields like `country` and `numeric`.

Curious to see if anyone has thoughts on these points. Happy to adjust this as needed if you still think it is a reasonable approach.


missinglink:
Looks good to me :)

orangejulius:
This is perfect. It's true we technically don't need the numeric code or country name, but it's not worth removing them.

I look forward to (hopefully) easier `npm install`s and obviously the blazing fast performance we will now see without a bunch of O(N) lookups :)

PR #1197, opened 2018-09-19
apply parser bugfix for incorrect parsing of diagonal directionals

missinglink:
This PR adds a new function which can be used to modify responses from `libpostal` that we know are buggy.

The bugfix I included in this PR is to fix an error where `libpostal` incorrectly thinks the directional prefix is the entire street name, such as:

```
4004 SW Beaverton-Hillsdale, Portland

      "parsed_text": {
        "number": "4004",
        "street": "sw",
        "neighbourhood": "beaverton-hillsdale",
        "city": "portland"
      }
```

Note: this bug only seems to effect diagonals, not N,S,E,W.

related: https://github.com/OpenTransitTools/trimet-mod-pelias/issues/20#issuecomment-417732128

### DIFF
--- a/controller/libpostal.js
+++ b/controller/libpostal.js
@@ -70,6 +70,7 @@ function setup(libpostalService, should_execute) {
     const initialTime = debugLog.beginTimer(req);
 
     libpostalService(req, (err, response) => {
+
       if (err) {
         // push err.message or err onto req.errors
         req.errors.push( _.get(err, 'message', err) );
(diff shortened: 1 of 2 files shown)


orangejulius:
This looks good, and does indeed fix the issue. I've added an acceptance-test in https://github.com/pelias/docker/commit/1031c8d8f37e995ec9625b8741bce2f12f53fc06, and validated that this change does not affect the global acceptance-tests on a full planet build.

In general this particular case, since it's fortunately so specific, does not really apply on geocode.earth or other generic full planet Pelias builds, since they don't use the same synonyms.

orangejulius:
I fixed a typo (`effects` -> `affects`) and shortened the commit message but otherwise this is good to go!

PR #1199, opened 2018-10-02
 Parse order on the label field for Portugal

npatroni:
Added PRT to the array on var flipNumberAndStreetCountries. 

fixes https://github.com/pelias/labels/issues/29


orangejulius:
Looks good to me, thanks! Welcome to Pelias :)

PR #1203, opened 2018-10-10
add a libpostal patch which allows recasting labels

missinglink:
It always annoyed me that libpostal parses `zoo` as a borough:

```
      "parser": "libpostal",
      "parsed_text": {
        "borough": "zoo"
      }
```

this means that we return the following for global search:

```
1)	Zooviertel, Hanover, Germany
2)	Zoo, Wuppertal, Germany
```

however, for regional search, we return nothing at all.

this PR adds to the 'libpostal bugfixes', it allows us to recast labels generated by libpostal into different labels, based on their value and assigned label.

an example response from libpostal:

```
{"label":"city_district","value":"zoo"}]
```

from the code comments:

```
const RECAST_LABELS = [{ value: 'zoo', label: { to: 'house' } }];

// recast labels for certain values, currently only applied to parses which return a single label.
// the RECAST_LABELS array contains match/replace conditions which are applied in order.
// the 'value' and 'label.to' properties are mandatory, they define the value to match on and
// the replacement label to assign. you may optionally also provide 'label.from' which will restrict
// replacements to only records with BOTH a matching 'value' and a matching 'label.from'.
```

use with caution


missinglink:
with this PR applied, a regional search for Portland goes from returning nothing to:

```
/v1/search?text=zoo

 1)	Oregon Zoo, Portland, OR, USA
 2)	Oregon Zoo, Portland, OR, USA
 3)	Petting Zoo, Washington County, OR, USA
 4)	Zoo Railway Sta, Portland, OR, USA
 5)	SW Zoo Rd, Portland, OR, USA
 6)	Christmas at the Zoo, Oregon City, OR, USA
 7)	Zoo Health Club, Hillsboro, OR, USA
 8)	Washington Park Zoo Railway Post Office, Portland, OR, USA
 9)	SW Zoo Rd & SW Knights Blvd, Portland, OR, USA
10)	SW Knights Blvd & SW Zoo Rd, Portland, OR, USA
```

orangejulius:
This looks good to me, let's test it out on a larger build next week and then we can merge.

controller/libpostal.js
```diff
@@ -114,12 +114,29 @@ function setup(libpostalService, should_execute) {
...
+  // the 'value' and 'label.to' properties are mandatory, they define the value to match on and
+  // the replacement label to assign. you may optionally also provide 'label.from' which will restrict
+  // replacements to only records with BOTH a matching 'value' and a matching 'label.from'.
+  if( response.length === 1 ){
+    let first = response[0];
+    for( var ii=0; ii<RECAST_LABELS.length; ii++ ){
+      let recast = RECAST_LABELS[ii];
+      if( !recast.hasOwnProperty('label') || !recast.label.hasOwnProperty('to') ){ continue; }
+      if( recast.value !== first.value ){ continue; }
+      if( recast.label.hasOwnProperty('from') && recast.label.from !== first.label ){ continue; }
```
orangejulius: This code is pretty defensive and hard to read. Some suggestions that might help:
- avoid use of `hasOwnProperty`, especially multiple times in a row to walk down a property chain. I've always disliked `hasOwnProperty` because it's such a lengthy function name (it's a shame it's so critical in Javascript). `_.get()` would probably work great here
- it's not clear whats different regarding the variable `first` vs `response[0]`. both are used, and `first` is not declared `const` so the expectation is it would be modified, but it's not
- RECAST_LABELS is not modified, so a `forEach` loop or something that provides more contextual hints and less looping boilerplate would help clarify what's going on over the for loop with generic iterator variable name

orangejulius:
This does have a great positive impact when searching globally for Zoo:

[/v1/search?text=zoo](http://pelias.github.io/compare/#/v1/search%3Ftext=zoo)
image

I had some trouble understanding exactly how the code worked, and offered up some suggestions for improvements, but none of them are absolutely necessary.

controller/libpostal.js
missinglink: I understand your reluctance to use `hasOwnProperty`, it's ugly and verbose, but it's the only mechanism which javascript provides for key-checking.

It's not analgous or correct to use a value check such as `!!recast.label` or `_.get()` because they do fundamentally different things.

So unfortunately we are stuck with the shitty language decisions of javascipt :P

I will change it to `_.has()` which saves a few extra keystrokes.

I favour using the core language features because I feel that despite them being shitty in javascript, are well known and predicable.

I think you favour using external libs to fix some of the ugly parts of js, and I'm cool with that, in the case of underscore I think it makes sense.

missinglink: Also worth having a look at the source code for `_.has()`, it's super complex and does a lot of things, in some cases it feels like overkill for a simple key-check.

Maybe there is a simpler solution for key-checking that we are not aware of?

Issue #1206, opened 2018-10-11
Search and autocomplete queries have different scale


orangejulius:
The `scale` parameter to our `center_point` calculations used when users pass a `focus.point` parameter differs on search and autocomplete:

- for search, it's [50km](https://github.com/pelias/api/blob/master/query/search_defaults.js#L31)
- for autocomplete, it's [250km](https://github.com/pelias/api/blob/master/query/autocomplete_defaults.js#L31).

Both changes date back to 2015, and there's no immediately obvious reason why they are at those particular values, let alone different values.

Relevant Elasticsearch documentation: [The Closer, The Better](https://www.elastic.co/guide/en/elasticsearch/guide/current/decay-functions.html)

Additionally, we should consider carefully if the `linear` decay function is what we want. As it stands, no focus boosting is applied to results over 2x the `scale` parameter, so 100km for search, and 250km for autocomplete. It _might_ make sense to switch to the `exponential` decay function, which would allow even very far away results to be compared by distance.

ivanakimov:
Nice, glad to see there's some progress on this! Here are some issues we hear about often from our customers (maybe this will help in your tests):

-  Searching for `Birmingham` is problematic for US and UK customers, as they obviously prefer to see their country's city.
- `93277` is another search. Both US and Finland have this postal code.

PR #1209, opened 2018-10-15
Use exponential scoring and consistent scale values for `focus.point`

orangejulius:
Our `scale` values for the `center_point` query used by Elasticsearch to score our `focus.point` queries was inconsistent between search and autocomplete. (50km for search, 250km for autocomplete).

This PR changes autocomplete to 50km, to be consistent with search. It's tough to judge which is the better value, but here's my reasoning:
- a smaller `scale` means scores drop off faster. So a smaller value means that only _very_ close records would have a high enough distance score to outweigh a far away record with high importance or population.
- We have tested the search endpoint much more in general, so this 50km value might be more well tested

Additionally, this PR changes the decay function from `linear` to exponential. I don't recall why we settled on linear, as it was very long ago, but I suspect it _might_ have been an attempt to prevent very far away records from being scored at all. In that case, our understanding of Elasticsearch at the time was incorrect.

By using exponential scoring, records of any distance will receive a non-zero score for the distance query. This means we can differentiate between otherwise identically-scoring records that are very far away from the focus point. This is a huge help when searching for administrative areas like localities, as well as postal codes.

Looking at the acceptance tests, I could not find any cases where these changes cause a failure. However, some [newly added](https://github.com/pelias/acceptance-tests/pull/484) acceptance tests now _pass_.

connects https://github.com/pelias/api/issues/1206

### DIFF
--- a/query/autocomplete_defaults.js
+++ b/query/autocomplete_defaults.js
@@ -26,7 +26,7 @@ module.exports = _.merge({}, peliasQuery.defaults, {
   'phrase:boost': 1,
   'phrase:slop': 3,
 
-  'focus:function': 'linear',
+  'focus:function': 'exp',
   'focus:offset': '0km',
   'focus:scale': '250km',
   'focus:decay': 0.5,
--- a/query/autocomplete_defaults.js
+++ b/query/autocomplete_defaults.js
@@ -28,7 +28,7 @@ module.exports = _.merge({}, peliasQuery.defaults, {
 
   'focus:function': 'exp',
   'focus:offset': '0km',
-  'focus:scale': '250km',
+  'focus:scale': '50km',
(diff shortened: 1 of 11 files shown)


missinglink:
https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-function-score-query.html

orangejulius:
It turns out from our testing that there is a small performance hit for this change (we saw a query from around 700 to around 800ms). Since some of our shorter text length autocomplete queries are already quite slow, we're going to look at some performance improvements like #1219 or #1215 before merging this.

orangejulius:
We have made great improvements in slow autocomplete in https://github.com/pelias/api/pull/1219, I think this can be merged now.

PR #1213, opened 2018-10-17
Utilize cutoff_frequency parameter in all queries

orangejulius:
This PR adds use of the Elasticsearch [cutoff_frequency](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/query-dsl-match-query.html#query-dsl-match-query-cutoff) feature across nearly all of our query clauses.

This feature helps with performance for queries that could potentially match _lots_ of documents by only looking at documents that match uncommon terms first.

It does this in a way that has no effect on the end result, but massively reduces the number of documents that are scored.

My understanding of the meaning of the parameter is as follows: what fraction of the documents in a shard must contain this term before it is treated as a filtering term, rather than a scoring term? The value most commonly seen in documentation is 0.01, meaning any term that is in more than 1% of documents will be used as a filter. It's worth testing different values to see which one gives the best performance, but my guess is there aren't huge wins to be had from extensive tuning of this variable (I'd love to be proven wrong).

Essentially, `cutoff_frequency` performs the "pseudo-stopword" handling that I suggested in https://github.com/pelias/schema/pull/310#issuecomment-429473365, except it does it automatically, without any work on our part such as crafting extra query clauses or coming up with a list of common tokens.

In my testing of the difficult `Washington University in St Louis` query from https://github.com/pelias/schema/pull/310, these changes have the following effect:

  | master | cutoff_frequency | change
-- | -- | -- | --
Document Hits | 67603219 | 4336420 | -93.59%
Lowest Latency Seen | 294 | 82 | -72.11%
Highest Latency Seen | 420 | 136 | -67.62%

This was using a full planet build for testing with 560 million documents. Before this change, a full 12% of the documents in the index were being searched.

Additionally, running a full set of acceptance tests showed exactly zero differences using this branch compared to master.

In addition to immediate performance improvements, this PR will allow us to make changes in the future to tokenize on whitespace, hyphens, or whatever we want, without fearing that extremely common terms (like Street, avenue, etc) will cause unacceptable query performance. Thus, it's basically a pre-requisite for merging https://github.com/pelias/schema/pull/310.

Early feedback on this PR as well as https://github.com/pelias/query/pull/88 appreciated

## References

https://www.elastic.co/guide/en/elasticsearch/reference/2.4/query-dsl-match-query.html#query-dsl-match-query-cutoff
https://www.elastic.co/guide/en/elasticsearch/guide/current/common-terms.html#common-terms
http://kempe.net/blog/2015/02/25/elasticsearch-query-full-text-search.html

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "pelias-logger": "^1.2.0",
     "pelias-microservice-wrapper": "^1.7.0",
     "pelias-model": "^5.5.2",
-    "pelias-query": "^9.7.0",
+    "pelias-query": "^9.9.0",
     "pelias-sorting": "^1.2.0",
     "predicates": "^2.0.0",
     "retry": "^0.12.0",
--- a/query/autocomplete_defaults.js
+++ b/query/autocomplete_defaults.js
@@ -20,11 +20,13 @@ module.exports = _.merge({}, peliasQuery.defaults, {
   'ngram:analyzer': 'peliasQueryPartialToken',
   'ngram:field': 'name.default',
   'ngram:boost': 100,
+  'ngram:cutoff_frequency': 0.01,
 
   'phrase:analyzer': 'peliasQueryFullToken',
   'phrase:field': 'name.default',
   'phrase:boost': 1,
   'phrase:slop': 3,
+  'phrase:cutoff_frequency': 0.01,
 
   'focus:function': 'exp',
   'focus:offset': '0km',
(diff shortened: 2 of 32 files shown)


orangejulius:
In a review @missinglink points out that using `cutoff_frequency` in combination with an `and` query doesn't accomplish much. This is a good point, and many of our existing queries use the `and` operator.

Many don't however, such as some autocomplete queries. It seems that we're very inconsistent about it's usage. One advantage of this PR is that we might be able to forgo usage of the `and` operator in more cases, where performance previously would have been prohibitive.

missinglink:
I'm a little worried about this feature just due to personal ignorance on the topic, I'd need to study it a bit more before I can give some considered feedback.

missinglink:
I just rebased master and added some new commits which coincide with the changes I made in `pelias/query` (per-field settings for `cutoff_frequency).

orangejulius:
With those changes this branch has no acceptance-test regressions and appears to be ready to go. I am gathering some performance stats now before merging.

orangejulius:
After a day of testing with real-world queries I can confirm this PR cuts off the long tail of slow search and structured search queries quite a bit!

In this chart, the master branch is the grey lines, and this branch is the white lines. There's a huge decrease in p99 latency for both endpoints. Autocomplete stays about the same, and reverse is of course not affected. So this is a good change :)
screenshot_2019-01-18_17-34-24

PR #1214, opened 2018-10-18
fix(autocomplete): detect the case where input text is unsubstantial

orangejulius:
It's possible for the `text` input to /v1/autocomplete to be of non-zero length after trimming whitespace and quotes, but still be insufficient to use for geocoding.

One common case is that it contains only commas, slashes, or other delimiters.

Our query logic currently does not handle this case, and will generate Elasticsearch queries that do not have a primary `must` clause and end up searching every document in the index. These queries are slow, take up cluster resources, and are not useful.

By detecting unsubstantial inputs, we can prevent this.


missinglink:
:+1:

PR #1218, opened 2018-10-24
debug interpolation

missinglink:
adds information about the interpolation service in the `?debug=true` output.

- time taken
- requests
- responses


orangejulius:
Good idea. Needs a rebase but otherwise good to go!

Issue #1161, opened 2018-06-13
Remove many hardcoded lists of layers and sources


orangejulius:
Pelias was written largely under the assumption that the number of possible layers (distinct types of records) was limited and unchanging. Since the Mapzen shutdown we've found many people wishing to use Pelias with new types of data. Instead of just addresses and venues, they'd like to add transit stops, or census tracts for example.

We need to go through the API code looking for cases where lists of layers (and sources, to a lesser extent) are hardcoded, and figure out ways to make them more extensible.

Some general guidelines:
## Assume there are lots of non-administrative area layers
Thsese should be essentially equivalent to `venues` but with additional semantic meaning. Transit stops is a good example here.

## Move towards the list of administrative areas being configurable as well
This is harder because admin areas also generally imply a hierarchy (countries contain regions, for example). In the long run, we want to support this not just in the API but placeholder and the PIP service as well.

## Allow addresses and streets to be a special case, to an extent
Addresses and streets are unique because they are part of well-formed postal addresses, so capturing that relationship and meaning is useful. However, our code currently assumes that venues, streets, and addresses are the _only_ non-administrative area layers, which isn't good going forward.

## Consolidate as much as possible into configuration driven lists
We [recently created the TypeMapping object](https://github.com/pelias/api/pull/1131), which is a great place to add any lists that must remain. All those lists should be configurable.

# Known hardcoded lists
- [ ] List of hardcoded non-admin layers: https://github.com/pelias/api/blob/master/controller/predicates/is_coarse_reverse.js#L4
- [ ] List of hardcoded non-admin layers: https://github.com/pelias/api/blob/master/controller/predicates/is_only_non_admin_layers.js#L9
- [ ] List of hardcoded non-admin **and** admin layers: https://github.com/pelias/api/blob/master/controller/coarse_reverse.js
- [ ] Big hardcoded list of admin layers: https://github.com/pelias/api/pull/780
- [ ] Hardcoded list of admin placetypes https://github.com/pelias/api/blob/master/helper/placeTypes.js

PR #1222, opened 2018-10-29
refactor dedupe middleware

missinglink:
ready to merge

this PR refactors the dedupe middleware for readability, the old code had some unusual conventions such as throwing instead of returning values and using `_.some` purely to break iterator execution.

the new code is *hopefully* much easier to read and edit.

I have made a couple of changes to the way the deduper works with custom data:
- all custom layers are treated as analogous to a venue
- when selecting a preferred match between a canonical source and a non-canonical source, the non-canonical one will be selected.

### DIFF
--- a/helper/TypeMapping.js
+++ b/helper/TypeMapping.js
@@ -1,6 +1,9 @@
 const _ = require('lodash');
 const elasticsearch = require('elasticsearch');
 
+// a list of the canonical sources included in the default Pelias configuration
+const CANONICAL_SOURCES = ['whosonfirst', 'openstreetmap', 'openaddresses', 'geonames'];
+
 var TypeMapping = function(){
 
   // A list of all sources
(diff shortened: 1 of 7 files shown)


helper/TypeMapping.js
```diff
@@ -1,6 +1,9 @@
 const _ = require('lodash');
 const elasticsearch = require('elasticsearch');
 
+// a list of the canonical sources included in the default Pelias configuration
+const CANONICAL_SOURCES = ['whosonfirst', 'openstreetmap', 'openaddresses', 'geonames'];
```
orangejulius: Let's move this to `pelias.json` to avoid more hardcoded lists of sources and layers (see https://github.com/pelias/api/issues/1161)

missinglink: Hmm.. the problem with moving it to somewhere it can be edited is that it's no longer canonical.

The whole idea with this is to be able to know which things differ from what we expect vs. what's configured.

orangejulius: Yeah, but it's also something that should be located centrally, I'm sure there's code in all our repos that needs this list. The idea is no one would change it, but that's not currently enforceable, unless we added it as _code_ to pelias-config.

orangejulius:
This is really great. I found some cases we should think about, that perhaps might fit in this PR, or might make sense for further improvements:

- if record one has only a postal code, and no other address fields, and record two has has no address fields at all, they won't be considered dupes
- Likewise, a record with an address and a record without an address aren't considered dupes. This is a bigger change but is currently happening for the Nike case, and is probably something Trimet wants. Otherwise, they can't "augment" an existing record by adding an address

helper/TypeMapping.js
missinglink: Resolved in https://github.com/pelias/config/pull/103

orangejulius:
Just gave this a whirl on a full planet build. All acceptance tests pass.

Do we know of any examples where this might change the deduping results in a full planet build, for example from the case where only one record had an `address_parts` object with postal code? It would be nice to add an acceptance test.

We have had several postal code deduping issues reported over the years, such as https://github.com/pelias/pelias/issues/541, but I don't currently know of a good test case.

PR #1223, opened 2018-10-29
feat(log): Add full clean context to parser logs

orangejulius:
Without the unmodified text input, it's hard to tell what cases is causing these messages.

Followup to https://github.com/pelias/api/pull/1220


missinglink:
It seems a bit redundant to pass both `(clean.text, clean)` as function arguments?

orangejulius:
Indeed. That is now fixed.

PR #1226, opened 2018-10-30
add postcode to variable store if we have it

missinglink:
This PR enabled https://github.com/pelias/query/pull/91 to function for `address_search_using_ids` queries

It's a simple change, we just add `postcode` data to the query variable store if we have it.

Note: although not strictly interdependent, it would be ideal to merge the query PR first and then merge this one after so it pulls in the update dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "pelias-logger": "^1.2.0",
     "pelias-microservice-wrapper": "^1.7.0",
     "pelias-model": "^5.5.2",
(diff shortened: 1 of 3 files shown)


orangejulius:
I haven't had a chance to test this yet, but very nice catch regarding these queries not taking the postalcode into account. Should be a good fix to have.

orangejulius:
For our history, this was first mentioned in https://github.com/pelias/pelias/issues/754#issuecomment-434883399

PR #1227, opened 2018-10-30
Update CI and Node.js

orangejulius:
Several changes regarding CI and Node.js versions
- The Docker build script now requires `bash`
- Support for Node.js 6 has been dropped (https://github.com/pelias/pelias/issues/752)
- Semantic-release was running (doing nothing but taking up build time) on all pull requests for no reason
- Readme is updated to point to a central version requirement document rather than listing the supported Node.js version
- Releases now run off the master branch (https://github.com/pelias/pelias/issues/749)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,4 +9,4 @@ jobs:
       - setup_remote_docker
       - run:
           name: Build and push image to Docker Hub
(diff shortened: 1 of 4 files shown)


orangejulius:
CI is so fast now! 🎉

PR #1228, opened 2018-10-31
improved deduping across languages

missinglink:
following on from https://github.com/pelias/api/pull/1222 this PR changes two behaviours:

- improved matching across languages
- matching on aliases

## improved matching across languages

The previous code would cycle through all the names in `item1` and first do a key-check on `item2`, only if the key also existed in the second item, it would then compare values, if they differed then the two items were considered different.

The issue with that approach is something like this, which is considered different:

```javascript
var item1 = {
  'name': {
    'default': 'Bern',
    'eng': 'Bern'
  }
};
var item2 = {
  'name': {
    'default': 'Bern',
    'eng': 'Berne'
  }
};
```

The new code does it differently, instead it first cycles through all the values in `item2`, comparing them to `item1.default`, then it cycles through all the values in `item1` and compares them to `item2.default`.

So it's essentially saying "If any of the language values matched the default value of the other item, consider them as a dupe"

[edit] I added another commit which now also considers the user agent language for comparison on top of the `default`.

## matching on aliases

We were only matching on the first item in an array of values, this code uses slightly more CPU but it considers all aliases in deduplication.

### DIFF
--- a/helper/diffPlaces.js
+++ b/helper/diffPlaces.js
@@ -54,7 +54,7 @@ function isParentHierarchyDifferent(item1, item2){
  * Compare the name properties if they exist.
  * Returns false if the objects are the same, else true.
  */
-function isNameDifferent(item1, item2){
+function isNameDifferent(item1, item2, requestLanguage){
   let names1 = _.get(item1, 'name');
   let names2 = _.get(item2, 'name');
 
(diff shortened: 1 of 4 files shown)


orangejulius:
I tested this one on dev, and it doesn't break any acceptance tests, but I also couldn't find any improvements. I'm guessing this requires some of the other deduping PRs as well

PR #1229, opened 2018-10-31
only check parent fields above the layer of the least granular match

missinglink:
This PR sets us up to be able to do better inter-layer deduping.

Prior to this PR we would iterate over all `parent` properties in `item1` and ensure that `item2` had identical ID values.

~~There was an exception to skip checking the parent property at the same level as the item placetype, I'm not sure why this rule existed, @orangejulius any ideas?~~

The new code only checks parent properties at ~~the same level as~~ less granular than the highest granularity item.

This will mean that if we have two different placetypes for comparison:

```
placetype = locality

locality: 'A'
localadmin: 'B'
country: 'C'
planet: 'D'
```

```
placetype = country

country: 'C'
planet: 'D'
```

the old code would expect the second document to have `localadmin` values set, so it would fail, the new code will only check `country` and `planet`

~~note: this code does not assume any placetypes are synonymous, that will need to be handled in another PR.~~

[edit] added another commit, crossed out errors

### DIFF
--- a/helper/diffPlaces.js
+++ b/helper/diffPlaces.js
@@ -39,11 +39,26 @@ function isParentHierarchyDifferent(item1, item2){
   if( !isPojo1 || !isPojo2 ){ return false; }
 
   // else both have parent info
-  // iterate over all the placetypes, comparing between items
-  return placeTypes.some( placeType => {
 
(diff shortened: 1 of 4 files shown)


missinglink:
included in https://github.com/pelias/api/pull/1230

PR #1230, opened 2018-10-31
Inter-layer deduping

missinglink:
branched off https://github.com/pelias/api/pull/1229, please merge that PR first.
[comparison](https://github.com/pelias/api/compare/dedupe_improved_parent_matching...inter_layer_deduping)

This PR enables inter-layer deduplication and provides a mechanism to prefer some layers over others.

Examples:

```
# before
1) Parramatta, NSW, Australia
2) Parramatta (C), NSW, Australia
3) Parramatta, NSW, Australia
4) Parramatta, NSW, Australia

# after
1) Parramatta, NSW, Australia
```

```
# before
1) Singapore
2) Singapore, Singapore
3) Central Singapore, Singapore

# after
1) Singapore, Singapore
2) Central Singapore, Singapore
```

I suspect this might make some acceptance-tests fail :)

### DIFF
--- a/helper/diffPlaces.js
+++ b/helper/diffPlaces.js
@@ -39,11 +39,26 @@ function isParentHierarchyDifferent(item1, item2){
   if( !isPojo1 || !isPojo2 ){ return false; }
 
   // else both have parent info
-  // iterate over all the placetypes, comparing between items
-  return placeTypes.some( placeType => {
 
-    // skip the parent field corresponding to the item placetype
-    if( placeType === item1.layer ){ return false; }
+  // iterate over all the placetypes, comparing between items
+
+  // we only want to check parent items representing places less granular
+  // than the highest matched layer.
+  // eg. if we are comparing layer=address & layer=country then we only
+  // check for differences in layers above country, so continent, planet etc.
+  let highestLayerIndex = Math.min(
+    placeTypes.findIndex(el => el === item1.layer),
+    placeTypes.findIndex(el => el === item2.layer)
+  );
+
+  // in the case where we couldn't find either later in the $placeTypes array
+  // we will enforce that all parent fields are checked.
+  if( highestLayerIndex === -1 ){ highestLayerIndex = Infinity; }
+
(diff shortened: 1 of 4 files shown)


missinglink:
Also worth noting that we currently consider all the placetypes listed in `./helper/placeTypes.js` as synonymous, we might consider changing this, although I don't think it's bad as it is.

orangejulius:
Just ran full acceptance tests, looks pretty good to me. There were 5 new failures, but they are mostly okay:

- A data issue was exposed, captured in https://github.com/whosonfirst-data/whosonfirst-data/issues/1365
- Interestingly, a search for [Vientiane, laos](http://pelias.github.io/compare/#/v1/search%3Ftext=Vientiane,%20laos) now returns _three_ records instead of two:
`whosonfirst:region:85673437`, `whosonfirst:region:85673433`, and `whosonfirst:locality:421168913`
- A query for `new york` expected both NYC and New York State. We should figure out how to handle this one

The final two failures were simply the expected layer changing from county and localadmin to locality, and aren't an issue.

missinglink:
hmm.. all these branches are getting confusing, this also really needs https://github.com/pelias/api/pull/1228 merged into work correctly.

I looked at the test cases and there was a bug which I've now fixed, so the `Vientiane` case now only returns one result.

I also excluded `continent` from the list of layers to dedupe, so that issue is also now fixed independently of the data fix.

The `NYC` vs. `New York State` I see as an improvement and one of the reasons we are doing this, from a user experience point-of-view it's unlikely that users would search for the state, and so we remove that entry in order to reduce the 'noise'.

It would be ideal if the state is returned for the query "New York State", where are you testing this? I tried the dev server but it seems to have something else running.

orangejulius:
Almost everything looks great in this PR. found one issue:

Pennsylvania, Illinois, (the city), should not be considered a dupe of Pennsylvania, USA (the state)
image
http://pelias.github.io/compare/#/v1/search%3Ftext=Pennsylvania

Issue #1108, opened 2018-03-24
Fix error handling code in `middleware/sendJSON`


orangejulius:
The error handling code in [middleware/sendJSON.js](https://github.com/pelias/api/blob/master/middleware/sendJSON.js#L25) has some potential problems:

- [ ] It relies on a module, [elasticsearch-exceptions](https://www.npmjs.com/package/elasticsearch-exceptions), that is 3 years old and reportedly only supports Elasticsearch 1.4
- [ ] In the case of errors, it sets the default error code to 400. This means that a lot of errors such as failures to connect to other services are reported as 400 instead of 500 errors. 400 errors should be reserved for cases where the user's input, such as query params, are incorrect.

PR #1231, opened 2018-11-01
Fix error code handling

orangejulius:
## Background

For a long time, Pelias has used 400 as the default HTTP error code, and only a select few Elasticsearch exceptions would result in an HTTP 500 response.

This has the effect of hiding a lot of times when something was in fact wrong.

Since HTTP 400 generally signals that the request from the user has been incorrectly crafted, sending 400 error codes instead of 500 is a big problem. Besides sending a misleading signal that it's user error, many user agents will not retry after a 400 response.

Additionally, it makes it harder to monitor the health of a Pelias install. There's no way to tell if users are sending lots of genuinely invalid requests, or if the service is unhealthy.

## Changes

This PR adds a new exception class, `PeliasParameterError`. All sanitizers now return errors that are instances of this class, and `middleware/sendJSON` checks if any errors it sees are instances of the class.

Requests that result in known sanitizer errors and one or two known Elasticsearch exceptions result in specific error codes. Everything else is now considered an unknown error and results in a 500.

The complexity of the error handling code is greatly reduced. As a bonus, we can now finally get rid of the 4 year old, massively out of date [elasticsearch-exceptions](https://www.npmjs.com/package/elasticsearch-exceptions) NPM module dependency.

Fixes https://github.com/pelias/api/issues/1108

### DIFF
--- a/middleware/sendJSON.js
+++ b/middleware/sendJSON.js
@@ -1,71 +1,49 @@
-var check = require('check-types'),
-    es = require('elasticsearch'),
-    logger = require( 'pelias-logger' ).get( 'api' ),
-    exceptions = require('elasticsearch-exceptions/lib/exceptions/SupportedExceptions');
+const _ = require('lodash');
+const es = require('elasticsearch');
+const logger = require( 'pelias-logger' ).get( 'api' );
+const PeliasParameterError = require('../sanitizer/PeliasParameterError');
 
-// create a list of regular expressions to match against.
-// note: list created when the server starts up; for performance reasons.
-var exceptionRegexList = exceptions.map( function( exceptionName ){
-  return new RegExp( '^' + exceptionName );
-});
+function isParameterError(error) {
+  return error instanceof PeliasParameterError;
+}
+
+function isTimeoutError(error) {
+  return error instanceof es.errors.RequestTimeout;
+}
+
+function isElasticsearchError(error) {
+  const knownErrors = [ es.errors.NoConnections,
+                        es.errors.ConnectionFault ];
+
+  return knownErrors.some(function(esError) {
+    return error instanceof esError;
+  });
+}
 
 function sendJSONResponse(req, res, next) {
 
   // do nothing if no result data set
-  if (!res || !check.object(res.body) || !check.object(res.body.geocoding)) {
+  const geocoding = _.get(res, 'body.geocoding');
+
+  if (!_.isPlainObject(geocoding)) {
     return next();
   }
 
-  // default status
-  var statusCode = 200; // 200 OK
+  const errors = geocoding.errors || [];
 
-  // vary status code whenever an error was reported
-  var geocoding = res.body.geocoding;
+  const errorCodes = errors.map(function(error) {
+    if (isParameterError(error)) {
+      return 400;
+    } else if (isTimeoutError(error)) {
+      return 408;
+    } else if (isElasticsearchError(error)) {
+      return 502;
+    } else {
+      return 500;
+    }
+  });
 
-  if( check.array( geocoding.errors ) && geocoding.errors.length ){
-
-    // default status for errors is 400 Bad Request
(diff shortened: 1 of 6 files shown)


middleware/sendJSON.js
```diff
@@ -1,71 +1,45 @@
...
+  });
+}
 
 function sendJSONResponse(req, res, next) {
 
   // do nothing if no result data set
-  if (!res || !check.object(res.body) || !check.object(res.body.geocoding)) {
+  const geocoding = _.get(res, 'body.geocoding');
+
+  if (!_.isObject(geocoding)) {
```
missinglink: `null` in JavaScript is an object, recently I've been using `_.isPlainObject(value)`.

what's your take?

middleware/sendJSON.js
```diff
@@ -1,71 +1,45 @@
...
-        logger.warn( 'unknown geocoding error string:', err );
-      } else {
-        logger.warn( 'unknown geocoding error type:', typeof err, err );
-      }
-    });
-  }
+  const errors = geocoding.errors || [];
+
+  const errorCodes = errors.map(function(error) {
+    if (error instanceof PeliasParameterError) {
```
missinglink: nit: `isTimeoutError` is a function but this `instanceof` check is inline, is there a reason they are different?

sanitizer/PeliasParameterError.js
```diff
@@ -0,0 +1,15 @@
...
+  toString() {
+    return this.message;
+  }
+
+  toJSON() {
+    return this.message;
+  }
+}
+
+module.exports = PeliasParameterError;
```
missinglink: no unit tests for this new class?

sanitizer/PeliasParameterError.js
```diff
@@ -0,0 +1,15 @@
+class PeliasParameterError extends Error {
+  constructor(message = '') {
+    super(message);
```
missinglink: Can message be either a string or type of error?

middleware/sendJSON.js
orangejulius: Just because it fit easily in one line, whereas the other's didn't as well. Good idea for a change though, the code reads much better if they all are functions. I'll update it

middleware/sendJSON.js
orangejulius: `_.isPlainObject` is probably a good idea in general.

In this case I'm not too concerned with checking the state of the `res.body.geocoding` object too carefully, as it is set to an object [unconditionally in middleware/geocodeJSON](https://github.com/pelias/api/blob/master/middleware/geocodeJSON.js#L41), but it's an easy change to make, so why not

orangejulius:
Regarding the Ciao tests, I would like to run them, but haven't been able to get them to work for at least a year. If you have more success, do let me know. They would be super valuable for testing this PR.

sanitizer/PeliasParameterError.js
orangejulius: No, personally I think that would be overtesting in this case, given the simple implementation. However I am going to add a new test or two in other places that use this class.

sanitizer/PeliasParameterError.js
orangejulius: It can only be a string. However it might make sense for other parts of the code to detect an attempt to stuff an error object in instead, and fix that. I will add a corresponding test.

sanitizer/PeliasParameterError.js
missinglink: Haha, yea I hear ya, just don't want any errors in our error handling code! :)

missinglink: Or more specifically, we *do* want an error, this error :)

sanitizer/PeliasParameterError.js
orangejulius: A test for the aforementioned case has now been added

PR #1238, opened 2018-11-08
refactor addressit sanitizer for readability

missinglink:
this file was super hard to read so I had a crack at making it legible

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
   },
   "dependencies": {
     "@mapbox/geojson-extent": "^0.3.1",
-    "addressit": "1.5.0",
+    "addressit": "1.6.0",
     "async": "^2.0.0",
     "check-types": "^7.0.0",
     "elasticsearch": "^15.0.0",
--- a/sanitizer/_text_addressit.js
+++ b/sanitizer/_text_addressit.js
@@ -1,10 +1,26 @@
-var check = require('check-types');
-var parser = require('addressit');
-var _      = require('lodash');
-var logger = require('pelias-logger').get('api');
+const check = require('check-types');
+const addressit = require('addressit');
+const _      = require('lodash');
+const logger = require('pelias-logger').get('api');
+
+/**
+  this module provides extremely basic parsing using two methods.
+
+  note: this code is old and well due for a makover/replacement, we
+  are not happy with either of these methods but they remain in place
+  for purely legacy reasons.
+
+  'naïve parser' provides the following fields:
+  'name', 'admin_parts'
+
+  'addressit parser' provides the following fields:
+  'unit', 'number', 'street', 'state', 'country', 'postalcode', 'regions'
+**/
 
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -37,7 +37,7 @@
   },
   "dependencies": {
     "@mapbox/geojson-extent": "^0.3.1",
-    "addressit": "1.5.0",
+    "addressit": "1.6.0",
```
orangejulius: lets handle this separately in https://github.com/pelias/api/pull/1122

missinglink:
Huh very weird, I was expecting it to be a no-op refactor, I'll have a look at those tomorrow.

orangejulius:
I told you, this wild old code will not go easily :P

missinglink:
Actually those failures look like the exact matching PR, are you sure the didn't get mixed up?

orangejulius:
I thought that too, and double checked. I will triple check in a minute

missinglink:
wow ok, so this refactor was a useful experiment to remember how all the autocomplete functionality works.

I wrote up some docs explaining the autocomplete logic here: https://github.com/pelias/api/wiki/Autocomplete-search-logic

missinglink:
This is fixed, it should now be a no-op refactor, which cleans it up significantly

sanitizer/_text_addressit.js
```diff
@@ -1,10 +1,25 @@
-var check = require('check-types');
```
orangejulius: yuss. I think now that we know and love lodash, we should eventually get rid of `check`. We are almost there

sanitizer/_text_addressit.js
```diff
@@ -14,106 +29,84 @@ function _sanitize( raw, clean ){
...
-
-function parse(query, clean) {
-  var getAdminPartsBySplittingOnDelim = function(queryParts) {
-    // naive approach - for admin matching during query time
-    // split 'flatiron, new york, ny' into 'flatiron' and 'new york, ny'
-
-    var address = {};
+// naive approach - for admin matching during query time
+// split 'flatiron, new york, ny' into 'flatiron' and 'new york, ny'
+var naïve = function(tokens) {
```
orangejulius: image

missinglink:
Just a note on this, I was intending it to be a `no-op` but in the end, it had one minor change to the logic.

The `admin_parts` object was super confusing to deal with since it could be an `object`, it could be `undefined` and also in the case of only matching `admin_parts` it would be set to `null` :man_shrugging: 

So I've made that much easier to use by ensuring it always returns an object, and you just check if the object is empty or not.

There was one part of the code which did an `hasOwnProperty` check on the `null` which was previously true and now is false.

The effect of the change is that for addressit parses which only returned `admin_parts` (the AUS/USA state abbreviations) we used to consider the tokens as complete, we now consider them incomplete.

Good to merge.

missinglink:
c8cw76gmohy11

Issue #1246, opened 2018-11-29
Search query with size=1 returns two results


orangejulius:
[/v1/search?size=1&text=berlin](http://pelias.github.io/compare/#/v1/search%3Fsize=1&text=berlin)

Returns
```
    1) Berlin, Germany
    2) Berlin, Germany
```

Perhaps this came in with some of the recent deduplication improvements?

PR #1248, opened 2018-12-03
fix for size=1 returning two results

missinglink:
This PR fixes a bug where `size=1` returns two results.

resolves https://github.com/pelias/api/issues/1246


orangejulius:
Cool. Nice easy fix :)

Issue #373, opened 2015-11-23
Use reverse geocoding of Who's on First ID's to enable boundary.wof on Search & Autocomplete


riordan:
# The Gist
Enable every place that's part of an administrative hierarchy in Who's on First to act as a `boundary`, letting users specify the administrative region (or regions) that a query should apply to.

# Additional Details
On `/search`, we offer `boundary.country`, which limits the matching results to only those places that have been found (by reverse geocoding during import) to be within that country. It's a very useful and very efficient query. It also works because we can tell our users to only use the fixed ISO country code abbreviations.

With Who's on First, each place in our search index will store the whole chain of WoF hierarchy ID's it's a part of. This combination of query efficiency + agreed upon ID's should make the kind of query we offer for countries available for any other administrative area. Users will just have to look up the WoF ID of the area of interest (be it through another Pelias/Mapzen Search query or through the [Spelunker](https://whosonfirst.mapzen.com/spelunker)).

This should also give us the kind of autocomplete cache efficiency we'd been looking for when we decided to forgo autocomplete bboxes.

Also it [will make Mr. @louh very happy](https://mapzen.com/blog/find-your-community)

hannesj:
+1

missinglink:
This issue is much closer to becoming a reality, let's deal with this one after our other [autocomplete improvements] are completed https://github.com/pelias/api/milestones/Autocomplete%20Improvements

missinglink:
FYI for external developers, the response geojson will soon be returning ids for the administrative areas which *could* be used to achieve this functionality, eg:

```javascript
"type": "Feature",
  "geometry": {
    "type": "Point",
    "coordinates": [
      -73.984537,
      40.740134
    ]
  },
  "properties": {
    "id": "node:2045635524",
    "gid": "openstreetmap:venue:node:2045635524",
    "layer": "venue",
    "source": "openstreetmap",
    "name": "DiDi Dumpling",
    "confidence": 0.893,
    "distance": 0.67,
    "country": "United States",
-> "country_id": "85633793",
    "country_a": "USA",
    "region": "New York",
-> "region_id": "85688543",
    "region_a": "NY",
    "county": "New York County",
-> "county_id": "102081863",
    "localadmin": "Manhattan",
-> "localadmin_id": "404522731",
    "locality": "New York",
-> "locality_id": "85977539",
    "neighbourhood": "Rose Hill",
-> "neighbourhood_id": "85868045",
[...]

PR #1249, opened 2018-12-03
Add boundary.wof parameter to search and autocomplete endpoints

thismakessand:
https://github.com/pelias/api/issues/373

Related PR: https://github.com/pelias/query/pull/93

Added optional 'boundary.wof' parameter to /search, /search/structured and /autocomplete endpoints. If present, it adds a multi_match filter to the elasticsearch query using parent.*_id as the fields. 

I didn't add it to the /reverse endpoint because I wasn't sure if it made sense, but let me know if I should add it there.
For the validation/sanitation it only checks that the param string is all digits, I wasn't sure if there was a better way to validate the WOF id.

Any feedback would be great!


sanitizer/_boundary_wof.js
```diff
@@ -0,0 +1,38 @@
...
+  // param 'boundary.wof' is optional and should not
+  // error when simply not set by the user
+  // must be valid string
+  if (check.assigned(wof_id)) {
+    if (!check.nonEmptyString(wof_id)) {
+      messages.errors.push('boundary.wof is not a string');
+    }
+
+    // must be only digits
+    else if ( /^\d+$/.test(wof_id) ) {
```
missinglink: Looks like it's possible for the `wof_id` to be a non-string value here, for instance, it could be an array if someone used the `[]` array URI scheme.

What would be the effect of calling `.test()` on a non-string value?

missinglink:
Thanks @thismakessand, looks great.

I'll take us a few days at least to read through and review your work, we're currently in transit from a conference in Australia so please bear with us.

@orangejulius this is a really great contribution, I'll schedule some time to do a more thorough review of the code.
Regarding adding a new API filter, any objections on your part? I will also have a think about it, we will need to update the documentation.

sanitizer/_boundary_wof.js
thismakessand: Do you know of a sample query that would result in a non-string value that I could use to test with?  I tried eg: ``boundary.wof[]=420551221`` and `` boundary.wof[0]=420551221&boundary.wof[1]=123 `` but they both get caught by the _single_scalar_parameters sanitizer.

orangejulius: Nice. Good to know one of @trescube's first ever contributions long ago is still preventing problems.

orangejulius:
This is an awesome PR, thanks @thismakessand. Like Peter said, a few days of jetlag recovery are needed for a proper review, but it's really good to see this feature be implemented. It will be big for a lot of people.

orangejulius:
Hey @thismakessand,
This PR looks great. I'd like to request two changes, which are only to the interface.

We'd like to support more administrative datasets than just WOF in the future, so it would be nice if the parameter was instead called something like `boundary.gid`. Additionally, the input would then have to be Pelias GIDs instead of WOF IDs. Pelias GIDs take the form of `source:layer:id`, and you can see many examples in the response from any Pelias query.

Since internally, we are a long way from supporting multiple administrative datasets, I think it's enough for you to only change the API interface. We can easily make any refactoring or rewriting of code as needed later, but changing the API interface to Pelias would be a breaking change for our users.

I think you existing `pelias/query` PR will work as it is. You'll only have to extract the ID field from Pelias GIDs, and can ignore the rest for now. Note that for OSM, the IDs contain a colon, so the procedure for extracting the ID field from a GID is to find the second colon, and take everything after that (rather than do something like `gid.split(':')[2])`.

@missinglink, let me know if this matches with what you think is best, and @thismakessand, let me know if you have any questions!

missinglink:
That sounds perfect, it would be great if the `boundary.wof` was future-proof enough to be able to one-day support other datasets.

I guess that also means coming up with another name for it, what about `boundary.gid`?

Should we support specifying multiple GIDs such as `boundary.gid[]=100&boundary.gid[]=200`? If so would that be AND or OR?

thismakessand:
Thanks - I should be able to make these changes this week.  For the ``source:layer:id`` gids, since the layer is specified, should I update the query to match only on that instead of a wildcard on all the parent ids?  eg, should ``whosonfirst:neighbourhood:85880873`` match on parent.neighbourhood_id only?

Also, should I add boundary.gid to the /reverse endpoint as well?

Thanks

orangejulius:
@thismakessand awesome. You shouldn't have to bother with specifying only a single layer. Elasticsearch is plenty fast at searching multiple fields and the ID itself is unique.

`boundary.gid` would make sense on the `/v1/reverse` endpoint too, so if you'd like to add it that would be great.

thismakessand:
@orangejulius I made the api changes requested and updated the name 'boundary wof' to 'boundary gid' everywhere (including pelias-query) so it's consistent.  Let me know if there's something I need to do re: the failing CI test,  it appears unrelated to anything I worked on and I can't reproduce it locally.

orangejulius:
@thismakessand awesome, thanks. Everything is looking great. The failing tests are just because the pelias/query PR wasn't merged. In a few minutes, once a new version of the pelias-query NPM package is published, I'll restart the tests and they should pass.

@missinglink as it stands, this code works great for everything except `/v1/search` when using Placeholder. Do we want to add this functionality to Placeholder too somehow? It would be a shame to have the parameter only work for certain queries.

thismakessand:
@orangejulius  Oops, I thought I tested the placeholder pathway but apparently not.  It looks like the placeholder results contain the ids in the hierarchy so I just pushed a new commit that adds a filter function to the placeholder controller, similar to how the boundary.country is working.

orangejulius:
@thismakessand thanks so much for this PR. I've tested it out, and it all looks great. I put some acceptance-tests together in https://github.com/pelias/acceptance-tests/pull/486 so this is all good to go.

We might end up writing a blog post about this feature at some point soon, would you mind if we credit you and perhaps learned a bit about your motivation for working on this, how it went in general, etc? Shoot us an email at hello@geocode.earth if so.

Thanks again :)

toregua:
> That sounds perfect, it would be great if the `boundary.wof` filter was future-proof enough to be able to one-day support other datasets.
> 
> I guess that also means coming up with another name for it, what about `boundary.gid`?
> 
> Should we support specifying multiple GIDs such as `boundary.gid[]=100&boundary.gid[]=200`? If so would that be AND or OR?

Hi Everyone,

I have a simple question.
Is it in the roadmap to have the availability to search on a boundary.gid array with multiple values ? (like explain by missinglink)
Thanks in advance

PR #1251, opened 2018-12-06
use fallback query for parses which contain a "query" and at least one admin field

missinglink:
Our current parser logic for `/v1/search` only uses a `libpostal` parse in the case of:
```
if (hasStreet(vs) || isPostalCodeOnly(vs) || isPostalCodeWithCountry(vs)) {
```

If the inputs don't match one of those conditions then we discard the `libpostal` parse and use `addressit` instead.

This PR adds an additional condition `isVenuePlusAdmin(vs)` so that we can use the `libpostal` parse for some venue matching queries.

This type of query is already supported by the `pelias/query` module, so no changes are required there.

In the case of a parse such as this:

```
[
  {
    "label": "house",
    "value": "starbucks"
  },
  {
    "label": "city",
    "value": "lake oswego"
  }
]
```

.. we can generate a venue query which targets all venues with a matching name within the admin location, if no venue by that name is found then we 'fallback' to returning the centroid of the most granularly matched admin area.

I'm not super familiar with this code and not 💯 on why this wasn't already the case, maybe @trescube might remember? I'm going with the assumption that it was just missed due to venues not being part of the original work.

### DIFF
--- a/query/search.js
+++ b/query/search.js
@@ -2,7 +2,10 @@ const peliasQuery = require('pelias-query');
 const defaults = require('./search_defaults');
 const textParser = require('./text_parser');
 const check = require('check-types');
-const logger = require('pelias-logger').get('api');
+const isVarSet = (vs, layer) => vs.isset(`input:${layer}`);
+const isPostalCodeOnly = _isPostalCodeOnly();
+const isPostalCodeWithCountry = _isPostalCodeWithCountry();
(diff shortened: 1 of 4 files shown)


orangejulius:
This change seems reasonable, I'll try to find time to do a quick test

missinglink:
This PR is pretty old and the issue which motivated it was just closed today as fixed.
I also don't remember enough about this to get it merged and released, so I'm going to close it.

PR #1253, opened 2019-01-14
fix stable sorting on node 11.6.0

missinglink:
unit tests are failing on node 11.6.0 due to unstable sorting, this PR fixes the issue and tests pass on both 10 & 11 now.

### DIFF
--- a/middleware/sortResponseData.js
+++ b/middleware/sortResponseData.js
@@ -1,4 +1,5 @@
 const _ = require('lodash');
+const stable = require('stable');
 
 const logger = require('pelias-logger').get('api');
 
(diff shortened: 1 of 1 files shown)


orangejulius:
nice

orangejulius:
Wait, can you explain how this fixes the problem? [Node.js 11 includes v8 version 7](https://github.com/nodejs/node/pull/22754#issuecomment-419551068) which [changes Array.prototype.sort to be stable by default](https://twitter.com/mathias/status/1036626116654637057). So my understanding is that using the `stable` NPM package should not be needed in Node.js 11.

orangejulius:
@missinglink I came across this again and was wondering if you could chime in with how this fixes things, as I mentioned in my last comment, before we forget everything about it.

missinglink:
honestly, I'm not really sure, I was working on another issue when I discovered this error.

so I just quickly fixed it and got back to the original task.

missinglink:
Is it possible that the stable sorting algorithms are different between core and the stable module?

PR #1254, opened 2019-01-14
[inter_layer_deduping] Compare parent layers above lowest record

orangejulius:
This expands the layers checked for hierarchy difference to be any layer
higher than the _lower_ of the two records being compared. Previously,
it was any layer higher than the _higher_ of the two record (another way
to think about it was it was only comparing hierarchy elements that BOTH
records have).

This fixes the issue where the city of Pennsylvania, Illinois was being
considered a duplicate of the state of Pennsylvania.

This is a PR into #1230, not master

### DIFF
--- a/helper/diffPlaces.js
+++ b/helper/diffPlaces.js
@@ -54,7 +54,7 @@ function isParentHierarchyDifferent(item1, item2){
   if( !isPojo1 && !isPojo2 ){ return false; }
 
   // if only one has parent info, we consider them the same
-  // note: this really shouldn't happen as at least on parent should exist
+  // note: this really shouldn't happen as at least one parent should exist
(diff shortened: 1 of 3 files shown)


Joxit:
I wrote a piece of code here for dedup based on coordinates https://github.com/Joxit/pelias-mapbox-gl-js/blob/253aca1642489974acd0e8984213095700a6e75e/pelias-geocoder.js#L173

For example [Boulevard Saint-Germain, Paris, France](http://pelias.github.io/compare/#/v1/search%3Ftext=Boulevard%20Saint-Germain,%20paris) will send 7 entries, but only one will be shown if you try it here [demo](https://joxit.github.io/pelias-mapbox-gl-js/demo).

I don't know if this could be interesting for pelias :smile:

PR #1259, opened 2019-02-19
Use ngram field when querying admin data using autocomplete endpoint

missinglink:
This PR matches the work done in https://github.com/pelias/schema/pull/347

Simply changes the target fields to use the newly created ngram versions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
     "ciao": "^2.0.0",
     "difflet": "^1.0.1",
     "istanbul": "^0.4.2",
-    "jshint": "^2.5.6",
+    "jshint": "^2.10.1",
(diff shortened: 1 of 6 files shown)


missinglink:
I ran the acceptance-tests today and found no regressions as a result of this PR.

orangejulius:
The acceptance tests in autocomplete mode look very promising with this change:

**Before**
Screenshot_2019-03-20_21-10-31

**After**
Screenshot_2019-03-20_21-10-47

As expected, there's lots of examples of the admin parts of queries improving a lot.
Screenshot_2019-03-20_21-12-53
Screenshot_2019-03-20_21-13-16

PR #1262, opened 2019-03-14
enable fallback queries when placeholder returns 0 results

missinglink:
This is a fix for nasty bug I've been trying to track down for some time 🎉 

We're noticing more and more erroneous `libpostal` parses, and in some cases they result in 0 results when we could have fallen back and used an alternate query approach to satisfy the query.

A good example is this query for Changi Airport in Singapore:

```
/v1/search?text=changi

no features
```

The reason for this is that `libpostal` is incorrectly parsing the input as a `city`:

```
"parsed_text": {
  "city": "changi"
}
```

Following on from the parse, the next step is to query placeholder for matching cities, which returns 0 results.

Looking at the code, I believe the intended fallback strategy is to re-parse the input using `addressit` and then perform a query more similar to how we do autocomplete.

The issue is with one of the predecates, it prevents `shouldDeferToAddressIt` from returning true and so all other query steps are skipped and 0 results are returned.

### DIFF
--- a/routes/v1.js
+++ b/routes/v1.js
@@ -224,8 +224,7 @@ function addRoutes(app, peliasConfig) {
   // defer to addressit for analysis IF there's no response AND placeholder should not have executed
   const shouldDeferToAddressIt = all(
     not(hasRequestErrors),
-    not(hasResponseData),
-    not(placeholderShouldHaveExecuted)
+    not(hasResponseData)
   );
 
   // call very old prod query if addressit was the parser


missinglink:
With this PR the query above shows the parse as:

```
"parser": "addressit",
"parsed_text": {}
```

... and returns many results, the top result being the airport.

orangejulius:
This looks good to me. Running against a full planet build I don't see _any_ differences in acceptance tests. We should add at least one for this case.

What's interesting is it seems like finding Changi airport is actually quite hard. Here's a few queries that don't work, even with this change:

* [/v1/search?text=changi](http://pelias.github.io/compare/#/v1/search%3Ftext=changi) (only returns admin areas called Changi with this change. without this change no results are returned.
* [/v1/search?text=changi airport](http://pelias.github.io/compare/#/v1/search%3Ftext=changi%20airport) - returns admin areas called "airport" with and without this change
- [/v1/search?text=changi international](http://pelias.github.io/compare/#/v1/search%3Ftext=changi%20international) - returns irrelevant venues and admin areas named "changi" with and without this change:
```
     1) Life Changing International Ministries, San Diego, CA, USA
     2) Changing Channels International Ministries, Little Rock, AR, USA
     3) Victory International Life Changing Ministries, Nashville, TN, USA
     4) Changing cabin, Routt County, CO, USA
     5) Changiz, Iran
     6) Changī, Iran
     7) Changī, Iran
     8) Changiāpahāri, India
     9) Changii, India
    10) Changi, Pakistan
```

And a few that do:
* [/v1/search?text=changi international airport, singapore](http://pelias.github.io/compare/#/v1/search%3Ftext=changi%20international%20airport,%20singapore)

missinglink:
Yea, that's interesting. It seems to work fine with a country filter so I think it's a global vs. local balance issue.

Here's some that work great when filtering by Singapore:

- http://pelias.github.io/compare/#/v1/search%3Fboundary.country=sg&text=sin
- http://pelias.github.io/compare/#/v1/search%3Fboundary.country=sg&text=changi%20airport

but this one could probably score better:

- http://pelias.github.io/compare/#/v1/search%3Fboundary.country=sg&text=changi

I'm not sure why that would be happening, as airports should [get a boost](https://github.com/pelias/openstreetmap/blob/08b84052775f70956ce3dd490f0fa7c7d61ce567/stream/tag_mapper.js#L77)

orangejulius:
Since a lot of the query results are the same, I think it comes down to the general approach of always querying Placeholder first (informed by the parse from libpostal), and then only if no results are returned asking Elasticsearch for results.

We don't have a way to identify that the results from Placeholder are not very relevant, and instead see if there are better results from Elasticsearch. I think it would be pretty hard to do so actually. But if there is a case where we might be able to solve it, airports is probably it.

PR #1263, opened 2019-03-18
revert ccb9bc8180535bc5a1e55a523843a53334bc1c07

missinglink:
This PR reverts commit `ccb9bc8180535bc5a1e55a523843a53334bc1c07`

I just merged https://github.com/pelias/api/pull/1259 because it's been reviewed and tested.
It's probably a better idea to leave some time between merging https://github.com/pelias/schema/pull/347 and the API PR in order to allow users to catch up.

The potential issue I'm trying to avoid with this PR is that when users with older indices upgrade to a new `pelias/api` docker image they will presumably have errors when trying to use autocomplete due to the ngrams indices being missing.

### DIFF
--- a/query/autocomplete_defaults.js
+++ b/query/autocomplete_defaults.js
@@ -53,47 +53,47 @@ module.exports = _.merge({}, peliasQuery.defaults, {
   'address:postcode:cutoff_frequency': 0.01,
 
   'admin:country_a:analyzer': 'standard',
-  'admin:country_a:field': 'parent.country_a.ngram',
+  'admin:country_a:field': 'parent.country_a',
   'admin:country_a:boost': 1000,
   'admin:country_a:cutoff_frequency': 0.01,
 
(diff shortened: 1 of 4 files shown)


missinglink:
I'd suggest waiting a calendar month before merging `ccb9bc8180535bc5a1e55a523843a53334bc1c07`, around 18 April 2019 should be long enough.

missinglink:
For reference, builds created at/after docker `master-2019-03-18-d7a828c041f17bb8155e8883ceb0713d4796a7fe` (commit `pelias/schema:d7a828c041f17bb8155e8883ceb0713d4796a7fe`) will work with `ccb9bc8180535bc5a1e55a523843a53334bc1c07` and anything before will not.

orangejulius:
Makes sense. A month might even be longer than needed. But holding off a little while seems reasonable.

PR #1272, opened 2019-04-05
Allow ?categories= and ?categories=true as an indicator to return cat…

NickStallman:
This will indicate that categories should be returned in a result, without performing any filter on the categories.

Allowed:
?categories
?categories=
?categories=true

### DIFF
--- a/query/reverse.js
+++ b/query/reverse.js
@@ -94,7 +94,7 @@ function generateQuery( clean ){
   }
 
   // categories
-  if (clean.categories) {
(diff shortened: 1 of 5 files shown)


NickStallman:
This isn't actually working just yet. Not entirely sure why but I'll investigate further.

NickStallman:
Actually this is working for ?categories=true
It's not working for ?categories= or ?categories
Not entirely sure why but if =true is the desired option then it's good to go.

Issue #1271, opened 2019-04-03
Add a switch to always return categories


NickStallman:
It would be useful in some cases to be able to return the categories in the results.
For example an autocomplete that shows a little icon or label next to each result to visually show the category.

The category is currently being returned but only if the category is being filtered.
E.g. adding ?category=transport to the query returns all the categories, but it appears to be impossible to access them otherwise.

This is handled by helper/geojsonify_place_details.js
```
function checkCategoryParam(params) {
  return _.isObject(params) && params.hasOwnProperty('categories');
}
```

I did find a very old ticket #118 which mentions returning the category if the details flag is true, however this flag doesn't seem to exist anymore?

Was that flag deliberately removed? Is there any specific reason why a similar flag couldn't be added to request all information to be in the results?

missinglink:
I did a bunch of work on this back in 2015 and the idea was that the taxonomy would be independently reviewed because it would end up becoming part of our public API, meaning that removing categories would break integrations.

That review never happened and so the feature was kind of 'dark launched', it's there but it's not really apparent that it's there.

Here's my evaluation from that time of other taxonomies which led to the structure we have now.

https://github.com/pelias/pelias/wiki/Taxonomy-v1

What are your thoughts on this? How would you feel about this being potentially unstable of we decided to remove a category?

orangejulius:
We have https://github.com/pelias/api/pull/1068 from a while back that proposes to fix this same thing.

I agree with @missinglink that at the current moment this is at best an alpha or beta feature, and we need to be careful to avoid committing to response content that we aren't sure we want permanent. 

On the other hand with that attitude we will likely never make progress unless there is a massive change in the prioritization of more work on categories. I would support updating the PR above to use a `pelias.json` config flag to enable category output. That would actually set us up for being able to disable categories at the schema level as well, saving space in Elasticsearch for those who aren't yet using it.

NickStallman:
I think the feature won't get much love while it's in this sort if 'there but not there' state.
As-is I think it's reasonably useful, in the context I'm looking at as an enhancement for an autocomplete that includes venues.

I think anyone seriously wanting to use the categories in production will likely maintain their own mapping file, possibly with some tweaks.
E.g. you can find my tweaks here: https://github.com/NickStallman/openstreetmap/blob/master/config/category_map.js
[...]

PR #1273, opened 2019-04-09
categories updates

missinglink:
This PR actions some of the feedback received in https://github.com/pelias/api/pull/1272 and https://github.com/pelias/api/issues/1271

The current status of `categories` is still that they are in open-alpha, use at your own risk.
Having said that, there are a couple of minor issues with the existing implementation which need to be addressed:

- enable filter for autocomplete (we were allowing the param to exist but not using it in the queries)
- reveal categories in geojson even when categories param is blank (but not when unset)

@NickStallman I would suggest this we merge this PR instead of https://github.com/pelias/api/pull/1272?

The reasons being that this also enables filtering for `autocomplete` and it works slightly differently in that any time `?categories` is included in the query string then they will be exposed in the `geojson`, it does not support the magic category `true` as I felt that might be a confusing convention to adopt.

resolves https://github.com/pelias/api/issues/1271
closes https://github.com/pelias/api/pull/1272

### DIFF
--- a/query/autocomplete.js
+++ b/query/autocomplete.js
@@ -54,6 +54,7 @@ query.filter( peliasQuery.view.sources );
 query.filter( peliasQuery.view.layers );
 query.filter( peliasQuery.view.boundary_rect );
 query.filter( peliasQuery.view.boundary_country );
+query.filter( peliasQuery.view.categories );
 query.filter( peliasQuery.view.boundary_gid );
 
 // --------------------------------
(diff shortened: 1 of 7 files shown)


missinglink:
I had to rush to finish this so I'd appreciate a review if someone has time :)

NickStallman:
I hadn't noticed that the categories filter didn't work in autocomplete. :)

I'll give it a go today and let you know how it performs.

NickStallman:
Swapped over to your branch on my dev server and it works perfectly.
I'll swap my production server to it as well. Looks good to merge.

missinglink:
I just added one additional commit https://github.com/pelias/api/pull/1273/commits/6ee479c0a413327e449fbaa2280af2d00a49caf2 which improves the handling of erroneous inputs (a mix of valid and invalid categories).

Prior to that commit it would fail to show any results and emit the warning if any one of the categories was invalid, now it will only show if *all* of the categories are invalid.

PR #1274, opened 2019-04-12
add address_layer_filter sanitizer

missinglink:
I took a look at our slow queries today and... fairly predictably most of them are single word inputs such as `calle`, `east`, `avenid`, `S` etc.
We're also noticing more slow queries with `autocomplete` than other endpoints, which is again fairly predictable for these sort of short inputs.

So this got me thinking about ways to improve performance for these queries...

When I looked at the results they all only return admin areas in the ~40 top results that we show, despite some of them matching 40 million+ documents in the index.

This PR improves performance based off the assumption that single-word inputs should not, and need not match results from the 'address' layer.

The rationale is that in order to specify enough information to retrieve an address the user must, at minimum enter both a housenumber and a street name.

Note: we cannot exclude other layers such as 'venue' (eg. Starbucks) or 'street' (eg. Gleimstraße) because they may have valid single-word names.

The address layer contains the most records by a large margin, so excluding address results where they are not nessesary will provide significant performance benefits.

Some care is taken to only apply the layer filter when the user has not already specified their target layers, so if they explicitely specify `address` as a target layer then address records will potentially still be included in the results.

It's only really required for `autocomplete` queries but I've enabled it for `search` too since the logic still applies and it can also benefit from the performance improvements.

### DIFF
--- a/helper/TypeMapping.js
+++ b/helper/TypeMapping.js
@@ -39,7 +39,7 @@ var TypeMapping = function(){
 
   /*
    * An object that has a key for each possible layer or alias,
-   * and returns either that layer, or gall the layers in the alias
+   * and returns either that layer, or all the layers in the alias
    */
   this.layer_mapping = {};
 };
--- a/helper/TypeMapping.js
+++ b/helper/TypeMapping.js
@@ -13,8 +13,8 @@ var TypeMapping = function(){
   this.layers = [];
 
   /*
-   * A list of all layers in each source. This is used for convenience elswhere
-   * and to determine when a combination of source and layer parameters is
+   * A list of all layers in each source. This is used for convenience elsewhere
+   * to determine when a combination of source and layer parameters is
    * not going to match any records and will return no results.
    */
   this.layers_by_source = {};
--- a/sanitizer/_address_layer_filter.js
+++ b/sanitizer/_address_layer_filter.js
@@ -16,7 +16,12 @@ const check = require('check-types');
  * The address layer contains the most records by a large margin, so excluding
  * address results where they are not nessesary will provide significant
  * performance benefits.
+ * 
+ * Update: added warning message to inform user when this functionality is enabled
+ * Update: added additional check that enforces that the input must also contain at least one numeral
  */
+
+const ADDRESS_FILTER_WARNING = 'performance optimization: excluding \'address\' layer';
 
 function _setup(layersBySource) {
 
@@ -42,13 +47,20 @@ function _setup(layersBySource) {
         return messages;
       }
 
-      // check that only a single word was specified
+      // count the number of words specified
       let totalWords = clean.text.split(/\s+/).filter(check.nonEmptyString).length;
-      if (totalWords < 2) {
+
+      // check that at least one numeral was specified
+      let hasNumeral = /\d/.test(clean.text);
+
+      // if less than two words were specified /or no numeral is present
+      // then it is safe to apply the layer filter
+      if (totalWords < 2 || !hasNumeral) {
 
         // handle the common case where neither source nor layers were specified
         if (!check.array(clean.sources) || !check.nonEmptyArray(clean.sources)) {
           clean.layers = layers;
+          messages.warnings.push(ADDRESS_FILTER_WARNING);
         }
 
         // handle the case where 'sources' were explicitly specified
@@ -66,6 +78,7 @@ function _setup(layersBySource) {
 
           // target all layers for the sources specified except 'address'
           clean.layers = sourceLayers.filter(item => item !== 'address'); // exclude 'address'
+          messages.warnings.push(ADDRESS_FILTER_WARNING);
         }
       }
 
--- a/sanitizer/_address_layer_filter.js
+++ b/sanitizer/_address_layer_filter.js
@@ -42,16 +42,38 @@ function _setup(layersBySource) {
         return messages;
       }
 
+      // default to using the full 'clean.text'
+      // note: this should already have superfluous characters removed
+      let input = clean.text;
+
       // no nothing if no input text specified in the request
-      if (!check.nonEmptyString(clean.text)) {
+      if (!check.nonEmptyString(input)) {
         return messages;
       }
 
+      // if a parser has removed tokens, use the parsed text instead, this
+      // is the text which will be queried against the 'name.default' field.
+      // @todo: this logic is duplicated from 'query/text_parser.js' and may
+      // be subject to change.
+      if (check.nonEmptyObject(clean.parsed_text)) {
+
+        // if 'addressit' or 'libpostal' identified input as a street address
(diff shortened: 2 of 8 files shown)


missinglink:
The user-experience changes were not noticeable in my tests, all the single-word inputs I tried (for the planet) currently only return admin areas.
I also prefer that you now see the `street` record until you specify enough information to query for addresses.

I'd be interested to see what effect it has on smaller geographies where there are fewer admin areas in the index.

missinglink:
I just pushed a second commit which also considers `clean.sources` if it's specified, so if the user asks for sources which contain no addresses then it's a no-op, otherwise we compute the layers which are part of the sources they asked for and then remove `address`, which means they don't get a filter condition to include places from `venue` when they specify `sources=wof`.

missinglink:
I just pushed another commit which adds two new features

#### Adds an additional numeric check

The rationale being that an address must contain at least one number to be considered an address.

I initially felt a bit uneasy about this, but after considering it for the weekend I think this is probably safe.
There are certainly addresses with no numeric house number, such as country houses in the UK, but Pelias considers these to be a 'venue' and so they will appear in the venue layer and will still be retrievable while the filter is in effect.

I think that this alone will have a hugely positive impact on autocomplete performance, although I think we should keep an eye out for any issues that might arise from it if there are any significant regressions then the numeric check can be disabled.

#### Emit a warning

Following on from the above, I think it's important to emit a warning to inform the user what has happened, especially in the case where they have provided either a very short string or a non-numeric input, it would be nice to see a warning so that they can author an informative bug report.

The warning message says `performance optimization: excluding 'address' layer`. I tried a few different versions but this one is short & to the point.

I also like this message because it lets users know that they can explicitly specify the `address` layer to disable the filter.

missinglink:
I have a small regional extract running on my laptop, for a query such as `/v1/autocomplete?text=n` the ES hits go down from ~676k to ~24k and the response time goes from ~67ms to ~3ms.

missinglink:
This PR is now ready for final review and testing.

Joxit:
Hi @missinglink,
I'm back in France, I'm on it now :)

NickStallman:
I think this resolves https://github.com/pelias/pelias/issues/781 which would be great.

I figure if someone is searching for a street, and it's not in OSM, then not returning random addresses is still fine. The issue there is lack of data not this patch really.

missinglink:
One change I'd consider making before merging this is to do the numeric and token length checks against the 'parsed' text rather than the raw 'clean' text

This would have the advantage of consider a query such as `/v1/autocomplete?text=a ca` to only have one token (not two) if the `ca` was parsed as a state.

orangejulius:
We've been running this as an experiment branch on geocode.earth for a few days now. Looking at our metrics for latency and Elasticsearch hit count, it shows a reasonable improvement. The standard acceptance tests do not show any issues, and the autocomplete acceptance tests actually show a fairly large improvement (probably because there are mostly non-address queries in our tests). And most of all, no one has complained about results yet!

@missinglink I think this is good to merge unless you did want to add anything else. Go ahead and merge it if not.

orangejulius:
Screen Shot 2019-04-23 at 11 31 58 AM
Here's a screenshot of our metrics. There's no drop in the _maximum_ for es_hits, which makes sense (many address queries start with a number), but there is a noticeable drop in the p99 latency and es_hit values! 🎉

missinglink:
I just added https://github.com/pelias/api/pull/1274/commits/7435c0c4aae03004a6a28352556bbf41e0c5c541 which uses the `parsed_text` value to test the token length (where available)

NickStallman:
I have found a small issue with this. It doesn't kick in if layers are specified.

I've got a query which queries most layers, but not all. It's just excluding some rather than querying a specific layer. Notably it doesn't query neighbourhoods which has too much garbage, but it also excludes a few of the others.
So I certainly do want this filter to kick in on that query, and if I get rid of the layer parameter it looks like it's working great.

Maybe only skip if only one layer is specified? Or make it a flag in the config file?

Tested just searching for a single word locality. With layers returns addresses, without layers it's perfect.

momentiris:
Hello! Perhaps this is not the best place to ask this but I found this while trying to understand how to use the autocomplete. I have pelias running in kubernetes with only imported custom csv data. 

The csv table looks like this:
`id,source,layer,name,housenumber,street,postcode,lat,lon`

...where I have specified  "address" on the `layer` column, and the street name and house number combined on the `name` column. With this setup, I am unable to do any valuable autocomplete queries, since I have to specify both a letter and a number in order to get results. If i only specify the address name, i.e /autocomplete?text=[streetname], I get the message telling me address has been filtered out due to performance reasons. I don't understand if I should get around this by explicitly passing the layers param to the query, or if I should structure the csv data in another way. Perhaps anyone here could point me in the right direction? Again, sorry for the wrong forum.

orangejulius:
@momentiris No problem. You definitely would have been justified in opening a new issue for this discussion, but since you found the PR involved it's alright

In your case I think the solution is to always pass `layers=address` in your queries. You have a bit of an unusual case since you don't have any other data besides addresses in your Pelias installation. Another option would be to import streets, in which case autocomplete queries would return the street until a user types part of a housenumber.

This PR is an important performance improvement for our full planet builds, so we'd like to keep it around, but with one of those two changes you should also be able to get better results.

momentiris:
@orangejulius Thank you for your response! I wanted to avoid opening an issue since I'm to green on your platform to decide if this is really an issue, or rather ignorance on my behalf, that I simply have not yet understood all the intricacies of how to properly structure and query various data sources.

 Ok, I understand. I'm planning to extend the datasources with oa/osm of Sweden later, so actually how I could use both that and custom csv data (appended to have more exact geolocation lookups in a certain area) is also a mystery at the time of writing. Is there a way to get autocompletion that includes both these types of data from typing just part of one street name/just the housenumber? I'm guessing part of the answer to this lies in how pelias parses and indexes oa/osm data and csv data correspondingly.

thanks again

orangejulius:
You should be able to use the standard OSM/OA importers along with your custom address data without any problem. By default, Pelias will deduplicate addresses that it identifies to be the same, and it should prefer your custom data in that case.

Another option to work around this performance optimization by the way, is we would gladly accept a pull request that put it behind a [config flag](https://github.com/pelias/api#configuration-via-pelias-config). We definitely recognize that it works better in places where the housenumber is written _before_ the street name, like here in the United States. You are probably not the only one who would prefer different behavior. Let us know if you're interested and we can help you along.

momentiris:
@orangejulius Awesome, thanks a lot for the clarification! Sweet, I'd be happy to!

missinglink:
@momentiris @NickStallman moving this discussion over to https://github.com/pelias/api/pull/1473 where I've drafted a PR to change this behaviour.

PR #1281, opened 2019-04-18
Support for multi `boundary.country` query parameter

Joxit:
Hi team!
I want to add a small feature : `boundary.country` for several countries.

## Background

I think this can be useful for companies working in several specific countries (eg all European countries/all French-speaking countries/USA and Canada only).

## How to use

It's the same query parameter `boundary.country`. If you want more than one country, add the other after a comma like this `boundary.country=FR,DE`.

There is no limit in the number of countries. I do not know if there will be performance issues if we put too many countries.

### DIFF
--- a/controller/placeholder.js
+++ b/controller/placeholder.js
@@ -77,7 +77,7 @@ function getLayersFilter(clean) {
 // return true if the hierarchy does not have a country.abbr
 // OR hierarchy country.abbr matches boundary.country
 function matchesBoundaryCountry(boundaryCountry, hierarchy) {
-  return !boundaryCountry || _.get(hierarchy, 'country.abbr') === boundaryCountry;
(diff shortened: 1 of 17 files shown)


orangejulius:
This is awesome! I'll try to test it out ASAP. As for performance issues, I suspect it will be fine. The worst case would be you specify all countries in which case it's the same as not having the parameter at all :)

orangejulius:
Okay, I just tested this out, and it works great as far as I can tell. I like the simplicity of the code and how it uses a standard, single `match` query with multiple tokens. Also thanks for updating the placeholder filtering code. Having two separate filtering systems is confusing and we have had bugs from that before.

Joxit:
Yeah, you are welcome :smile: I'm happy to help :smiley:

PR #1294, opened 2019-05-10
Allow postalcode + admin area queries

orangejulius:
We currently have very strict logic for which combinations of inputs the fallback search query supports. Any input that is out of alignment with those expectations will be deferred to the lower quality addressit query system.

This changes some of that expectation logic. Where previously it ignored inputs with a postcode and any other field, now a postcode in combination with any administrative fields are allowed.

This is so, for example, a user can query for '90210, California' and hopefully see postalcodes called 90210 that are in California first in the results.

Obviously postalcodes are complex so there might be a few cases that don't work, but it should help. As a bonus, it will remove fragility and complexity from our query generation.

### DIFF
--- a/query/search.js
+++ b/query/search.js
@@ -122,7 +122,7 @@ function generateQuery( clean ){
 }
 
 function getQuery(vs) {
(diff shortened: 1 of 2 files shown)


orangejulius:
All acceptance tests check out, I think this is safe to merge.

PR #1303, opened 2019-05-24
fix(package): Stick Geolib to version 2.0.24

Joxit:
They pushed v2.0.26 on npm with a breaking change


orangejulius:
I just pushed two commits to https://github.com/pelias/api/pull/1302 to take care of the breaking changes. It wasn't anything too complicated, just renaming some functions and updating a test because they changed the precision of their distance calculations slightly.

Joxit:
Oh yes, that's right ! Thx :smile:

PR #1311, opened 2019-06-14
display population & popularity in geojson when debug flag enabled

missinglink:
as per the title, this PR allows developers to view the `population` and `popularity` information when the `?debug=true` param is enabled.


missinglink:
related https://github.com/pelias/openstreetmap/pull/493

bboure:
Would it make sense to always expose popularity in the api?
This could be useful, for example, to display places differently depending on their popularity on a map? Like, with different pin size or colour.

missinglink:
I think it's best we hide it under the `debug` flag for now and see how it is, we can always make it public later.

PR #1312, opened 2019-06-14
limit max text length to 140 characters

missinglink:
This PR limits the `?text=` parameter to 140 characters in length, truncating it and emitting a warning if it is longer.

I also considered returning an error here but on reflection the warning and truncation is a nicer user experience.

Note: this is only for `search` at this stage, it doesn't include `autocomplete` which is in transition at the moment (although we should do this for autocomplete too).

### DIFF
--- a/sanitizer/_text.js
+++ b/sanitizer/_text.js
@@ -1,4 +1,5 @@
 const _ = require('lodash');
+const MAX_TEXT_LENGTH = 140;
 
 // ref: https://en.wikipedia.org/wiki/Quotation_mark
 const QUOTES = `"'«»‘’‚‛“”„‟‹›⹂「」『』〝〞〟﹁﹂﹃﹄＂＇｢｣`;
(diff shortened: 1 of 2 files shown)


missinglink:
The warning message is `param 'text' truncated to 140 characters`, although this isn't technically correct, it's 140 `codepoints`.
But yea, I think this is a more human-centric warning message, although I'm open to changing it.

missinglink:
The general idea here is to strengthen the server against invalid inputs, I've seen many cases in the past where users send entire XML documents.
These sort of queries use up a lot of resources in parsing, network overhead and elasticsearch query time.

orangejulius:
Definitely a good idea. Do you think some people would want to make it longer? I can imagine some Pelias users needing this to be configurable.

That said, based on queries going into geocode.earth, 140 characters looks good. We have occasional inputs up to 3000(!) characters, but most are under about 100:
Screenshot_2019-06-14_07-13-47

missinglink:
Cool graph, 140 was fairly arbitrary, I just wrote the longest query I could think of and added some extra space on the end!
I think that graph confirms my assumption that 99.9% of queries are under that value and the rest are likely user errors.

missinglink:
I'm happy with this as-is, we can open it up for easy configuration later if requested.

PR #1313, opened 2019-06-14
improved elasticsearch debugging

missinglink:
display more information about the elasticsearch request when in debug mode.

before:
```
      {
        "controller:search": {
          "queryType": {
            "autocomplete": {
              "es_result_count": 20
            }
          }
        }
      },
```

after:
```
      {
        "controller:search": {
          "queryType": {
            "autocomplete": {
              "es_took": 37,
              "response_time": 370,
              "retries": 0,
              "es_hits": 46359,
              "es_result_count": 0
            }
          }
        }
      },
```

### DIFF
--- a/controller/search.js
+++ b/controller/search.js
@@ -99,11 +99,6 @@ function setup( apiConfig, esclient, query, should_execute ){
           // store the query_type for subsequent middleware
           res.meta.query_type = renderedQuery.type;
 
-          const messageParts = [
-            '[controller:search]',
-            `[queryType:${renderedQuery.type}]`,
-            `[es_result_count:${message.result_count}]`
-          ];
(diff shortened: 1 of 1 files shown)


controller/search.js
```diff
@@ -102,11 +102,15 @@ function setup( apiConfig, esclient, query, should_execute ){
           const messageParts = [
```
orangejulius: Oh, is `messageParts` actually unused here? This looks like our old (pre-JSON) logging format that is hard to parse and doesn't really do anything useful anymore.

missinglink: Hmm.. good catch, I added a second commit to remove that code

orangejulius: Thanks. Its my fault for leaving it in long ago :)

PR #1314, opened 2019-06-14
Remove search_with_ids controller

orangejulius:
I was attempting to use the new debug output from https://github.com/pelias/api/pull/1313 only to realize it was not added to the `search_with_ids` controller, only the `search` controller.

These two controllers are very similar, so this PR combines them into one that can be used in all cases.


orangejulius:
The logic for the controllers is pretty complex and core to Pelias. While all the acceptance tests pass with this change I'd love as many eyes on it as possible to catch potential issues.

PR #1315, opened 2019-06-17
Use a consistent name for the final search query

orangejulius:
_Branched off of https://github.com/pelias/api/pull/1314, please merge that first_

The search endpoint uses multiple query types that are executed one after the other: if the first does not return acceptable results, the next is executed, and so on.

The last query in this list used to be the only query the search endpoint had. Partially because it's been around for a long time, it is referred to in several different ways across different parts of the code.

This change unifies all of them so that the query is called `search_addressit` everywhere.


missinglink:
@orangejulius I already rebased this on to `pelias_parser` and it merged cleanly, so feel free to merge this if you are ready.

PR #1316, opened 2019-06-17
discover sources & layers by querying elasticsearch

missinglink:
This is a feature I started a while back but never quite finished.
It allows users to set [api.targets.auto_discover](https://github.com/pelias/config/blob/master/config/defaults.json#L42) to `true` in order to query a list of `sources` and `layers` from elasticsearch when the api starts up.

It works by sending an aggregation to elasticsearch, this query can take some time (up to several seconds for a full planet build) and so isn't recommended for a production build. (which is why it's disabled by default).

It is, however, very useful for clients who use a lot of custom layer names and forget to update the list manually.
I caught an example of this today from a client, combined with the `address_layer_filter` can cause these custom layers to be excluded from search results when we were only intending to filter `address` documents.

It's probably worth enabling this for all docker projects by default and explicitly disabling it in production environments where either the mapping is well known ahead of time, or where startup performance is sensitive.

### DIFF
--- a/helper/type_mapping_discovery.js
+++ b/helper/type_mapping_discovery.js
@@ -12,7 +12,6 @@ const logger = require('pelias-logger').get('api:type_mapping_discovery');
 
 const DISCOVERY_QUERY = {
   requestCache: true,
-  preference: '_replica_first',
   timeout: '10s',
   body: {
     aggs: {
--- a/helper/type_mapping_discovery.js
+++ b/helper/type_mapping_discovery.js
@@ -1,0 +1,73 @@
+const _ = require('lodash');
+const elasticsearch = require('elasticsearch');
+const peliasConfig = require('pelias-config').generate();
+const logger = require('pelias-logger').get('api:type_mapping_discovery');
+
+/**
+ * This module allows discovery of the sources and layers used
+ * in an existing elasticsearch index.
+ * 
+ * note: this will override any previously configured type mappings.
+ */
+
+const DISCOVERY_QUERY = {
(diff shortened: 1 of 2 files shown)


helper/type_mapping_discovery.js
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * This module allows discovery of the sources and layers used
+ * in an existing elasticsearch index.
+ * 
+ * note: this will override any previously configured type mappings.
+ */
+
+const DISCOVERY_QUERY = {
+  requestCache: true,
+  preference: '_replica_first',
```
orangejulius: I'd suggest removing this line: unlike say, a traditional SQL database which might have many replicas which are perfect for expensive read-only operations, most Elasticsearch setups do not give different resources to a replica shard vs a primary shard, so it won't have any effect on performance.

orangejulius: Yes, actually this line should definitely be removed: `_replica_first` and other similar preference values will [be removed in Elasticsearch 7](https://www.elastic.co/guide/en/elasticsearch/reference/6.1/search-request-preference.html).

There's a big warning at the bottom of that page saying that those preference values didn't do what people expected.

missinglink: I see, the docs say:

>  They do not help to avoid inconsistent results that arise from the use of shards that have different refresh states

I wasn't actually considering that because we use a read-only index, it was more about trying to reduce the impact on the primary nodes, either way, it was only really wishful thinking.

I've removed it via an additional commit

PR #1317, opened 2019-06-17
fix(predicate): Fix address related layers detection

orangejulius:
This fixes an error in the logic from https://github.com/pelias/api/pull/1307.

The `address_search_with_ids` query can only return results in the address or street layers. Admin results would come from a previous query to the Placeholder service.

However, the predicate logic was allowing the query to execute if an admin layer was specified, which could result in the API returning results from the street or address layers when not intended.


orangejulius:
@missinglink the code _is_ a bit confusing, and takes a bit of reading through to grasp. Fundamentally, the issue this PR will fix is as follows:

Take any address input with an admin area, and append `layers=coarse`. You'll still get the address record back in violation of the layers parameter. For example:

https://pelias.github.io/compare/#/v1/search%3Flayers=coarse&text=1705%20E%202nd%20St,%20Casper,%20WY%2082601
Screenshot_2019-06-18_08-45-23

PR #1319, opened 2019-06-18
allow dynamic updates to type mapping, update README

missinglink:
- missed updating the README in https://github.com/pelias/api/pull/1316
- refactored the code to allow for a dynamically updating type mapping

The original code was written with the notion that the 'type mapping' would be static, so in some cases, the `setup()` function was pre-processing the source/layer mapping in order to cache the values.

This would result in incorrect behaviour for sources/layers where the new values set by `auto_discover` were not being used.

This PR removes any pre-processing of the sources/layers and also uses a new function which ensures that existing pointer references are kept up-to-date.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,8 @@ The API recognizes the following properties under the top-level `api` key in you
 |---|---|---|---|
 |`services`|*no*||Service definitions for [point-in-polygon](https://github.com/pelias/pip-service), [libpostal](https://github.com/whosonfirst/go-whosonfirst-libpostal), [placeholder](https://github.com/pelias/placeholder), and [interpolation](https://github.com/pelias/interpolation) services. For a description of when different Pelias services are recommended or required, see our [services documentation](https://github.com/pelias/documentation/blob/master/services.md).|
 |`defaultParameters.focus.point.lon` <br> `defaultParameters.focus.point.lat`|no | |default coordinates for focus point
-|`targets.layers_by_source` <br> `targets.source_aliases` <br> `targets.layer_aliases`|no | |custom values for which `sources` and `layers` the API accepts (See more info in the [Custom sources and layers](#custom-sources-and-layers) section below).
+|`targets.auto_discover`|no|false|Should `sources` and `layers` be automatically discovered by querying elasticsearch at process startup. (See more info in the [Custom sources and layers](#custom-sources-and-layers) section below).
+|`targets.layers_by_source` <br> `targets.source_aliases` <br> `targets.layer_aliases`|no | |custom values for which `sources` and `layers` the API accepts (See more info in the [Custom sources and layers](#custom-sources-and-layers) section below). We recommend using the `targets.auto_discover:true` configuration instead of setting these manually.
 |`customBoosts` | no | `{}` | Allows configuring boosts for specific sources and layers, in order to influence result order. See [Configurable Boosts](#custom-boosts) below for details |
 |`autocomplete.exclude_address_length` | no | 0 | As a performance optimization, this optional parameter allows excluding address results for queries below the configured length. Addresses are usually the bulk of the records in Elasticsearch, and searching across all of them for very short text inputs can be slow, with little benefit. Consider setting this to 1 or 2 if you have several million addresses in Pelias. |
 |`indexName`|*no*|*pelias*|name of the Elasticsearch index to be used when building queries|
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -88,7 +89,22 @@ The `timeout` and `retry` values, as show in in the `pip` service section, are o
...
-Pelias allows importing your own data with custom values for `source` and `layer`, however you MUST tell Pelias about them via `pelias.json` using the `targets.layers_by_source`, `targets.source_aliases` and `targets.layer_aliases` configuration parameters.
+Pelias allows importing your own data with custom values for `source` and `layer`.
+
+Custom sources and layers are not automatically detected, you MUST set `targets.auto_discover` to `true` in your `pelias.json` to make Pelias aware of them.
+
+The `auto_discover` functionality sends a request to elasticsearch in order to automatically discover sources and layers from elasticsearch when the API server starts-up.
+
+Be aware that the query sent to elasticsearch can take several seconds to execute the first time against a large index, the query is cached in elasticsearch for subsequent requests.
+
+If you are importing custom layers and are running a city or small region sized build then the start-up delay will likely not affect you, you can safely use `targets.auto_discover:true`.
```
orangejulius: This reads like the problem with the query to Elasticsearch is that it causes the API to wait around for a while. Really the danger is that it puts a lot of load on the Elasticsearch cluster. So maybe this should read more like

>Be aware that the query sent to Elasticsearch can take several seconds to execute the first time against a large index, potentially impacting the impact on other queries hitting Elasticsearch. The query is cached in Elasticsearch for subsequent requests.
> 
> If you are importing custom layers and are running a city or small region sized build then the impact of this query will likely be too small to notice, you can safely use `targets.auto_discover:true`

README.md
```diff
@@ -88,7 +89,22 @@ The `timeout` and `retry` values, as show in in the `pip` service section, are o
...
+
+For advanced users running a full-planet build with custom layers or sources, and also concerned about this start-up delay, you have two options:
+
+1. execute the `auto_discover` query once manually to prime the cache **or**
+2. set `targets.auto_discover: false` and manually define the layers as documented below.
+
+> note: manually assigning targets is an advanced configuration which should be avoided where possible.
```
orangejulius: This last line feels a little preachy given that we already recommend using `auto_discover` where possible earlier. I think it can be removed.

README.md
```diff
@@ -88,7 +89,22 @@ The `timeout` and `retry` values, as show in in the `pip` service section, are o
...
+
+The `auto_discover` functionality sends a request to elasticsearch in order to automatically discover sources and layers from elasticsearch when the API server starts-up.
+
+Be aware that the query sent to elasticsearch can take several seconds to execute the first time against a large index, the query is cached in elasticsearch for subsequent requests.
+
+If you are importing custom layers and are running a city or small region sized build then the start-up delay will likely not affect you, you can safely use `targets.auto_discover:true`.
+
+For advanced users running a full-planet build with custom layers or sources, and also concerned about this start-up delay, you have two options:
+
+1. execute the `auto_discover` query once manually to prime the cache **or**
```
orangejulius: maybe something like "execute the `auto_discover` query once before any important traffic is hitting Elasticsearch"?

sanitizer/autocomplete.js
```diff
@@ -11,7 +11,7 @@ module.exports.middleware = (_api_pelias_config) => {
       size: require('../sanitizer/_size')(/* use defaults*/),
       layers: require('../sanitizer/_targets')('layers', type_mapping.layer_mapping),
       sources: require('../sanitizer/_targets')('sources', type_mapping.source_mapping),
-      address_layer_filter: require('../sanitizer/_address_layer_filter')(type_mapping.layers_by_source),
+      address_layer_filter: require('../sanitizer/_address_layer_filter')(type_mapping),
```
orangejulius: nice, this is way cleaner.

README.md
```diff
@@ -88,7 +89,22 @@ The `timeout` and `retry` values, as show in in the `pip` service section, are o
...
+
+Custom sources and layers are not automatically detected, you MUST set `targets.auto_discover` to `true` in your `pelias.json` to make Pelias aware of them.
+
+The `auto_discover` functionality sends a request to elasticsearch in order to automatically discover sources and layers from elasticsearch when the API server starts-up.
+
+Be aware that the query sent to elasticsearch can take several seconds to execute the first time against a large index, the query is cached in elasticsearch for subsequent requests.
+
+If you are importing custom layers and are running a city or small region sized build then the start-up delay will likely not affect you, you can safely use `targets.auto_discover:true`.
+
+For advanced users running a full-planet build with custom layers or sources, and also concerned about this start-up delay, you have two options:
```
orangejulius: same comment start-up delay > performance impact

missinglink:
Actioned PR feedback in https://github.com/pelias/api/pull/1319/commits/091b1ffdd61ee702ae543fef370cd54f0347a6e8

PR #1322, opened 2019-06-19
feat(addressSearchUsingIds): Set default street slop to 1

orangejulius:
Combined with https://github.com/pelias/query/pull/101, this change allows for improved matching of streets when using the addressSearchUsingIDs query (the first query executed by the search endpoint).

Essentially, by allowing a `slop` of 1 in the `match_phrase` clauses of the Elasticsearch query, its now possible to omit a single word _in the middle_ of a street name and still match it. Omitting words at the start or end of the street name has been supported since https://github.com/pelias/schema/pull/310.

Some examples of where this helps:

| Actual street name | Newly matching input text |
| ---                | ---                       |
| SE Martin Luther King Jr Blvd | Martin Luther King Blvd |
| Carrer de Balmes | Carrer Balmes |

### DIFF
--- a/query/search_defaults.js
+++ b/query/search_defaults.js
@@ -45,6 +45,7 @@ module.exports = _.merge({}, peliasQuery.defaults, {
   'address:street:analyzer': 'peliasStreet',
   'address:street:field': 'address_parts.street',
   'address:street:boost': 5,
+  'address:street:slop': 1,
   'address:street:cutoff_frequency': 0.01,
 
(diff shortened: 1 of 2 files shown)


orangejulius:
@Joxit i would be curious to hear if this PR allows any interesting new queries to pass in France :)

orangejulius:
The acceptance tests all pass with this change, and I'll also add some additional acceptance tests for new queries that now pass.

PR #1329, opened 2019-07-08
Query for old and new contents of `_id` field in `/v1/place` endpoint

orangejulius:
_This change completes step 2 of the `_id` field contents migration described in https://github.com/pelias/pelias/issues/672#issuecomment-508837195_

As part of the transition away from Elasticsearch types, we must change the contents of the `_id` field in our Elasticsearch documents.

This change modifies the `/v1/place` endpoint so that it queries Elasticsearch for documents based on both the old and new contents of the `_id` field, allowing the same API code to work regardless of the version of Pelias used to import data.

Connects https://github.com/pelias/pelias/issues/672

### DIFF
--- a/controller/place.js
+++ b/controller/place.js
@@ -33,13 +33,25 @@ function setup( apiConfig, esclient ){
     // setup a new operation
     const operation = retry.operation(operationOptions);
 
-    const cmd = req.clean.ids.map( function(id) {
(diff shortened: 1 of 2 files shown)


orangejulius:
@missinglink please review this PR and the associated [issue](https://github.com/pelias/pelias/issues/672) when you get a chance

PR #1330, opened 2019-07-10
replace usage of peliasQueryPartialToken and peliasQueryFullToken analyzers

missinglink:
replace usage of `peliasQueryPartialToken` and `peliasQueryFullToken` analyzers with the new `peliasQuery` analyzer.
related schema PR: https://github.com/pelias/schema/pull/370

this PR will ensure that query-time synonym substitution is disabled for clauses targeting the `name.*` fields.
see: the schema PR linked above for more info.

I am hoping to see a performance improvement as a result and expect to see 0 regressions 🤞

### DIFF
--- a/query/autocomplete_defaults.js
+++ b/query/autocomplete_defaults.js
@@ -16,12 +16,12 @@ module.exports = _.merge({}, peliasQuery.defaults, {
 
   'boundary:rect:type': 'indexed',
 
-  'ngram:analyzer': 'peliasQueryPartialToken',
+  'ngram:analyzer': 'peliasQuery',
   'ngram:field': 'name.default',
   'ngram:boost': 100,
   'ngram:cutoff_frequency': 0.01,
 
-  'phrase:analyzer': 'peliasQueryFullToken',
+  'phrase:analyzer': 'peliasQuery',
   'phrase:field': 'name.default',
   'phrase:boost': 1,
   'phrase:slop': 3,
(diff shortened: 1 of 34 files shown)


orangejulius:
I tested this out on a full planet build, and there are indeed no regressions! :tada: 

Thinking about it a bit further though, I doubt this would result in any performance improvements. Since we still are now exclusively using synonym expansion at index time, the same number of documents will still be matched (any document that is a match for `south`, for example, is a match for `s`, so having only one, instead of both tokens in the query wouldn't change the result).

So unless there are big performance wins from not having to consider a couple extra tokens when the document hit count is the same, I wouldn't expect us to see a difference.

But it can't hurt performance and if it keeps our analyzers a bit cleaner then it's still a win.

missinglink:
I figured it would need to do fewer lookups on the inverted-index and fewer bitmask operations for the same hit count.
It seems to me that considering variations of the input is unnecessary, but not sure about how much of a perf benefit will result.

orangejulius:
Okay, we took another look at this, and we definitely want to merge it! By avoiding synonym handling at query time, we can massively reduce the number of elasticsearch hits to consider. 

Here are some metrics showing the significance of the reduction in es hits:
Screen Shot 2019-07-30 at 10 36 27 AM


Additionally, while there aren't any acceptance test regressions, I noticed at least two improvements
Screen Shot 2019-07-29 at 5 09 41 PM
Screen Shot 2019-07-29 at 5 09 16 PM
https://pelias.github.io/compare/#/v1/autocomplete%3Fdebug=true&layers=street,address,venue,coarse&text=South%20Street,%20Augusta-Richmond%20County,%20GA,%20USA

We suspect that reducing the hit count dramatically (in this case from 75M to <5M) reduces the chance that a record would be erroneously scored above the "correct" result.

🎉

PR #1331, opened 2019-07-11
Netherlands use 'street number'

stvno:
The localized results of Pelias for the Netherlands are in the wrong order. The Dutch way is `street number`, not `number street` eg `damstraat 1`


orangejulius:
Thanks! Another country on the list :)

PR #1333, opened 2019-07-15
feat(search): Use minimum_should_match to filter hits

orangejulius:
This change utilizes the [minimum_should_match](https://www.elastic.co/guide/en/elasticsearch/reference/5.6/query-dsl-minimum-should-match.html) Elasticsearch query parameter to reduce the number of hits to search queries. For now, autocomplete is not changed, we can take a look at that later.

Previously, only one token in an input had to match, regardless of the input size. This allows for queries to potentially match a huge number of documents, especially as the number of tokens grows.

Now, most `ngrams` queries have the `minimum_should_match` parameter set to `1<-1 3<-25%`

Breaking this down, it means that the number of optional tokens follows this pattern:

| token count | optional tokens |
| --- | --- |
| 1   | 0 (obviously) |
| 2   | 1   |
| 3   | 1   |
| 4   | 1   |
| 5   | 1   |
| 6+  | at least 75% of tokens must match |

This should help ensure quality results are returned where possible, even if long inputs contain a few extraneous bits of information, while reducing the chance of extremely expensive queries.

### DIFF
--- a/query/search_defaults.js
+++ b/query/search_defaults.js
@@ -20,6 +20,7 @@ module.exports = _.merge({}, peliasQuery.defaults, {
   'ngram:field': 'name.default',
   'ngram:boost': 1,
   'ngram:cutoff_frequency': 0.01,
+  'ngram:minimum_should_match': '1<-1 3<-25%',
 
   'phrase:analyzer': 'peliasPhrase',
   'phrase:field': 'phrase.default',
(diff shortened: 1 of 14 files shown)


orangejulius:
We saw some very nice reductions in request latency and number of Elasticsearch hits due to this change. It looks like it will really help cut down on slow queries.

The effect of rolling this out was pretty dramatic:
Screenshot_2019-07-16_11-12-42

Issue #1280, opened 2019-04-17
Problems with pelias/api commit "feat(confidence): Add support for new query names"?


sayanb1983:
Hi @orangejulius ,

 I believe that there is a problem with your commit for ["feat(confidence): Add support for new query names"](https://github.com/pelias/api/commit/2a668612edc1ce383835f84b6bbf9da60bb40c7b). In middleware/confidenceScoreFallback.js: line 28, I see 2 potential problems:-

- Shouldn't res.meta.queryType be res.meta.query_type? 
- Shouldn't 'search_original' be present in  the include list along with ''search_fallback', 'address_search_with_ids' and 'structured'?

I would be happy to get your feedback regarding this.

orangejulius:
Hi @WaughWaugh,

I _believe_ everything is correct here. Yes, it's confusing that some places in the code use `queryType`, and others use `query_type`, but we have enough unit tests. And as https://github.com/pelias/api/pull/1224 taught us, the code breaks if the query type isn't passed along properly.

It's also confusing that `search_original` is not a query type in the list of queries, but as that query type has its own confidence score calculation, that is definitely correct.

What behavior are you observing that leads you to believe there's an issue?

WaughWaugh:
Thanks for getting back to me, @orangejulius.

I'm facing an issue with this address:

1328 COUNTRY RIDGE, INDIANAPOLIS, IN, 46234

Now, previously I was using pelias/api version 3.35.0 and was getting a match with a confidence level of 0.919. But after I upgraded to pelias/api version 3.60.0, I saw the confidence level drop to 0.8.

After debugging, I think the issue lies in this piece of code from middleware/confidenceScoreFallback.js:line 28,

```
if (['search_fallback', 'address_search_with_ids', 'structured'].includes(res.meta.queryType)) {
    return next();
  }
```

Here are my findings:-
- This search query runs through your processing pipeline as defined by your router and  gets processed by your search_original query, which returns the `renderedQuery.type` as 'search_original'.
- In your `controller/search.js`, you are using the `renderedQuery.type` property in 2 places,
     - line 62, you are setting the `queryType `of the message object with this above type, but message is used only for logging purposes.
    - line 100, you are setting `res.meta.query_type` with this above type.
- Now, in your `middleware/confidenceScore.js::computeScores()`, you are checking against this `res.meta.query_type`, which is correctly set by the code mentioned above.
[...]

PR #1334, opened 2019-07-15
Fix multiple bugs in confidence score handling

orangejulius:
There were multiple issues with the fallback query confidence score
calculation middleware.

1. Incorrect detection of handled query types. Essentially, the logic was negated
2. Incorrect list of supported query types. Not all list entries matched the actual query names
3. Incorrect parameter in mock data for unit tests. This combined with
the first bug to mask the issue in unit tests

Fixes https://github.com/pelias/api/issues/1280


orangejulius:
Looking at the acceptance test output, merging this might actually be problematic. It turns out a large number of queries return confidence scores from the newer, `confidenceScoreFallback` middleware. This middleware sets the `match_type` property (to, for example, `exact`, `fallback` or `interpolated`).

We already identified in https://github.com/pelias/pelias/issues/739 that the inconsistency in returning the `match_type` property is problematic for API consumers. Unfortunately, this would make that problem more prevalent.

It might be nice to figure out a reasonable way to set `match_type` for all queries before merging this.

Issue #1343, opened 2019-09-02
Flip street name and number for Hungarian addresses


laszbalo:
Hi,

just reading through the source code for some post-processor and found [here](https://github.com/pelias/api/blob/master/middleware/localNamingConventions.js#L5) that you flip the street number and address for some country.

FYI, Hungary should be on that list as well.

The first line of my address is:  `István király utca 14` (King Stephan street 14)

Also, would it not be possible to just get the list of those countries from some data source?

missinglink:
We could probably use this config file from the 'ID Editor' https://github.com/openstreetmap/iD/blob/master/data/address-formats.json

missinglink:
That config doesn't actually have Hungary listed, any idea why that might be @laszbalo ?

missinglink:
resolved in https://github.com/pelias/api/pull/1344

PR #1344, opened 2019-09-03
update 'flipNumberAndStreetCountries' config

missinglink:
update the country-code array to cover more country codes where the label format is "{{street}} {{housenumber}}".
sourced from the ID Editor config: https://github.com/openstreetmap/iD/blob/master/data/address-formats.json

resolves https://github.com/pelias/api/issues/1343

### DIFF
--- a/middleware/localNamingConventions.js
+++ b/middleware/localNamingConventions.js
@@ -2,7 +2,42 @@ const check = require('check-types');
 const _ = require('lodash');
 const field = require('../helper/fieldValue');
 
-var flipNumberAndStreetCountries = ['DEU', 'FIN', 'SWE', 'NOR', 'DNK', 'ISL', 'CZE', 'PRT', 'ESP', 'ROU', 'COL', 'NLD'];
+var flipNumberAndStreetCountries = [
+  /* from the ID Editor config */
+  /* https://github.com/openstreetmap/iD/blob/master/data/address-formats.json */
+  'AUT' /* Austria */,
+  'CHE' /* Switzerland */,
+  'DEU' /* Germany */,
+  'SVN' /* Slovenia */ ,
+  'POL' /* Poland */,
+  'AND' /* Andorra */,
+  'BIH' /* Bosnia and Herzegovina */,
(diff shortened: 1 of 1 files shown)


missinglink:
Yea I looked at that, it was the last one in the config, but my very unscientific approach was to google '{country} national library' and look at the address format returned.

For China I got `33 Zhongguancun S St, Zi Zhu Qiao, Haidian Qu, China`, but maybe this is either incorrect or an odd case?

Joxit:
Oh, you are right, it seems that when you write an address in Chinese, the housenumber is after the street, but when it's translated in English the housenumber is before the street.

For example the address of the French Embassy in China is:
```
法国驻华使馆
北京市朝阳区天泽路60号 100600
```
With the Chinese order:
法国驻华使馆: French Embassy in China
北京市: Beijing
朝阳区: Chaoyang District
天泽路: Tianze Road
60号: N°60
100600: postalcode ?

But we write in English: 60 Tianze Rd, San Yuan Qiao, Chaoyang Qu, Chine, 100600

This works almost like Japanese addresses :thinking:

missinglink:
Oh yea that's super weird 😄 
So I think its safest to merge this as-is since we don't have the 'smarts' to handle `JPN` and `CHN` properly?

Joxit:
I took the address from the embassy website in [Chinese](https://cn.ambafrance.org/%E5%8A%9E%E5%85%AC%E6%97%B6%E9%97%B4%E5%92%8C%E8%81%94%E7%B3%BB%E6%96%B9%E5%BC%8F,1219) and [French](https://cn.ambafrance.org/Contacts).

missinglink:
There is this cool lib https://github.com/wooorm/franc which I've been considering using for some time.

I just wanted to do this PR quickly to solve the linked issue but if anyone has time to do those countries properly using `franc` then we could improve how we handle them.

orangejulius:
Nice. It's good to have (hopefully) handled this completely rather than waiting for indivdiual PRs to come in.

Issue #1261, opened 2019-03-13
Phrase queries in Elasticsearch 6


nickpeihl:
Hey all. I wanted to reach out about your efforts to migrate to [Elasticsearch 6](https://github.com/pelias/pelias/issues/719). I'm very interested in helping out. 

Many of the phrase queries in the unit test fixtures of pelias/api fail because the match queries have a [`type: 'phrase'`](https://github.com/pelias/api/blob/master/test/unit/fixture/autocomplete_boundary_country.js#L13) parameter that is no longer supported in Elasticsearch 6+. This has been replaced by the `match_phrase` query type.

The [`match_phrase` query type](https://www.elastic.co/guide/en/elasticsearch/reference/6.6/query-dsl-match-query-phrase.html) does not support `cutoff_frequency` and `operator` parameters. However, these parameters are still supported in the [`match` query type](https://www.elastic.co/guide/en/elasticsearch/reference/6.6/query-dsl-match-query.html). The `match` query type no longer supports `slop`. 

For an example, let's look at https://github.com/pelias/api/blob/master/test/unit/fixture/autocomplete_boundary_country.js. There are two possible ways to make it compatibile with ES 6.

1)
```diff
diff --git a/test/unit/fixture/autocomplete_boundary_country.js b/test/unit/fixture/autocomplete_boundary_country.js
index 5da3698..2ad0df6 100644
--- a/test/unit/fixture/autocomplete_boundary_country.js
+++ b/test/unit/fixture/autocomplete_boundary_country.js
@@ -3,16 +3,14 @@ module.exports = {
     'bool': {
       'must': [{
         'constant_score': {
-          'query': {
+          'filter': {
             'match': {
               'name.default': {
                 'analyzer': 'peliasQueryPartialToken',
                 'boost': 100,
                 'query': 'test',
                 'cutoff_frequency': 0.01,
-                'type': 'phrase',
-                'operator': 'and',
-                'slop': 3
+                'operator': 'and'
               }
             }
           }
```

2)
```diff
diff --git a/test/unit/fixture/autocomplete_boundary_country.js b/test/unit/fixture/autocomplete_boundary_country.js
index 5da3698..6b1e87e 100644
--- a/test/unit/fixture/autocomplete_boundary_country.js
+++ b/test/unit/fixture/autocomplete_boundary_country.js
@@ -3,15 +3,12 @@ module.exports = {
     'bool': {
       'must': [{
         'constant_score': {
-          'query': {
-            'match': {
[...]

PR #1354, opened 2019-10-01
ES6 compatible queries

orangejulius:
This PR uses the work in https://github.com/pelias/query/pull/109 to convert our Elasticsearch queries into an ES6 compatible format.

The main change is that we can no longer set the `phrase` property on `match` queries. Phrase queries must now use the `match_phrase` query which makes the distinction between the two query types much more clear

Note: this PR is currently based on top of https://github.com/pelias/api/pull/1356, which should be merged first.

Closes https://github.com/pelias/api/issues/1261
Connects https://github.com/pelias/pelias/issues/719

### DIFF
--- a/query/search_defaults.js
+++ b/query/search_defaults.js
@@ -22,11 +22,10 @@ module.exports = _.merge({}, peliasQuery.defaults, {
   'ngram:cutoff_frequency': 0.01,
   'ngram:minimum_should_match': '1<-1 3<-25%',
 
-  'phrase:analyzer': 'peliasPhrase',
-  'phrase:field': 'phrase.default',
-  'phrase:boost': 1,
-  'phrase:slop': 2,
-  'phrase:cutoff_frequency': 0.01,
+  'match_phrase:main:analyzer': 'peliasPhrase',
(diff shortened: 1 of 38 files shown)


orangejulius:
Turns out there are a bunch of acceptance test failures on this branch. Its possible that the previous queries were not being interpreted by Elasticsearch as `match_phrase` queries. Some investigation will be needed to decide how to proceed. We can turn all the `match_phrase` queries into `match` queries if we need to though.

orangejulius:
Turns out the error was due to a subtle mistake while converting the queries to ES6:

https://github.com/pelias/api/blob/d117af1194889cdcbdd9423750b0d797bc8dfcfa/query/view/phrase_first_tokens_only.js#L20

The `input:name` variable was being overridden due to a line that should have been removed. This actually didn't have any effect on the query, but did cause the [max_character_count_layer_filter](https://github.com/pelias/api/blob/master/query/view/max_character_count_layer_filter.js) to exclude addresses since it believed the input text was only 3 characters. I'm struggling to figure out how to put a unit test in place to test for this scenario.

But the good news is that this code is now good to go!

PR #1356, opened 2019-10-01
feat(query): Use match_all for subquery matching

orangejulius:
This PR brings the `search_pelias_parser` queries into alignment with autocomplete and `search` queries by using the `match_all` query, instead of a phrase query, for use with function scoring to apply boosts for popularity and population.

While in theory this could result in longer query times because the population and popularity scoring will be applied to more documents, its unlikely we'll notice that.

However, what we _would_ notice is if there's highly populated or popular record that does not match the phrase subquery, but does match the non-phrase query.

All that said, there are no changes to the acceptance tests with this change, so we can mostly consider it a pure refactoring that reduces complexity and technical debt.

### DIFF
--- a/query/search_pelias_parser.js
+++ b/query/search_pelias_parser.js
@@ -24,8 +24,8 @@ query.score( peliasQuery.view.ngrams, 'must' );
 // scoring boost
 query.score( peliasQuery.view.phrase );
 query.score( peliasQuery.view.focus( peliasQuery.view.phrase ) );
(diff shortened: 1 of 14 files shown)


orangejulius:
After about two days of split testing, it looks like the performance changes due to this PR are minimal. There is a slight change to the average latency, but its probably within the margin of error for even a large number of queries.

Screenshot_2019-10-03_09-27-44

PR #1361, opened 2019-10-03
feat(addressSearchUsingIds): Set default street slop to 1

orangejulius:
This is a replacement PR for https://github.com/pelias/api/pull/1322. Due to a merge mixup (always use pull requests folks!) That PR was closed showing as merged.


orangejulius:
As expected, this does slightly increase the query time for search queries, since there are on average, slightly more documents matched. However the effect is very minor, around 5ms:
image

PR #1366, opened 2019-10-07
rewrite dedupe logic for simplicity & readability

missinglink:
I took a stab at a pure refactor of the `middleware/dedupe.js` file without changing any other files or tests.

The motivation here is to remove complexity by making the workflow explicit and easier to follow.

Note that the logging logic changed slightly, I'm not sure what is preferred here, I changed it to only print log lines when the deduplication logic resulted in a modification to results (previously it was pretty chatty).


missinglink:
Staged this today and there were no changes in the stats (which is to be expected with a pure refactor)

```
Aggregate test results
Pass: 607
Improvements: 5
Fail: 116
Placeholders: 0
Regressions: 0
Took 114276ms
Test success rate 100%
```

PR #1372, opened 2019-10-09
finer control over which layers deduplicate with each other

missinglink:
I think our deduplication logic isn't quite right when it comes to inter-layer deduping.

we currently have it set up to deduplicate between `locality` and `country`, this is nice for `Singapore` and `Luxemburg` but they're really the edge cases, it's undesirable for `Mexico City` and can cause issues for any other country in which there is a locality with the same name as the country (or one of its aliases).

a similar rule was put in place for `region` and `locality` with the intention of deduplicating `New York State`, `New York County` and `New York City`, although I don't believe this is totally correct.
I think in this situation *both* the State and the City should be shown but the County should be deduped by the City.

this PR adds some improved logic around which layers can be deduplicated with others.


NickStallman:
I wonder if deduping based on population or area might be viable?

Especially for autocomplete, you wouldn't want to dedupe areas with the same name but vastly different populations/areas as both options are likely valid.
But if the population/area is the same or similar then there is a high liklihood of duplication.
It would also handle things like Mexico City vs Mexico, and Singapore vs Singapore correctly.

orangejulius:
> I wonder if deduping based on population or area might be viable?

Yes @NickStallman this is a really good idea. Like you said, it would allow us to know that the city of Mexico City and the country of Mexico aren't the same and shouldn't be deduped, whereas the city and country of Singapore or Luxembourg are (basically) the same thing and should be deduped.

PR #1373, opened 2019-10-16
improved unicode support

missinglink:
This PR adds extended unicode support:
- normalize Unicode marks to their 'NFC' (composed) form.
- remove any control codes
- convert non-breaking spaces to regular spaces
- strip incorrectly used combining marks

Note this does not remove diacritics for either correctly decomposed glyphs or composed glyphs.

resolves https://github.com/pelias/pelias/issues/829

I added a second commit which strips a bunch of symbols unrelated to geocoding:
- arrows/box drawing/music symbols etc
- dingbats/emoji
- pictographs

It would be fun to support emoji (and we could do again later) but for now I think it's nicer to strip anything we cant use in order to reduce the downstream CPU costs on the parsers and in elasticsearch analysis.

### DIFF
--- a/helper/unicode.js
+++ b/helper/unicode.js
@@ -1,0 +1,41 @@
+const _ = require('lodash');
+const regenerate = require('regenerate');
+
+// non-printable control characters
+// ref: https://en.wikipedia.org/wiki/List_of_Unicode_characters
+const CONTROL_CODES = regenerate()
+  .addRange(0x0000, 0x001F) // C0 (0000-001F)
+  .add(0x007F) // Delete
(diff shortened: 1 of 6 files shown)


missinglink:
@Joxit FYI I added one more commit which removes a bunch of unicode marks which are unrelated to geocoding (including emoji!)

orangejulius:
Wow and here I was thinking this would be a really simple fix! Nice work. This is definitely worth a blog post to explain :)

missinglink:
Acceptance tests all passing on tag `unicode-2019-10-16-c25002633931ff55164ad63cbd036ec7ba976d57`, I'm going to merge this.

PR #1376, opened 2019-11-04
constant_score subqueries should use "filter" not "query"

missinglink:
I ran an elasticsearch@6 build last night and a bunch of acceptance tests failed with the message:

```bash
"[parsing_exception] [constant_score] query does not support [query], with { line=1 & col=150 }"
```

This makes total sense, the `constant_score` query is non-scoring and so should accept a 'filter' context and not a 'query' context.
Since elasticsearch@6 the 'query' top-level parameter is no longer accepted.

Looking at the docs, the `filter` parameter is [compatible back as far as 2.4](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/query-dsl-constant-score-query.html) so I don't foresee any backwards-compatibility issues.

There are also some query types which are supported by 'query' but not by 'filter' although we only seem to be using 'match_phrase' which is supported in both contexts.

### DIFF
--- a/query/view/ngrams_last_token_only.js
+++ b/query/view/ngrams_last_token_only.js
@@ -29,7 +29,7 @@ module.exports = function( vs ){
   // return the view rendered using the copy
   return {
     'constant_score': {
-      'query': ngrams_strict( vsCopy )
+      'filter': ngrams_strict( vsCopy )
     }
   };
 };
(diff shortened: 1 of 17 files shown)


orangejulius:
Nice, good catch. Interesting that the ES5 deprecation warning log didn't show this.

missinglink:
The docs don't mention using "query" in this context since ES@1.7, possibly showing our age 😄

missinglink:
I tested this on the staging server with an ES5 instance and there were no regressions, good to merge!

PR #1381, opened 2019-11-08
chore(pelias_parser): add logging for chosen parser solution scores

missinglink:
When I was writing up https://github.com/pelias/parser/issues/76 today I noticed that the 'scoring' in the Pelias parser is pretty good, it's fairly indicative of when the parse was low quality and you'll see scores < `0.5` next to a lot of incorrect parses 🎉 

So this PR is pretty simple, it adds logging of the solution scores so that we can graph them and get a better idea about how it's doing in a variable production environment.

example log line:
```
info: [api] pelias_parser_solution score=0.44
```

### DIFF
--- a/sanitizer/_text_pelias_parser.js
+++ b/sanitizer/_text_pelias_parser.js
@@ -95,6 +95,11 @@ function parse (clean) {
   // '    VVVV NN SSSSSSS AAAAAA PPPPP      '
   let mask = solution.mask(t);
 
+  // log information about the selected solution
+  logger.info('pelias_parser_solution', {
+    score: solution.score.toFixed(2)
+  });
+
   // special handling of intersection queries
   // here we do not trust intersection parses which also contain another
   // classification, such as a house number, postcode or admin field.


sanitizer/_text_pelias_parser.js
```diff
@@ -95,6 +95,11 @@ function parse (clean) {
   // '    VVVV NN SSSSSSS AAAAAA PPPPP      '
   let mask = solution.mask(t);
 
+  // log information about the selected solution
+  logger.info('pelias_parser_solution', {
+    score: solution.score.toFixed(2)
```
Joxit: It would be interesting to have the input too here

missinglink: The input is already being logged at https://github.com/pelias/api/blob/ef2969b0415a4d182247baa5e5624be73ac32186/sanitizer/_text_pelias_parser.js#L66

I decided to split up those log sections so the `pelias_parser` log entry provides an overview of the parse and then we can have one or more `pelias_parser_solution` lines which log info about specific solutions.

At this stage, we only pick the first solution, but we could explain this later and if we end up having multiple `pelias_parser_solution` log lines it'll be cleaner to have the input only logged once since it doesn't change.

Joxit: This was to have a simple way to analyze the logs.
For example if we want to quickly know which queries have a low score and the associated input all at once (using kibana or whatever). Unless it is possible to have links between log lines ? :thinking:

orangejulius:
It sounds like the purpose of this code is for manual investigation into a single query at a time, rather than providing data that we could analyze across many requests.

In that case, it might be better as part of the debug output, since it would then be along side all the other context from a request.

If we are trying to log something for later analysis, we'll need to do as @Joxit suggests and associate the solution score with a bunch of other data like the input text, or it won't be very useful.

Issue #1384, opened 2019-11-13
discuss: ES query cancellation


missinglink:
For years now we've been having issues with long-running search queries 'bouncing around the cluster' and causing performance issues for all other queries executed within the same timeframe.

The contemporary thought here is that this is unavoidable due to a lack of a proper cancellation method within elasticsearch itself which would allow cancelling of these renegade queries.

As a result, we have some configuration options such as `"requestTimeout": "120000"` (within pelias/config).
The idea here is that we pass the delay back to the user in order to prevent retries before the ES cluster has finished processing the request.

This always felt wrong to me, so I decided to revisit the topic last night and see what the current state of play is...

It turns out that elasticsearch does in fact have an internal query-cancellation method [documented here](https://www.elastic.co/guide/en/elasticsearch/reference/5.4/search.html#global-search-timeout) which allows us to specify a timeout either globally (via `search.default_search_timeout`) or per-request (using the `timeout` query property).

If the query times out you get a message in the response:
```js
{
  "took" : 7077,
  "timed_out" : true,
```

This is great in theory but if you try it out you'll find that queries can run well over the prescribed timeout and there are lots of people online complaining about this behaviour.

My understanding of the reasons for this is that, by default, the setting is only enforced at 'segment boundaries', which is to say that the query continues to run until it reaches a segment boundary, then it checks and only then does it error or continue processing.

For fairly obvious reasons that means that the query can continue to run for a lot longer than the prescribed timeout on large segments. In fact, if you do a force merge with `_forcemerge?max_num_segments=1` after generating a build (as is recommended for performance by the ES team) then you're getting no timeout at all, because there's only one segment!

Luckily there is [another setting which can be enabled](https://www.elastic.co/guide/en/elasticsearch/reference/5.4/search.html#global-search-cancellation) called `search.low_level_cancellation` which "improves cancellation responsiveness". My understanding of this is that (in addition to the segment boundary checks) there are additional checks performed at the completion of more granular operations such as sub-query sections like `match` or `terms` (this seems to be confirmed in the benchmarks linked below).

The caveat to this is that it comes with a warning about enabling the setting due to the performance penalties which would obviously apply when adding these sorts of checks all over the place.

> However, it comes with an additional overhead of more frequent cancellation checks that can be noticeable on large fast running search queries

[...]

PR #1390, opened 2019-11-14
improved_timeout_handling

missinglink:
following on from the discussion in https://github.com/pelias/api/issues/1384 this PR preemptively adds improved awareness of how elasticsearch handles timeouts.

**check for a 'timed_out=true' property in the response body**

It's possible for elasticsearch to return a `200 OK` status when in fact it's hit a timeout and is only returning a partially computed response.

I added some code to check this and return an `es.errors.RequestTimeout` error. In some ways I'm worried that this might conflate two separate signals, one for a low level socket timeout and one for a gateway (application-level) timeout. On the other hand it does what it says, and (I think?) our handling logic should really be the same regardless.

```
< HTTP/1.1 200 OK
< content-type: application/json; charset=UTF-8
< content-length: 215
<
{
  "took" : 5237,
  "timed_out" : true,
  "_shards" : {
    "total" : 12,
    "successful" : 12,
    "skipped" : 0,
    "failed" : 0
  },
  "hits" : {
    "total" : 0,
    "max_score" : 0.0,
    "hits" : [ ]
  }
}
```

**do not return `408` responses**

There is a discussion in the [elasticsearch issues](https://github.com/elastic/elasticsearch/issues/35582) about their use of `408`, I agree with them it's not a great HTTP code to use, but that's the way it is.

Luckily for us, we have the luxury of returning a `502` here because we actually do act as a gateway, so this is a much more appropriate code as its clearer that it's a server error and not a client error.

> The HyperText Transfer Protocol (HTTP) 502 Bad Gateway server error response code indicates that the server, while acting as a gateway or proxy, received an invalid response from the upstream server.

### DIFF
--- a/service/search.js
+++ b/service/search.js
@@ -4,7 +4,9 @@
 
 **/
 
-var logger = require( 'pelias-logger' ).get( 'api' );
+const _ = require('lodash');
+const es = require('elasticsearch');
+const logger = require( 'pelias-logger' ).get( 'api' );
 
 function service( esclient, cmd, cb ){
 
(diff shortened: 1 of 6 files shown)


service/search.js
```diff
@@ -21,13 +23,28 @@ function service( esclient, cmd, cb ){
...
+
     // map returned documents
     var docs = [];
     var meta = {
       scores: []
     };
 
-    if( data && data.hits && data.hits.total && Array.isArray(data.hits.hits)){
+    if (
+        _.has(data, 'hits') &&
```
orangejulius: I think you can remove this line, since the line below it will end up running the comparison `0 > 0` if data.hits is undefined

orangejulius:
Yeah, this looks good. For what it's worth I don't think Pelias returns many 408s. In our geocode.earth logs, I don't see any instances of it in the past several weeks.

service/search.js
missinglink: Yeah I was thinking that too

missinglink: ok so I ended up rewriting it with future though towards ES7 where they decided to make `hits.total` an object 🤷‍♂ 
anyway, it's nicer to not rely on checking another property which we don't even reference in the code.

I'll add it as a new commit for clarity.

missinglink: done, this what they changed it to BTW:

```js
"hits": {
        "max_score": 1.0,
        "total" : {
            "value": 2048,    
            "relation": "eq"  
        },
        "hits": ...
    }
```

PR #1392, opened 2019-11-20
Reset boosts for street matches, region_a, and country_a to 1

orangejulius:
These boosts, while possibly helpful for longer matches, can have very significant and negative impact on shorter autocomplete queries.


missinglink:
Looks good to me, I'm going to merge this now so I can build on top of it.
@orangejulius could you please update acceptance tests?

PR #1393, opened 2019-11-21
increase search query "focus:weight" from 2->3

missinglink:
It seems as though the autocomplete balance has changed slightly in ES6, causing `focus.point` search queries to not work as expected.

Setting "weight": 3 instead of "weight": 2 for the `function_score` query resolves this.

This issue doesn't affect `autocomplete` because it sets the `focus:weight` to 15!

I don't recall why there is such a large difference in these settings between `search` and `autocomplete` but I'd prefer to open another issue for that than include it with this work.


missinglink:
I ran the acceptance tests against our staging server (running ES5) and a dev environment (running ES6) and everything passes 🎉

PR #1398, opened 2019-11-25
Allow Elasticsearch host to be configured by ENV var

orangejulius:
We have long been pretty strict that nearly all configuration for Pelias should come through `pelias.json`, including all settings related to Elasticsearch.

While this is nice for consistency, there are times when setting configuration through `pelias.json` is not ideal.

In particular, if setting up an A/B split test, controlling parameters independently outside of `pelias.json` can make that setup much simpler.

So in this PR I propose that we allow setting the Elasticsearch host via environment variable. The value from the env var will override whatever is in `pelias.json`.

One downside I see to this PR is that it would pave the way for adding "just a few more" options to configuration via env var, such as the hostnames of other services. So it's possible we shouldn't go down that path, as I believe strongly that having a nice, structured JSON config file helps keep all the many options for Pelias organized.

### DIFF
--- a/routes/v1.js
+++ b/routes/v1.js
@@ -102,6 +102,15 @@ const Language = require('../service/configurations/Language');
 const Interpolation = require('../service/configurations/Interpolation');
 const Libpostal = require('../service/configurations/Libpostal');
 
+function getElasticsearchConfig(peliasConfig) {
+  const config = _.extend({}, peliasConfig.esclient);
+
+  if (process.env.ELASTICSEARCH_HOST) {
+    config.hosts = [ process.env.ELASTICSEARCH_HOST ];
+  }
+  return config;
+}
+
 /**
  * Append routes to app
  *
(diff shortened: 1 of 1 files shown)


orangejulius:
We have very quickly decided this isn't the way we'd like to go. Some arguments against the approach outlined in this PR:

- `pelias.json` is designed to be a "single source of truth" for configuration. So that for example when triaging a bug report, a Pelias maintainer can say "send me your `pelias.json`, and everything will be there
- This approach would quickly lead to a slippery slope problem, where we want to implement more and more functionality as ENV vars. For example, testing different Pelias service hosts (such as interpolation, Placeholder, etc) would probably follow closely
- This approach requires more code to write, test, and maintain. If people using Pelias want to implement more complex behavior or configuration management than having a single `pelias.json` file, they are free to do so, as working with files is already a powerful, flexible interface that nearly all infrastructure management systems can support.

PR #1402, opened 2019-11-29
fix js type detection bug in TypeMapping

missinglink:
This bug was reported today via email.

Due to the use of `_.isObject()` returning true for Arrays the `TypeMapping` functionality had a bug which caused Arrays to be incorrectly cleared when using `type_mapping_discovery`, an easy fix was to use `_.isPlainObject` instead.

An example of what the Array would look like on the command-line:

```bash
reference keys [ '0', '1' ]
reference cleared [ <2 empty items> ]
```

An example of the effect of this in a query:

```js
{
  "terms": {
    "layer": [
      "venue",
      "pois",
      "locality",
      "localadmin",
      "neighbourhood",
      "macrohood",
      "region",
      "continent",
      "country",
      "macroregion",
      null,
      null,
      null,
      null,
      null,
      null,
      null,
      null,
      null,
      null
    ]
  }
}
```

### DIFF
--- a/helper/TypeMapping.js
+++ b/helper/TypeMapping.js
@@ -91,7 +91,7 @@ TypeMapping.prototype.generateMappings = function(){
       return acc.concat(this.layers_by_source[key]);
     }.bind(this), []))
   );
-  
+
   safeReplace(
     this.layer_mapping,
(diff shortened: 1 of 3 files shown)


missinglink:
the client replied via email today that this PR has fixed their issue :tada:

PR #1404, opened 2019-12-02
remove npm check-types module

missinglink:
something I've been wanting to do for a while....
this PR removes the `check-types` module and replaces it with `lodash`.

the `check-types` module is quite convenient for functions such as `check.nonEmptyString`, however all the same functionality is available within the `lodash` module which we have made available pretty much across all repos.

this is intended as a no-op refactor and all the tests are passing afterwards.
the two libraries have slightly different functionality and I've done my best to check the source and replace them appropriately.

note: my editor is set up to strip extra whitespace and it also highlights when `require` statements are not used, so there was some cleanup there. I've also refactored some of the code to be more modern or read more easily.

closes https://github.com/pelias/api/pull/1403

### DIFF
--- a/middleware/accuracy.js
+++ b/middleware/accuracy.js
@@ -3,16 +3,15 @@
  * Accuracy level should be set for each item in the results.
  * The level can be any of the following:
  *  - point
- *  - interpolated
+ *  - interpolated (not currently used)
  *  - centroid
  */
 
-var check = require('check-types');
+const _ = require('lodash');
 
-var accuracyLevelPoint = 'point';
-var accuracyLevelInterpolated = 'interpolated';
-var accuracyLevelCentroid = 'centroid';
-
+const accuracyLevelPoint = 'point';
+// const accuracyLevelInterpolated = 'interpolated';
+const accuracyLevelCentroid = 'centroid';
 
 function setup() {
   return computeAccuracy;
@@ -20,7 +19,7 @@ function setup() {
 
 function computeAccuracy(req, res, next) {
   // do nothing if no result data set
-  if (check.undefined(res) || check.undefined(res.data)) {
+  if (_.isUndefined(res) || _.isUndefined(res.data)) {
     return next();
   }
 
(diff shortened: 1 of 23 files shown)


middleware/accuracy.js
```diff
@@ -3,24 +3,23 @@
  * Accuracy level should be set for each item in the results.
  * The level can be any of the following:
  *  - point
- *  - interpolated
+ *  - interpolated (not currently used)
```
orangejulius: huh funny. We do return `match_type: 'interpolated'` for interpolated results, but I guess the plan was to return `accuracy: 'interpolated'` too? It kinda makes sense, but also feels redundant, should we still do that?

middleware/confidenceScore.js
```diff
@@ -11,24 +11,24 @@
...
 const logger = require('pelias-logger').get('api');
-const check = require('check-types');
 const field = require('../helper/fieldValue');
 
 var RELATIVE_SCORES = true;
 
 function setup(peliasConfig) {
-  if (check.assigned(peliasConfig)) {
+  if (!_.isNil(peliasConfig)) {
     RELATIVE_SCORES = peliasConfig.hasOwnProperty('relativeScores') ? peliasConfig.relativeScores : true;
```
orangejulius: can't believe you didn't rewrite this line :P

lots of ways to make it better and shorter with `peliasConfig.get`

sanitizer/_address_layer_filter.js
```diff
@@ -1,22 +1,22 @@
 const _ = require('lodash');
-const check = require('check-types');
+const nonEmptyString = (v) => _.isString(v) && !_.isEmpty(v);
```
orangejulius: nice

middleware/confidenceScore.js
missinglink: I was just trying to do a no-op, was tempted in a few places to change a few things but resisted!

middleware/accuracy.js
missinglink: I'm not sure about the history of it, I just noticed that this was dead code and so I commented it out for clarity.

missinglink:
:wave: `check-types`

PR #1415, opened 2020-02-28
Uky

pclemente:

---

---

---


orangejulius:
Hi @pclemente,
The code you've changed is an important performance enhancement for Pelias. Can you tell us a little bit about what you're trying to accomplish? I'm sure there's a way to fix it without removing this feature.

PR #1287, opened 2019-05-02
replace addressit with pelias native parser

missinglink:
This PR replaces `addressit` with the new [pelias/parser](https://github.com/pelias/parser) module.

Ready for testing and review

Issue #142, opened 2015-06-26
[ngram branch] query weighting


missinglink:
@hkrishna I'm opening this ticket to track the admin weighting feature we discussed and so I have a place to put some test cases.

the gist of this is that on the *ngram* branch the part of the query which gave higher scores to queries which matched an admin value and scripts seem to have less effect than it used to.

missinglink:
different results for these?

"new york, new york" http://pelias.compare.s3-website-eu-west-1.amazonaws.com/#/search%3Fsize=20&input=new%20york,%20new%20york
"new york, ny" http://pelias.compare.s3-website-eu-west-1.amazonaws.com/#/search%3Fsize=20&input=new%20york,%20new%20york

missinglink:
"newark airport" http://pelias.compare.s3-website-eu-west-1.amazonaws.com/#/search%3Fsize=1000&input=newark%20airport

the `transport:air:aerodrome` entry should come higher than the other entries due to the category weighting script

missinglink:
http://pelias.compare.s3-website-eu-west-1.amazonaws.com/#/search%3Fsize=20&input=40%20dean%20street,%20brooklyn,%20ny

hkrishna:
yes! the script only acts like a score breaker.. 

Here's a query that uses functional_score and boosts ```_score``` value if the match happened to be a ```transport:air:aerodrome``` however, the side effect is that it always surfaces airports. We could use this query when we detect/determine that the user is searching for an airport - this needs to happen at the query_parsing stage in the API

```json
GET pelias/_search?search_type=dfs_query_then_fetch
{
   "query": {
      "function_score": {
         "query": {"bool": {
            "must": [
               {
                   "match": {
                      "name.default": "san francisco"
                   }
               }
            ], 
            "should": [
                { "match": {
                   "admin0": ""
                }},
                { "match": {
                   "admin1": ""
                }},
                { "match": {
                   "alpha3": "USA"
                }},
                {
                  "match": {
                    "admin1_abbr": ""
                  }
                }

            ]
          }
         },

         "boost": 1,
         "functions": [
            {
      "script_score": {
        "file": "category",
        "params": {
          "category_weights": {
            "transport:air": 2,
            "transport:air:aerodrome": 2,
            "transport:air:airport": 2
          }
        },
        "lang": "groovy"
      }
    }
        ]
      }
   },
   "size": 10,
   "sort": [
      "_score"
   ],
   "track_scores": true
}
```

[...]

PR #1420, opened 2020-04-17
Update pelias-parser

Joxit:
## Backgroud

Our parser rocks ! We stick the version of the perser in https://github.com/pelias/api/pull/1287/commits/c0749a0cf78a5118e6a62b1774d912d31510e67d.

But we did some good improvement since v1.38.0 such as 
- Supports for hyphen as alternative spans (pelias/parser#56)
- Add Classifier and Classification for road types (pelias/parser#65)
- Add EN toponyms for highway streets (pelias/parser#70)
- Add more place cases (pelias/parser#77)
- Add lang meta data in toponym for housenumber penalty (pelias/parser#72) 
- Fix parsing of "Empire State Building" (pelias/parser#84 **need release**)
- Add -kanal as seperable street suffix (pelias/parser#85 **need release**)
- Add support for unit type numbered (pelias/parser#87 to be merged)

## So?...

The parsing is evolving every day, and results can change and may [break CI](https://github.com/pelias/api/pull/1368)...

We definitely need to update the parser, but there are a few more questions...
- Should we reduce parser specific tests from the API ? Such as autocomplete-like test which can change at every releases.... 
- Should we duplicate tests in API and parser to monitor changes like `Calle Principal Bar` ?
- Should we stay like this and do manual updates ? :man_shrugging: 

I removed 2 tests due to parser response order
- `Calle Principal Bar`
- `Calle Principal Barc`

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "pelias-logger": "^1.2.0",
     "pelias-microservice-wrapper": "^1.7.0",
     "pelias-model": "^7.0.0",
-    "pelias-parser": "1.38.0",
+    "pelias-parser": "1.44.0",
     "pelias-query": "^9.14.0",
     "pelias-sorting": "^1.2.0",
     "predicates": "^2.0.0",
--- a/test/unit/sanitizer/_text_pelias_parser.js
+++ b/test/unit/sanitizer/_text_pelias_parser.js
@@ -206,8 +206,9 @@ module.exports.tests.text_parser = function (test, common) {
   cases.push(['Calle Principal ', { subject: 'Calle Principal' }, true]);
   cases.push(['Calle Principal B', { subject: 'Calle Principal' }, true]);
   // cases.push(['Calle Principal Ba', { subject: 'Calle Principal' }, true]); // jitter issue
-  cases.push(['Calle Principal Bar', { subject: 'Calle Principal' }, true]);
-  cases.push(['Calle Principal Barc', { subject: 'Calle Principal' }, true]);
+  // cases.push(['Calle Principal Bar', { subject: 'Calle Principal' }, true]);
+  // (0.91) ➜ [ { street: 'Calle Principal Barc' } ] && (0.91) ➜ [ { street: 'Calle Principal' }, { locality: 'Barc' } ]
+  // cases.push(['Calle Principal Barc', { subject: 'Calle Principal' }, true]);
   // cases.push(['Calle Principal Barce', { subject: 'Calle Principal' }, true]); // jitter issue
   // cases.push(['Calle Principal Barcel', { subject: 'Calle Principal' }, true]); // jitter issue
   // cases.push(['Calle Principal Barcelo', { subject: 'Calle Principal' }, true]); // jitter issue


missinglink:
Yeah good question.

Regarding `package.json`:

When `pelias/parser` was such a new project, we decided to pin to a specific version number for the 'big release', just so we could go through the extensive testing and be 100% sure we were releasing what we tested.

Now that the `pelias/parser` repo has matured we may choose to relax the version matching to a semver range since the public API is likely to stay fairly stable anyway, this will mean `feature:` and `fix:` commits would be pulled in automatically.

Regarding the tests:

The `test/unit/sanitizer/_text_pelias_parser.js` file is covering a mixture of the responses from `pelias/parser` with the [code used to interpret the solutions](https://github.com/pelias/api/blob/master/sanitizer/_text_pelias_parser.js), so I think it's hard to completely avoid testing `pelias/parser` unless we mock/stub it out in tests?

In general, I think it's better to add test cases directly to the `pelias/parser` repo where possible, rather than adding them to `pelia/api`.

Somehow its nice to have the tests failing, so we can see what changed and decide if we're ready to accept the new functionality yet or not (in this repo).

Although the flip-side of that is that we might end up maintaining tests in two places and then have to update them in both places for every change, albeit positive or negative.

I'm happy to continue as-is with the tests until they become a maintenance pain.
For the versioning, we can begin to relax the semver version?

@orangejulius thoughts?

orangejulius:
Despite it being a departure from our normal way of handling dependencies, I think realistically upgrading the parser version in the API has to be a manual operation.

The parser is not a simple library that we can comprehensively cover with unit tests, unlike something like `iso-639-3`, to pick another API dependency. Ultimately, **confidence in the parser comes from running the acceptance tests on a full planet or other large build**.

While we run acceptance tests against branches of the parser during development, the most straightforward place to decide "this code passes the acceptance tests for a given data build" seems to me to be the API. We could try to make that decision when merging into the parser, but that feels very indirect and we'd always have to do some level of testing at the API level anyway (not to mention fixing API unit tests as Joxit mentioned).

So my vote is we continue to periodically open PRs like this one, and then we can essentially do our normal API workflow: evaluate whether the changes in the PR are good or bad for the acceptance tests and go from there.

missinglink:
Can we please combine this PR https://github.com/pelias/api/pull/142
closes https://github.com/pelias/api/pull/1429

@orangejulius last I recall you were testing this, is it good to merge?

msschl:
I'm closing #1429 since it's a duplicate.

[v1.45.1](https://github.com/pelias/parser/releases/tag/v1.45.1) is the latest version of [pelias/parser](https://github.com/pelias/parser)

orangejulius:
I ran all our test suites on this PR comparing it to master with the same data. Overall there were only minor differences. A little jitter added here, a little taken away here.

Here are some screenshots of some diffs of the autocomplete test output:
Screenshot_2020-05-07_16-16-11
Screenshot_2020-05-07_16-15-49
Screenshot_2020-05-07_16-15-33


I think this PR in particular is good to go.

For later versions of the parser, there are issues with the Pelias unit tests. @missinglink if you can resolve those in a new PR, I'll test that one too.

PR #1425, opened 2020-04-28
refactor routes to be easier to read

missinglink:
I was looking in the `routes/v1.js` file today and It's very difficult to read due to all the code at the top of the file.
This PR is a cosmetic refactor to make the file read more easily.

I've used `require-all` which I know isn't everyones cup-o-tea, but it allowed me to delete a bunch of lines.
I've also used camel-casing for the predicate file names so they match the variable names, making them easier to follow.


orangejulius:
Looks like there might be some duplicate files currently though, the diff should be negative, yeah?
Screenshot_2020-04-28_08-31-57

missinglink:
@orangejulius yeah unfortunately so, the OSX case-insensitive FS is a pain in the ass...
Could you please remove the two files in the `test/unit/controller/predicates` dir that end in `Whosonfirst` (as opposed to `WhosOnFirst`) on your system?
I tried and couldn't figure it out 🤷

orangejulius:
@missinglink LOL sure thing. Done and force-pushed :)

Issue #1239, opened 2018-11-08
Autocomplete tests affected by pelias.json contents


orangejulius:
The following two test cases can fail based on a user's own `pelias.json` contents. Specifically, custom layers will cause the tests to not match the expected fixture.

We need to use `proxyquire` to pass in a specific pelias-config setup for these tests.

```
919 tests
 2371 passed
 2 failed  

 Failed Tests:   There were 2 failures

   x autocomplete_linguistic_one_char_token
   x autocomplete_linguistic_two_char_token
```

PR #1431, opened 2020-05-07
always use an empty pelias.json in test

blackmad:
I noticed that if I had a pelias.json in api/ or in ~/, it seemed like it would get used by tests and a couple of them would break because they were taking in some of my pelias configuration (other sources mostly).

To make unit tests always pass for developers, always point the unit test command at an empty pelias.json in test/

Closes https://github.com/pelias/api/issues/1239


package.json
```diff
@@ -15,7 +15,7 @@
     "start": "./bin/start",
     "test": "npm run unit",
     "travis": "npm test",
-    "unit": "./bin/units",
+    "unit": "PELIAS_CONFIG=./test/pelias-config-empty.json ./bin/units",
```
orangejulius, line 18: This actually would work great in `./bin/units` itself (I think). That way anyone who runs the unit tests through any mechanism would get the right PELIAS_CONFIG

orangejulius:
Ah yeah, we should have done this a long time ago. It looks like we need the actual empty `pelias.json` file, but otherwise we should absolutely merge this.

orangejulius:
I went in and fixed it all myself, and will take care of merging this from here. Thanks for not putting up with the cruft we've let sit for too long :P

Issue #1296, opened 2019-05-14
Autocomplete: Multi-lang search (based on user's lang)


Joxit:
> Transcription of https://github.com/pelias/api/issues/127#issuecomment-490002821

## What is this for ?

We want Pelias to send responses to queries written in other languages than English.
For example, a Dutch looking for `Parijs` (Paris in Dutch) will get `Parijs, Frankrijk`.

## What should we do ?

- [ ] Use `multi_match` in autocomplete queries
- [ ] Should search in `name.$LANG` index with higher boost
- [ ] Should search in `name.default` index with standard boost
- [ ] Should search in `name.en` index as fallback with lower boost (when `$LANG` is not `en`)

## Some use cases 

| Text | Lang | Result |
| --- | --- | --- |
| Parijs | nl | Parijs, Frankrijk (whosonfirst:locality:101751119 Paris) |
| Londre | fr | Londres, Angleterre, Royaume-Uni (whosonfirst:locality:101750367 London) |
| ブラジル | ja | ブラジル (whosonfirst:country:85633009 Brazil) |

cc @mihneadb Have you started working on it ? I can take the task if you want :smile:

mihneadb:
@Joxit I haven't, I was waiting for some clarifications in the other issue. Sure, go ahead and take it, thanks a ton for doing this! :) 🥂

Joxit:
It seems that in order to validate this issue, all the importers must support the multi-lang index.

At this time, only [OSM](https://github.com/pelias/openstreetmap/blob/039513cae54c377ced75104e53a325e96b1e602a/stream/tag_mapper.js#L39) supports it.
WOF will be supported with pelias/whosonfirst#446
Geonames needs `alternateNamesV2` file to add multi-lang (we want that ?)
OpenAddresses and Polylines are unavailable

I think, the most important importer is WOF, the city/country search is the most common use case of the geocoder.

mihneadb:
@Joxit FWIW there is some level of support for that already - when looking for something, passing e.g. `lang=en` or `lang=ru` yields the same `name` but the city name is translated.

https://pelias.github.io/compare/#/v1/autocomplete%3Flang=en&text=red%20square%20moscow
vs
https://pelias.github.io/compare/#/v1/autocomplete%3Flang=ru&text=red%20square%20moscow
(see `label`)

I thought that data was based on WOF.

Joxit:
[...]

Issue #1205, opened 2018-10-11
Possible autocomplete focus point query improvements


orangejulius:
A rough outline of our autocomplete query structure when using a focus point is as follows:

```
{
  "must": {
       // up to two queries against name/phrase fields
       // (analyzers: peliasQueryFullToken and peliasQueryPartialToken)
       // uses custom logic to query only the last word using peliasQueryPartialToken
  },
  "should": [
       {
          "function_score": {
               "query": {
                    // a different query against name field
                    // analyzer: peliasQueryPartialToken
                    // boost: 100
                },
               "functions": {
                    // `center_point` query to handle focus point, weight 15
                },
                "boost_mode": "replace"
           }
      }, {
         // other should queries for admin areas
      }]
}
```

From this structure alone, there are several clear problems:
- the query used to combine with the focus point calculations is different than the primary scoring query
- the text query for the focus point clause will match MANY documents, since it is using partial tokens
- because they are separate query clauses, the scoring for the focus point and main queries may have completely separate magnitudes

This last point is what lead me to this investigation, as I found a query where a slightly more nearby result was out-scoring a significantly better text result. Elasticsearch applied a queryNorm multiplication factor of `0.000050113387) to the main query. Even though the sub-score for the fully matching result was 13.6, while the poor match's core was 6.8, the final scores ended up being 0.013 vs 0.006. While the relative difference of the scores is over 2x, the absolute difference is tiny, meaning a slight distance difference was enough to ruin the results.

We should explore whether it's possible to move the function score calculation for distance from the focus point to the "main" query clause. This would ensure that the distance scoring is always applied as a fair multiplier to the actual text query, rather than a completely separate scoring addition.

PR #1432, opened 2020-05-12
Use match instead of match_phrase query for autocomplete

orangejulius:
This is a small but meaningful change I've been meaning to make to our autocomplete queries for a while.

The primary `must` query in many of our autocomplete queries is a `match_phrase` queries. This ends up being a little too strict, since a `match_phrase` requires that every token must match.

This PR replaces that query with a `match` query, and uses `minimum_should_match` to allow some tokens to be missing when there are 3 or more tokens. The original `match_phrase` query is moved to the `should` clause to provide an additional scoring boost when all tokens are matching and in the correct order.

While it's a small change, when it makes a difference, it's big. Queries that previously returned no results now can return very good results.

An example is [/v1/autocomplete?text=3929 saint Marks Avenue, Niagara Falls, ON, Canada](https://pelias.github.io/compare/#/v1/autocomplete?text=3929+saint+Marks+Avenue%2C+Niagara+Falls%2C+ON%2C+Canada&debug=0).
Screenshot_2020-05-11_20-44-31

Because we only _contract_ abbreviations when indexing, this query (which includes the word `saint` when looking for an address that uses the abbreviation `st` in our data) was finding no results. Now it not only returns the correct result, but several variations with typos or incorrect words now also return the correct result.

Some examples that now all return the correct result at least in the top 10:

[/v1/autocomplete?text=3929 st mark Avenue, Niagara Falls, ON, Canada](https://pelias.github.io/compare/#/v1/autocomplete?text=3929+st+mark+Avenue%2C+Niagara+Falls%2C+ON%2C+Canada&debug=0)
[/v1/autocomplete?text=3929 st foo Avenue, Niagara Falls, ON, Canada](https://pelias.github.io/compare/#/v1/autocomplete?text=3929+st+foo+Avenue%2C+Niagara+Falls%2C+ON%2C+Canada&debug=0)
[/v1/autocomplete?focus.point.lat=43.103698780971875&focus.point.lon=-79.02009080158956&text=3929 st Mark Avenue](https://pelias.github.io/compare/#/v1/autocomplete?focus.point.lat=43.103698780971875&focus.point.lon=-79.02009080158956&text=3929+st+Mark+Avenue&debug=0)

While I didn't see any major regressions comparing our autocomplete test suites for this PR, we should do another quick round of testing before merging this.

### DIFF
--- a/query/autocomplete_defaults.js
+++ b/query/autocomplete_defaults.js
@@ -20,6 +20,7 @@ module.exports = _.merge({}, peliasQuery.defaults, {
   'ngram:field': 'name.default',
   'ngram:boost': 100,
   'ngram:cutoff_frequency': 0.01,
+  'ngram:minimum_should_match': '1<-1 3<-25%',
 
   'phrase:analyzer': 'peliasQuery',
   'phrase:field': 'phrase.default',
--- a/query/autocomplete.js
+++ b/query/autocomplete.js
@@ -14,6 +14,7 @@ var views = {
   admin_multi_match_first: require('./view/admin_multi_match_first'),
   admin_multi_match_last: require('./view/admin_multi_match_last'),
   phrase_first_tokens_only:   require('./view/phrase_first_tokens_only'),
+  match_first_tokens_only:   require('./view/match_first_tokens_only'),
   boost_exact_matches:        require('./view/boost_exact_matches'),
   max_character_count_layer_filter:   require('./view/max_character_count_layer_filter'),
   focus_point_filter:         require('./view/focus_point_distance_filter')
@@ -40,7 +41,7 @@ adminFields = adminFields.concat(['add_name_to_multimatch']);
 var query = new peliasQuery.layout.FilteredBooleanQuery();
 
 // mandatory matches
-query.score( views.phrase_first_tokens_only, 'must' );
+query.score( views.match_first_tokens_only, 'must' );
 query.score( views.ngrams_last_token_only_multi( adminFields ), 'must' );
 
 // admin components
@@ -54,6 +55,7 @@ query.score( peliasQuery.view.address('cross_street') );
 query.score( peliasQuery.view.address('postcode') );
 
 // scoring boost
+query.score( views.phrase_first_tokens_only, 'should' );
 query.score( peliasQuery.view.focus( views.ngrams_strict ) );
 query.score( peliasQuery.view.popularity( peliasQuery.view.leaf.match_all ) );
 query.score( peliasQuery.view.population( peliasQuery.view.leaf.match_all ) );
--- a/query/view/match_first_tokens_only.js
+++ b/query/view/match_first_tokens_only.js
@@ -1,0 +1,28 @@
+const peliasQuery = require('pelias-query');
+
+/**
+  Phrase view which trims the 'input:name' and uses ALL BUT the last token.
+
+  eg. if the input was "100 foo str", then 'input:name' would only be '100 foo'
+  note: it is assumed that the rest of the input is matched using another view.
+**/
+
+module.exports = function( vs ){
+  const view_name = 'match_first_tokens_only';
+
+  // get a copy of the *complete* tokens produced from the input:name
+  const tokens = vs.var('input:name:tokens_complete').get();
+
+  // no valid tokens to use, fail now, don't render this view.
+  if( !tokens || tokens.length < 1 ){ return null; }
+
+  // set the 'input' variable to all but the last token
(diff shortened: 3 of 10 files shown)


orangejulius:
Here's an example of a downside for this PR:

when querying for [85 grey mountain drive, sedona, az](https://pelias.github.io/compare/#/v1/autocomplete?text=85+grey+mountain+drive%2C+sedona%2C+az&debug=1), we see several not-quite-matching results:
Screenshot_2020-06-01_09-00-44

As shown here, some of the incorrect matches even score higher than the desired result.

One way we might be able to fix this is some post-processing after queries in the API that would attempt to prune out clearly non-matching records. This could be useful in lots of different cases actually.

orangejulius:
I just realized I actually had a misspelling in my example above, which happened to both show off some positives, and some negatives, of this PR. It allowed us to return the correct result even with slight difference in the input and address record (grey vs gray).

Spelling the streetname as it is in the data, we see the more common way this would likely manifest, with lots of non-matching housenumbers coming in after the expected result:
Screenshot_2020-06-01_09-04-42

missinglink:
I noticed that queries for ordinal streets (such as `1st st`) can become noisier with this PR:

Screenshot 2020-06-04 at 13 37 29

The last 3 results are undesirable because they aren't what the user requested:
ie. `5 E 10th St` != `10 E 10th St`

orangejulius:
I continue to do bits of analysis on the impact of this PR. Most queries are noisier with this, which is expected. We'll want to do some analysis of the performance impact on the increased number of documents that will match and have to be scored. It might be minor, or there might be some new cases that are problematic.

Another minor but interesting example of negative changes from this PR: it will effectively increase the relative amount of scoring that comes from text matches, rather than popularity or population.

Here's an example of a simple [autocomplete query for Kansas City](https://pelias.github.io/compare/#/v1/autocomplete?text=kansas+city&debug=1). Before, the more popular (and populated) city of Kansas City, MO came first. As it happens, due to [a bug](https://github.com/pelias/whosonfirst/pull/511#issue-432108114), the `Kansas City, KS` record had a _slight_ boost to the text match. The boost was amplified by this PR enough to swap the results.

While this particular issue will be fixed, we should keep in mind that by adding another scoring query based on text match, we are essentially "devaluing" the focus point/population/popularity component of scoring. Over in https://github.com/pelias/api/issues/1205 we discuss the idea of moving from an "additive" scoring model (text + focus point + population + popularity) to a "multiplicative" scoring model (text * focus point * population * popularity), which might help avoid this problem someday.

orangejulius:
This PR has already somehow grown quite old!

Since then we've had the incredible #1296 come in to start using multi-match queries for autocomplete to improve support for multiple languages. I think this PR will require some reworking to support that.

What I'd actually like to do is take the pattern explored in this PR (where there is a loose `match` query in a `must`, and a strict `match_phrase` query in a `should`), and create a view for it in pelias/query, so that we can easily reuse it throughout Pelias. We might want both a `match` and `multi_match` variant.

query/autocomplete_defaults.js
```diff
@@ -20,6 +20,7 @@ module.exports = _.merge({}, peliasQuery.defaults, {
   'ngram:field': 'name.default',
   'ngram:boost': 100,
   'ngram:cutoff_frequency': 0.01,
+  'ngram:minimum_should_match': '1<-1 3<-25%',
```
orangejulius, line 23: This setting was used just because it's what we've used before. The explanation of what it does and why is contained in https://github.com/pelias/api/pull/1333

blackmad:
This change overall seems negative in my testing. It seems to REALLY want to make things into intersections.

some examples

/v1/autocomplete?lang=&size=10&focus.point.lat=37.78657&focus.point.lon=-122.42001999999998&text=Smoke Vape & Beyond
--> goes from perfect venue match to intersection

/v1/autocomplete?lang=&size=10&focus.point.lat=37.85597&focus.point.lon=-122.28915&text=Tenth Street & Grayson Street
--> goes from perfect intersection match to         "label": "East Grayson Street & Grayson Street, Mexia, TX, USA"

/v1/autocomplete?lang=&size=10&focus.point.lat=37.76429&focus.point.lon=-122.46603&text=9th+%26+Irving
--> was perfect match, now different intersection match 130km away from focus

/v1/autocomplete?lang=&size=10&focus.point.lat=37.77306&focus.point.lon=-122.4734&text=15th Avenue & Fulton Street
--> was perfect now, now does out-of-order on the street suffixes to  "15th Street & Fulton Street, Tell City, IN, USA" many KM away and out of order

Out of 500 venue & intersection queries I saw ~60 diffs, with no wins in the first 20

PR #1435, opened 2020-05-21
feat(debug): add new debugging levels

blackmad:
Hey all,

I made a change to add some more verbose debugging output to pelias.

they are:
- debug=2 or debug=elastic returns raw elastic response in debug output
- debug=3 or debug=explain returns raw elastic response with explain=true in the request for much more detailed scoring information

I am not at all wedded to these parameter names/values/setup. I am not wedded to this change in its current form. I think it will be generally useful for most users of pelias to be able to easily access this information, and I would like to figure out the best solution for everyone.

-d

### DIFF
--- a/routes/v1.js
+++ b/routes/v1.js
@@ -213,7 +213,7 @@ function addRoutes(app, peliasConfig) {
       // try 3 different query types: address search using ids, cascading fallback, pelias parser
       controllers.search(peliasConfig.api, esclient, queries.address_search_using_ids, searchWithIdsShouldExecute),
       controllers.search(peliasConfig.api, esclient, queries.search, fallbackQueryShouldExecute),
-      sanitizers.defer_to_pelias_parser(shouldDeferToPeliasParser), //run additional sanitizers needed for pelias parser
+      sanitizers.defer_to_pelias_parser(peliasConfig.api, shouldDeferToPeliasParser), //run additional sanitizers needed for pelias parser
       controllers.search(peliasConfig.api, esclient, queries.search_pelias_parser, searchPeliasParserShouldExecute),
       middleware.trimByGranularity(),
       middleware.distance('focus.point.'),
@@ -272,7 +272,7 @@ function addRoutes(app, peliasConfig) {
       middleware.sendJSON
     ]),
     reverse: createRouter([
-      sanitizers.reverse.middleware,
+      sanitizers.reverse.middleware(peliasConfig.api),
       middleware.requestLanguage,
       middleware.sizeCalculator(2),
       controllers.search(peliasConfig.api, esclient, queries.reverse, nonCoarseReverseShouldExecute),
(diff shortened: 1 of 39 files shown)


orangejulius:
Cool, we've needed something like this for a long time and we should move forward on the concept. It would save a lot of debugging time.

At a practical level, we have to consider that the Elasticsearch explain endpoint can make queries over 10x slower, so anyone exposing a public Pelias API would have to consider how to handle that. Allowing anyone to make explain queries against a shared Pelias installation could get out of hand very quickly. That's the main reason we haven't done something like this so far.

blackmad:
A lot of the changes I've been thinking about are appropriate when there's an internal instance of pelias like at radar - either for local debugging, or that's proxied by a public API. There's the other style of set up like on geocode.earth or in joxit's deploy.

Would it make sense to wrap everything (/frontend, higher debug levels ...) in an internalInstance config var? Ideally it would be nice to have it set to true, since I think the default case is wanting to be able to debug a pelias instance, and have it set to false in the docker images, geocode.earth and jawg setups?

orangejulius:
I think wrapping it in another config flag that has to be flipped to enable explain queries makes sense. It's probably too impactful of a parameter to allow by default, so the config option would ensure anyone enabling it is doing so with _some_ level of awareness.

We would actually like to support it for Geocode Earth, since we also host the compare app. It would then be up to us to build the appropriate limits into our rate limiting layer :)

routes/v1.js
```diff
@@ -313,7 +313,7 @@ function addRoutes(app, peliasConfig) {
       middleware.sendJSON
     ]),
     place: createRouter([
-      sanitizers.place.middleware,
+      sanitizers.place.middleware(peliasConfig.api),
```
orangejulius, line 316: @missinglink, what do you think is our current thinking here? We can either add a parameter to all the middleware that expects an instance of the Pelias configuration, or potentially we could add a `const config = require('pelias-confg').generate().api` line to the middleware, right?

The first is technically more efficient from a code perspective (only a single instance of the config object is generated), but the second probably lowers the conceptual overhead (there's nothing to my knowledge that would prevent the middleware from being able to gather the config on its own) and allows for a cleaner interface with no parameters needed.

@blackmad since you were just in this code, can you confirm that we could, if we wanted, put a new `require` line in the middleware logic instead of passing it around a config object everywhere?

orangejulius:
Cool, I think with the new config parameter and description this is fundamentally ready to be merged.

I'd like to briefly discuss how we pass around config objects in the code (comment made on the code in question), but it's not a big deal at the end of the day.

In terms of logistics, once we're happy with any code nitpicking, @blackmad can you:
- make sure this is rebased on top of latest master
- squash to one commit
- make the commit message title something like `feat(debug): add new debugging levels` so that Semantic release takes care of it appropriately? :checkered_flag:

blackmad:
> Cool, I think with the new config parameter and description this is fundamentally ready to be merged.
> 
> I'd like to briefly discuss how we pass around config objects in the code (comment made on the code in question), but it's not a big deal at the end of the day.
> 
> In terms of logistics, once we're happy with any code nitpicking, @blackmad can you:
> 
> * make sure this is rebased on top of latest master
> * squash to one commit
> * make the commit message title something like `feat(debug): add new debugging levels` so that Semantic release takes care of it appropriately? 🏁

all three are done

routes/v1.js
blackmad: autocomplete.js and search.js already passed this in, so this was bringing in the other two big sanitizers into same pattern

blackmad:
bump.

PR #1437, opened 2020-05-21
 use parsed_text.housenumber across pelias/api codebase

blackmad:
I think this a bug in pelias where it is sometimes calling an exact match a fallback result because it was looking for the wrong field in parsed_text.

the "Street" fallback rule was applying in some cases when it shouldn't have been. I believe this is because it was looking in parsed_text for a field called "number" which is now called "housenumber"

### DIFF
--- a/controller/libpostal.js
+++ b/controller/libpostal.js
@@ -9,7 +9,7 @@ var field_mapping = {
   island:         'island',
   category:       'category',
   house:          'query',
-  house_number:   'number',
+  house_number:   'housenumber',
   road:           'street',
   suburb:         'neighbourhood',
   city_district:  'borough',
--- a/controller/predicates/isAdminOnlyAnalysis.js
+++ b/controller/predicates/isAdminOnlyAnalysis.js
@@ -10,7 +10,7 @@ module.exports = (request, response) => {
   }
 
   // return true only if all non-admin properties of parsed_text are empty
-  const is_admin_only_analysis = ['unit', 'number', 'street', 'query', 'category', 'postalcode'].every((prop) => {
+  const is_admin_only_analysis = ['unit', 'housenumber', 'street', 'query', 'category', 'postalcode'].every((prop) => {
     return _.isEmpty(request.clean.parsed_text[prop]);
   });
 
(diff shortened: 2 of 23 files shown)


blackmad:
I put a console.log in this file to verify I'm not crazy

function checkFallbackOccurred(req, hit) {

  console.error('parsed text', req.clean.parsed_text)
  // short-circuit after finding the first fallback scenario
  const res = _.find(fallbackRules, (rule) => {

    return (
      // verify that more granular properties are not set
      notSet(req.clean.parsed_text, rule.notSet) &&
      // verify that expected property is set
      areSet(req.clean.parsed_text, rule.set) &&
      // verify that expected layer(s) was not returned
      rule.expectedLayers.indexOf(hit.layer) === -1
    );
  });

  return !!res;
}

...

parsed text {
  subject: '23 Huckleberry Hill',
  housenumber: '23',
  street: 'Huckleberry Hill',
  locality: 'Fort Mitchell',
  region: 'KY',
  postcode: '41017',
  country: 'USA',
  admin: 'Fort Mitchell, KY , USA'
}

...

but what I don't understand is how all the other places in the code where parsed_text.number is referenced are working?

orangejulius:
I think what's happening here is the Pelias parser returns this field as `housenumber`, but the output of the parse from libpostal is `number`. Here's the code that converts from the properties libostal returns to what is sent through to the rest of Pelias:

https://github.com/pelias/api/blob/0c1700af811a66bde868bdf792143293e4fb1937/controller/libpostal.js#L12

So this fallback issue will happen only when the Pelias parser is involved, but not `libpostal`.

blackmad:
blehhhh. So is the right fix here to bring housenumber from pelias/parser into parsed_text as .number, and try to clean up all the two places that reference parsed_text.housenumber?

remind me, on the geocode path, what causes it to pick pelias instead of libpostal? I was testing this on a /v1/search query that uses pelias for some reason (I thought geocode was always libpostal and autocomplete was always pelias).

orangejulius:
Yeah, I think everything for both parsers should be called `housenumber` (it's a better name than `number`, even though `number` was chosen first).

The search endpoint will always try to use the parse from libpostal first, and then generate one of several possible queries depending on what libpostal and placeholder return. If that fails, it will ask the pelias parser to parse the input, and try a final query type based on that:

https://github.com/pelias/api/blob/0c1700af811a66bde868bdf792143293e4fb1937/routes/v1.js#L210-L217

blackmad:
okay, let me try to see if I can do that surgery

PR #1438, opened 2020-05-25
Reduce search query multi_match's neighbourhood boost

orangejulius:
After the venue popularity changes in https://github.com/pelias/openstreetmap/pull/531, many venue results are improved.

However, our test for finding the [Statue of Liberty](https://github.com/pelias/acceptance-tests/blob/81e15adae7a318afa228820d519115ccc0e3271c/test_cases/landmarks.json#L16-L29), which corresponds to one of our [oldest venue popularity related issues](https://github.com/pelias/pelias/issues/171), was still showing the statue as the second result, after the ferry terminal on Liberty Island.

I traced this down to scoring boosts given to the ferry terminal because it's part of the [Liberty Island](https://spelunker.whosonfirst.org/id/907131683/) neighbourhood, whereas the statue itself is not (for some reason).

In general, I don't think neighbourhoods should make this much of a difference in queries, so this PR reduces the boost given to those matches in search queries.

This doesn't break any other tests, and while we generally don't like to adjust settings for one particular test, I think this change is worth considering.

### DIFF
--- a/query/search_defaults.js
+++ b/query/search_defaults.js
@@ -97,7 +97,7 @@ module.exports = _.merge({}, peliasQuery.defaults, {
 
   'admin:neighbourhood:analyzer': 'peliasAdmin',
   'admin:neighbourhood:field': 'parent.neighbourhood',
-  'admin:neighbourhood:boost': 1,
+  'admin:neighbourhood:boost': 0.75,
   'admin:neighbourhood:cutoff_frequency': 0.01,
 
   'popularity:field': 'popularity',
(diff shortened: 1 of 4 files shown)


orangejulius:
I looked into this more, and for the Statue of Liberty in the Pelias build I was using at least, the Statue itself is in the WOF neighbourhood [Southern Tip](https://spelunker.whosonfirst.org/id/85869681/), which is similar, but not identical to the [Liberty Island](https://spelunker.whosonfirst.org/id/907131683/) neighbourhood. Maybe we should look at having the "Southern Tip" neighbourhood removed, instead of merging this PR.

orangejulius:
The relevant neighbourhood here has been removed from WOF data after being deemed invalid.

I don't think we need this PR in general, so I'm closing it.

PR #1442, opened 2020-06-15
Run Pelias parser-based query in more cases

orangejulius:
Currently, there are 3 different types of queries against Elasticsearch that might be run from the `/v1/search` endpoint.

The last of these, which uses the Pelias Parser to parse the input, will only be run if there are _no_ results from either the previous queries, or from Placeholder (which is run first if any admin areas are recognized in the input by `libpostal`).

There are many cases where Placeholder will return results (such as a city), but the Pelias Parser could have recognized that other elements, especially _street intersections_ were present, and constructed an appropriate query to fetch them from Elasticssearch.

This PR changes the conditions under which the Pelias Parser and query based on it will run. There are two commits: one to expand the conditions for running the query at all, and a second to add new logic to detect a few exceptions to those new conditions, under which we _shouldn't_ run Pelias Parser based queries.

More details are in the two commit messages.

### DIFF
--- a/routes/v1.js
+++ b/routes/v1.js
@@ -146,10 +146,18 @@ function addRoutes(app, peliasConfig) {
     not(placeholderShouldHaveExecuted)
   );
 
-  // defer to pelias parser for analysis IF there's no response AND placeholder should not have executed
-  const shouldDeferToPeliasParser = all(
-    not(predicates.hasRequestErrors),
-    not(predicates.hasResponseData)
+  const shouldDeferToPeliasParser = any(
+    // we always want to try pelias parser based queries if there are no results
+    not(predicates.hasResponseData),
+    all(
+      // if there are only admin results, but parse contains more granular items than admin,
+      // then we want to defer to pelias parser based queries
+      predicates.hasAdminOnlyResults,
+      not(predicates.isAdminOnlyAnalysis),
+      // exception: if the 'sources' parameter is only wof, do not use pelias parser
+      // in that case Placeholder can return all the possible answers
+      not(predicates.isRequestSourcesOnlyWhosOnFirst),
+    )
   );
 
   // call search_pelias_parser query if pelias_parser was the parser
--- a/sanitizer/defer_to_pelias_parser.js
+++ b/sanitizer/defer_to_pelias_parser.js
@@ -1,4 +1,19 @@
-// middleware
+function shouldUsePeliasParse(req, res) {
+  //always use pelias parser output if old parse and queries returned no results
+  if (!res.data || res.data.length === 0) {
+    return true;
+  }
+
+  // if there's no intersection parse, the pelias parser based queries
+  // are likely not better, so don't use them
+  if (!req.clean.parsed_text.cross_street) {
+    return false;
+  }
+
+  // usually, pelias parser should be used
+  return true;
+}
+
 module.exports = (_api_pelias_config, should_execute) => {
   const sanitizeAll = require('../sanitizer/sanitizeAll'),
   sanitizers = {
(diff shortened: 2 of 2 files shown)


routes/v1.js
```diff
@@ -147,10 +147,25 @@ function addRoutes(app, peliasConfig) {
...
   // defer to pelias parser for analysis IF there's no response AND placeholder should not have executed
-  const shouldDeferToPeliasParser = all(
-    not(predicates.hasRequestErrors),
-    not(predicates.hasResponseData)
-  );
+  const shouldDeferToPeliasParser = function(req, res) {
+    if (not(predicates.hasResponseData)(req, res)) {
+      return true;
+    }
+    const ph = placeholderShouldHaveExecuted(req,res);
```
missinglink, line 154: Is there any advantage to returning here if `ph === false` to avoid doing the work below in cases where it'll never influence the result?

orangejulius: This code was written more to facilitate my own debugging than anything else. I had to split up the logic and insert lots of `console.log`s to really figure out what was happening.

One the one hand, I like the flexibility. On the other hand it probably shouldn't stay in the final version.

orangejulius: To be clear, I'm hoping that before merging this PR we can make the logic clear and clean enough that this function can be defined more similarly to the other predicate functions, where it's composed purely of other predicates without having to define a Javascript function directly.

missinglink: :+1: sorry I'll stop reviewing your draft PR 😝

orangejulius:
This is the last bit of functionality for much improved intersection handling in the `/v1/search` endpoint. #1453 and #1463 are already merged, they both also improve intersection results, but only after this PR is merged.

routes/v1.js
orangejulius: Its ready for real review now, check it out :)

routes/v1.js
```diff
@@ -147,10 +147,27 @@ function addRoutes(app, peliasConfig) {
...
+    const sources_not_only_wof = not(predicates.isRequestSourcesOnlyWhosOnFirst)(req,res);
+
+    // if there are only admin results, but parse contains more granular items than admin,
+    // then we want to defer to pelias parser based queries
+    const admin_only = all(
+      predicates.hasAdminOnlyResults,
+      not(predicates.isAdminOnlyAnalysis)
+    )(req, res);
+
+    return sources_not_only_wof && admin_only;
```
missinglink, line 169: minor nit: are there pros/cons to having `sources_not_only_wof` outside of the `all()` and then using `&&`?

I'm guessing you did it like this so you could document it better with the 👍 comments?

orangejulius: Yeah, It was a bit easier to work through the logic. We could rewrite it, perhaps something like this:

```javascript
  const shouldDeferToPeliasParser2 = any(
    // we always want to try pelias parser based queries if there are no results
    not(predicates.hasResponseData),
    all(
      // if there are only admin results, but parse contains more granular items than admin,
      // then we want to defer to pelias parser based queries
      not(predicates.isAdminOnlyAnalysis),
      predicates.hasAdminOnlyResults
      // exception: if the 'sources' parameter is only wof, do not use pelias parser
      // in that case Placeholder can return all the possible answers
      not(predicates.isRequestSourcesOnlyWhosOnFirst),
    )
  );

```

sanitizer/defer_to_pelias_parser.js
```diff
@@ -1,23 +1,47 @@
-// middleware
+function shouldUseNewParse(req, res) {
```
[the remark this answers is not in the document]
orangejulius: Good point. We've never come up with a good name for these. They're just the query layout that predates the FallbackQuery and AddressSearchWithIds query. We've previously called them the `old` query, but that's not a good name. I'll rename some of this code.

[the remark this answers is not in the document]
orangejulius: Okay, I've changed the function and comments to refer to the pelias parser rather than "new".

routes/v1.js
orangejulius: After some thought I like the standard predicate version better. So I've gone with that after testing that it does indeed behave the same.

PR #1444, opened 2020-06-16
Use peliasQuery analyzer for address_parts.street

orangejulius:
This changes the query-time analyzer for the `address_parts.street` field from `peliasStreet` to `peliasQuery`.

The latter does not generate synonyms, so it'll be much more performant and also likely produce better quality results, so win-win 🤞

### DIFF
--- a/query/autocomplete_defaults.js
+++ b/query/autocomplete_defaults.js
@@ -42,12 +42,12 @@ module.exports = _.merge({}, peliasQuery.defaults, {
   'address:housenumber:boost': 2,
   'address:housenumber:cutoff_frequency': 0.01,
 
-  'address:street:analyzer': 'peliasStreet',
+  'address:street:analyzer': 'peliasQuery',
   'address:street:field': 'address_parts.street',
   'address:street:boost': 1,
   'address:street:cutoff_frequency': 0.01,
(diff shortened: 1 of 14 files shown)


missinglink:
I added another commit to enable this for autocomplete

orangejulius:
I ran our complete automated test suite against this PR and the results are positive. There are very few changes, and they nearly all look like improvements:

For example, `15 call str` used to match `Callstraße 15, Hückelhoven, Germany`, but now it doesn't, removing one small case of autocomplete jitter:
Screenshot_2020-06-25_17-20-24

Another interesting one is that `West Wrigley Field` has now dropped out of the list of results in our still-failing Wrigley Field test:
Screenshot_2020-06-25_17-20-05

I think this PR is good to merge. We are hoping for a performance improvement here, and while we don't have any hard numbers, all the test suites did finish slightly faster (up to 10% faster) running on the same hardware and configuration.

missinglink:
Looks good to me!

orangejulius:
For the record, we tested this PR in a split test and there is no noticeable performance difference compared to the master branch. I suppose a huge performance win would have been preferable, but we'll take it!

PR #1445, opened 2020-06-16
Increase slop to 3 on `address_parts.street` field

orangejulius:
This can help allow for more flexible matches when running `/v1/search` queries.

### DIFF
--- a/query/search_defaults.js
+++ b/query/search_defaults.js
@@ -44,7 +44,7 @@ module.exports = _.merge({}, peliasQuery.defaults, {
   'address:street:analyzer': 'peliasQuery',
   'address:street:field': 'address_parts.street',
   'address:street:boost': 5,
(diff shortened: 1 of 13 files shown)


orangejulius:
After running a full test of test suites on this PR, there's very little difference with this change (as expected). We have a few internal examples of where this helps, I'll look to add some examples here soon.

orangejulius:
Okay, testing this PR has shown that it pretty much has the desired effect, and doesn't break much else.

A good example of what we want from this change is a search for [North Broad St, Canfield, OH](https://pelias.github.io/compare/#/v1/search?text=broad+st+north%2C+canfield%2C+oh&debug=1), but with the query `Broad St North`:
Screenshot_2020-06-29_08-13-48

Before this PR, we return only the city. With this PR, we can return `N Broad St` no problem :tada:

PR #1450, opened 2020-06-23
Add new debugging levels

orangejulius:
debug=2 or debug=elastic returns raw elastic response
debug=3 or debug=explain returns raw elastic response with
explain=true in the request

The new `api.exposeInternalDebugTools` configuration property must be
set to enable these, since they may expose sensitive information about a
Pelias installation, or allow end users to run expensive explain queries
that may hurt performance.

See also:
https://www.elastic.co/guide/en/elasticsearch/reference/current/search-explain.html

_This PR replaces #1435 after some rebasing and minor commit message cleanup_
Closes #1435


missinglink:
@orangejulius I noticed https://github.com/pelias/api/pull/1450/files#diff-ce94db358bac24e85d9c4cc5bbaa898aL23 removes a setting for serving the compare frontend, was that intentional? Just asking because it seems unrelated.

orangejulius:
@missinglink that was intentional.

I suppose technically that means this PR would inconvenience anyone using that flag, but I doubt there are many users since it only came in with #1416. The new parameter effectively supersedes the one it replaces.

PR #1451, opened 2020-06-24
[WIP] Reduce boost for postalcode matches

orangejulius:
For many years now, we have often seen queries where incorrect results that happened to match on a postalcode were boosted above the correct match.

Considering that there is no guarantee any particular address or venue record will have a postalcode, this case is in fact fairly common.

I believe our desired outcome is that a postalcode acts only as a 'tiebreaker' between otherwise very similar results.

With that in mind, the current default boost of 20 is _far_ too high.

### DIFF
--- a/query/search_defaults.js
+++ b/query/search_defaults.js
@@ -49,7 +49,7 @@ module.exports = _.merge({}, peliasQuery.defaults, {
 
   'address:postcode:analyzer': 'peliasZip',
   'address:postcode:field': 'address_parts.zip',
-  'address:postcode:boost': 20,
+  'address:postcode:boost': 1,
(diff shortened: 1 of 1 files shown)


orangejulius:
I just gave this PR a quick test. It does break one of our acceptance tests, but I'm not sure its a [good one](https://github.com/pelias/acceptance-tests/blob/b011dbc6cd34205b50e2a3a1852b3f1da50b14ca/test_cases/address_parsing.json#L134-L157)
Screenshot_2020-08-10_16-48-41

There were no other changes to any of our test suites that I could find, but if we ever find something that this PR helps we could merge it.

PR #1454, opened 2020-06-26
patch libpostal output for post-directionals

missinglink:
this is an extension to the work in https://github.com/pelias/api/pull/1197 which now covers post-directionals (Foo St **NE**).
we already patch pre-directionals (**NE** Foo St) so this work is to allow both positions for the directionals.

### DIFF
--- a/controller/libpostal.js
+++ b/controller/libpostal.js
@@ -155,18 +155,35 @@ function patchBuggyResponses(response){
   let idx = {};
   response.forEach((res, pos) => idx[res.label] = _.assign({ _pos: pos }, res));
 
-  // known bug where the street name is only a directional, in this case we will merge it
-  // with the subsequent element.
+  // known bug where the road name directional is parsed into an adjacent label, in
+  // this case we will merge it with the road label.
   // note: the bug only affects diagonals, not N,S,E,W
   // https://github.com/OpenTransitTools/trimet-mod-pelias/issues/20#issuecomment-417732128
-  if( response.length > 1 ){
-    let road = _.get(idx, 'road');
-    if( _.isPlainObject(road) && _.isString(road.value) && road.value.length === 2 ){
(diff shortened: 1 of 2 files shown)


missinglink:
I think this is good to merge, I was going to hit merge myself but it would be nice to have a second pair of eyes over it.
Feel free to hit merge if you're happy with it.

controller/libpostal.js
```diff
@@ -155,18 +155,35 @@ function patchBuggyResponses(response){
...
   // https://github.com/OpenTransitTools/trimet-mod-pelias/issues/20#issuecomment-417732128
-  if( response.length > 1 ){
-    let road = _.get(idx, 'road');
-    if( _.isPlainObject(road) && _.isString(road.value) && road.value.length === 2 ){
-      if( DIAGONAL_DIRECTIONALS.includes( road.value.toLowerCase() ) ){
-        let subsequentElement = response[road._pos+1];
-        if( subsequentElement && _.isString(subsequentElement.value) ){
-          response[road._pos].value += ' ' + subsequentElement.value; // merge elements
-          response.splice(road._pos+1, 1); // remove merged element
+  if (_.size(response) > 1) {
```
blackmad, line 162: overall style comments, none of this is required

- why so much use of lodash? a bunch of these seem to be things that are very native, like _.size(response) is just response.length or response?.length if you're worried about undefined. ditto _.nth - I think you only want that if your index might be negative, yes?
- idx is a weird name for an object, I get that it's an "index" but idx is often how people refer to array indexes. anything else, like tokenIndex, would be more understandable.
- would probably be more readable if all this logic were split into a function with a bunch of early-exit returns, so it's not so deeply nexted

orangejulius:
Nice. No regressions at all in our acceptance tests, but does allow us to construct more examples of where #1445 will help :)

PR #1456, opened 2020-06-29
fix [Object] in debugUrl

blackmad:
fixes https://github.com/pelias/compare/issues/33 - it was assuming hosts is always an array of strings. It seems like it can either be an array of strings or an array of objects

tested this with a hosts array that looks like

  "esclient": {
    "apiVersion": "7.5",
    "hosts": [{
      "protocol": "http",
      "host": "XXXX.ws.com", 
      "port": 9200
    }]
  },

and one that looks like

  "esclient": {
    "apiVersion": "7.5",
    "hosts": ["http://XZZZZ.amazonaws.com:9200"]
  },

### DIFF
--- a/controller/search.js
+++ b/controller/search.js
@@ -60,12 +60,19 @@ function setup( peliasConfig, esclient, query, should_execute ){
       cmd.explain = true;
     }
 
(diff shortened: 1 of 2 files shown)


orangejulius:
Thanks!

Issue #1391, opened 2019-11-19
Layers are not filtered correctly in the Search API


mihaicosareanu-bolt:
Hey team! 

I was using your awesome geocoding engine when I noticed something interesting.
Let me tell you more about it.

---
#### Here's what I did :innocent:

https://pelias.github.io/compare/#/v1/search%3Ftext=Aleea%20Ilia%2057&layers=venue,address&focus.point.lat=44.4378&focus.point.lon=26.0245

---
#### Here's what I got :scream_cat:

Notice the "layer":"street" in the properties, although in the request I filtered only "venue" and "address" layers.

```
{
  "geocoding": {
    "version": "0.2",
    "attribution": "https://geocode.earth/guidelines",
    "query": {
      "text": "Aleea Ilia 57",
      "size": 10,
      "layers": [
        "venue",
        "address"
      ],
      "private": false,
      "focus.point.lat": 44.4378,
      "focus.point.lon": 26.0245,
      "lang": {
        "name": "English",
        "iso6391": "en",
        "iso6393": "eng",
        "defaulted": false
      },
      "querySize": 20,
      "parser": "libpostal",
      "parsed_text": {
        "street": "aleea ilia",
        "number": "57"
      }
    },
    "engine": {
      "name": "Pelias",
      "author": "Mapzen",
      "version": "1.0"
    },
    "timestamp": 1574171084378
  },
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "geometry": {
        "type": "Point",
        "coordinates": [
          26.149821,
          44.444454
        ]
      },
      "properties": {
        "id": "polyline:8918983",
        "gid": "openstreetmap:street:polyline:8918983",
        "layer": "street",
        "source": "openstreetmap",
        "source_id": "polyline:8918983",
        "name": "Aleea Ilia",
        "street": "Aleea Ilia",
        "confidence": 0.8,
        "match_type": "fallback",
        "distance": 9.988,
        "accuracy": "centroid",
        "country": "Romania",
        "country_gid": "whosonfirst:country:85633745",
        "country_a": "ROU",
        "region": "Bucharest",
        "region_gid": "whosonfirst:region:85687595",
        "region_a": "BI",
        "locality": "Bucharesti",
        "locality_gid": "whosonfirst:locality:101751949",
        "neighbourhood": "Coloniști",
        "neighbourhood_gid": "whosonfirst:neighbourhood:85901737",
        "continent": "Europe",
[...]

PR #1458, opened 2020-06-30
Fixed layer filtering for the search API

mihaicosareanu-bolt:

---

- Fixes https://github.com/pelias/api/issues/1391
---

- Added a hard filter for layers on the address_search_using_ids query

---

In the issue linked above I described how to reproduce.

### DIFF
--- a/query/address_search_using_ids.js
+++ b/query/address_search_using_ids.js
@@ -17,6 +17,7 @@ addressUsingIdsQuery.filter( peliasQuery.view.boundary_circle );
 addressUsingIdsQuery.filter( peliasQuery.view.boundary_rect );
 addressUsingIdsQuery.filter( peliasQuery.view.sources );
 addressUsingIdsQuery.filter( peliasQuery.view.boundary_gid );
+addressUsingIdsQuery.filter( peliasQuery.view.layers );
 // --------------------------------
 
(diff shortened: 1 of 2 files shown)


missinglink:
Thanks for the PR, I don't recall why the layers filter was never present for this query, anyone remember?

Joxit:
Hum... Maybe for fallback ? 

For example `/v1/search?layers=address&text=20 unknown road Paris France` will return Paris or `/v1/search?layers=address&text=999999 Boulevard Saint Germain Paris France` will return  `Boulevard Saint Germain, Paris, France` ? Or it was forgotten :sweat_smile: 

But yes, this may be unexpected behaviour.

orangejulius:
I think the idea was that these queries have layer filters "built in" to them. There are sections that search for addresses and must be in the address layer, streets that must be in the street layer, etc. But the ability to limit what layers can come back at all wasn't a part of that design. It might have been an oversight, but I also think that as we added more complexity over time (the queries predate our importing of streets, for example), some bugs crept in.

orangejulius:
Anyways, this PR looks great, we will test it out real quick and hopefully merge it right away :) Thanks!

mihaicosareanu-bolt:
Wow, thanks a lot guys! You are always so responsive! :) 🚀

PR #1461, opened 2020-07-07
remove "address layer" message no one looks at

blackmad:
simple PR to remove "performance optimization: excluding 'address' layer" which we seem not to look at

### DIFF
--- a/sanitizer/_address_layer_filter.js
+++ b/sanitizer/_address_layer_filter.js
@@ -106,7 +106,6 @@ function _setup(tm) {
 
           // target all layers for the sources specified except 'address'
           clean.layers = sourceLayers.filter(item => item !== 'address'); // exclude 'address'
-          console.log('layers2', sourceLayers);
           messages.warnings.push(ADDRESS_FILTER_WARNING);
         }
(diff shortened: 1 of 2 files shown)


orangejulius:
I'm all for this, @missinglink do you think we still need this message?

missinglink:
Sorry I forgot the context, what's the motivation for removing this?

The original intention was to indicate that the default behaviour of querying on all layers was not being used for this specific query.

If it's a problem that it's unsightly in the compare app then we can deal with it there?

orangejulius:
Yeah, I also just thought about it more and it is important to message to users of the API that this is happening.

In fact it might make sense to expand the message to say something like: "performance optimization: excluding 'address' layer because the input does not contain both letters and numbers".

The underlying pain we want to solve is that we do get sick of seeing that message in the compare app.

Maybe we hack something together to skip displaying the message like Peter said? Or maybe we even get a bit fancier and hide this message, but still have some sort of indicator it was sent in the response (I could see even us Pelias developers wondering why no address results were returned for a specific query without a reminder).

PR #1462, opened 2020-07-08
fix(dedupe): consider category in deduplication

### LINKED ISSUE #1460: Deduplication issue
Hi there,

I encountered an issue concerning the dedupe strategy.
Autocomplete/search for "Manneken Pis" does not return the [little peeing guy](https://www.openstreetmap.org/node/240866019) I am expecting. Instead, only the [Bus stop](https://www.openstreetmap.org/node/5648503045) is returned.

After some debugging, I found that [they are considered duplicates](https://github.com/pelias/api/blob/master/helper/diffPlaces.js#L166-L172):
- they have the same layer and parent hierarchy
- they both have the same name in some languages
- the statue node does not have an address or postal code
- The bus stop **does** have an address and postal code
(if one element does not have an address they are considered equal)

Then, the bus stop is [preferred](https://github.com/pelias/api/blob/d0cf44f36f68472185d4d20c7b545ac214e4f160/middleware/dedupe.js#L113) because it has a zipcode.

This issue can possible happen many times as bus/metro/train stops often have the name of a nearby famous venue.

Suggestion:
Should we add a dedupe rule, maybe on category and/or addendum? venues with different categories should not be considered as duplicates.
Although, this could generate real duplicates since venues on osm are often duplicated and do not necessarily have the same tags.

Alternatively, another solution could also involve popularity. In this case, the statue has a higher popularity than the bus stop. But if you are actually looking for the bus stop, then this does not work either.

Any idea?

Thanks

https://pelias.github.io/compare/#/v1/autocomplete?focus.point.lat=50.843183&focus.point.lon=4.371755&text=Manneken+pis&debug=0

bboure:
Fixes #1460 

Open for discussion: 
Based on the other comparison methods, I considered records are the same if:
- one of them is missing the category fields, or it is empty
- at least one category match in both records

### DIFF
--- a/helper/diffPlaces.js
+++ b/helper/diffPlaces.js
@@ -39,6 +39,24 @@ function isLayerDifferent(item1, item2){
 }
 
 /**
+ * Compare the category properties if they exist.
+ * Returns false if the objects are the same, else true.
+ */
+function isCategoryDifferent(item1, item2){
+  let category1 = _.get(item1, 'category');
+  let category2 = _.get(item2, 'category');
+
+  // check if these are array objects
+  // if only one has category info, we consider them the same
+  if( !_.isArray(category1) || !_.isArray(category2) ){ return false; }
+
+  // if both categories are empty, we consider them as same
+  if( _.isEmpty(category1) || _.isEmpty(category2) ){ return false; }
+
+  return isPropertyDifferent(item1, item2, 'category');
+}
+
+/**
  * Compare the parent properties if they exist.
  * Returns false if the objects are the same, else true.
  */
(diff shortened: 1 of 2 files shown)


missinglink:
Looks good to me :+1:

Did you get a chance to test it out?
Did it fix your transit stop issue?

bboure:
> Did you get a chance to test it out?
Did it fix your transit stop issue?

Yes, that solved my issue. 
Thanks.

orangejulius:
Hey @bboure,
Thanks for the next day code delivery, very speedy :)

I think we will have to consider a record with no categories different from one with categories. For example, your example query for [Manneken pis](https://pelias.github.io/compare/#/v1/autocomplete?categories=&focus.point.lat=50.843183&focus.point.lon=4.371755&text=Manneken+pis%2C+belgium&debug=0) shows the same results with your changes (which I've put on the Geocode Earth dev server) because the statue record does not have any categories.

I believe lots of OSM records do not have categories, so there would be a lot of cases where this PR would not change much.

If you get a chance to try out that change, feel free to add it to your PR. Otherwise we may try it out as well.

bboure:
@orangejulius Sorry, I did not notice that. In my case I do have a category because I am using a custom category mapper.

I will add that change to the PR.

PR #1463, opened 2020-07-08
Upgrade to pelias-parser-1.53.0

orangejulius:
This includes all the latest intersection parsing abilities from https://github.com/pelias/parser/pull/107


orangejulius:
We did some testing on this, and it improves intersection parses considerably, with minimal regressions on other search queries :tada:

PR #1465, opened 2020-07-10
fix(debug): esclient typo / fix es debugUrl

blackmad:
verified debugUrl is working again


blackmad:
this mainly fixes the typo "eslclient" to "esclient"

and moves a ternary check into the if block it's in

PR #1466, opened 2020-07-10
feat(debug): add deep links to ES explain per result

blackmad:
adds esExplainUrl per result to better be able to understand why certain results scored the way they did without adding in all the explain output

image

supersedes https://github.com/pelias/api/pull/1465

### DIFF
--- a/controller/search.js
+++ b/controller/search.js
@@ -175,12 +175,12 @@ function setup( peliasConfig, esclient, query, should_execute ){
             if (req.clean.exposeInternalDebugTools && esHostUrl) {
               // add an ES explain link to each document
               docs.forEach((doc) => {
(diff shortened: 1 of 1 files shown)


blackmad:
not sure how useful this is but ...
image

blackmad:
not sure how useful this will be since peliasQuery is such a simple parser, but I bet there's something useful here
image

blackmad:
Added more comments. 

Here's what this change adds. I am unsure if the analyzerLinks are actually useful. I'm happy to take them out

image

blackmad:
I'm going to pull out the analyzer changes here into another proxied endpoint like the expanded-doc and restrictIds change

PR #1469, opened 2020-07-14
Remove address parts scoring in autocomplete

orangejulius:
This is an experimental PR that removes some sub-queries from our autocomplete queries. Specifically, all queries that target the `address_parts` fields (`housenumber`, `street`, `cross_street`, and `postalcode`) are removed.

While I'm not proposing we merge this PR as is, now, or possibly, ever, it's worth exploring the effects. After similar changes in the search endpoint (#1468) lead to much better intersection queries, its possible there are more benefits to be had.

I haven't fully tested this PR, but it shows improvements in some of our POI tests, particularly the query for [Wrigley Field](https://pelias.github.io/compare/#/v1/autocomplete?focus.point.lat=41.8855268583&focus.point.lon=-87.63152791895091&text=wrigley+field&debug=1)
Screenshot_2020-07-13_16-58-19

Some _potential_ advantages of the changes in this PR could be:
- Any POI that shares all or part of its name with a street (probably many of them) will be returned higher in the results
- Because there are fewer query clauses, performance might improve: each returned document will have a simpler score calculation
- Because the `address_parts` queries use the output of the Pelias Parser, queries might be more resilient against bad parses and would also have less jitter if incomplete inputs have rapidly changing parses
- Records that do not have a name match, but _do_ match on postalcode, can no longer be scored higher than records that _do_ have a name match, but _do not_ match on postalcode (for example because the postalcode is not present in the data, a common occurrence)

Some _potential_ disadvantages of this PR:
- Address-specific analysis performed on the `address_parts.housenumber` or `address_parts.street` fields in Elasticsearch would no longer apply
- No boost will be applied to records matching a given postalcode. We know there are places where the same housenumber and street name exist in a single city, with only postalcode to differentiate

I did some _very_ initial testing of some address queries against the autocomplete endpoint, and there do not appear to be any significant changes. More testing is definitely required.

### DIFF
--- a/query/autocomplete.js
+++ b/query/autocomplete.js
@@ -47,12 +47,6 @@ query.score( views.ngrams_last_token_only_multi( adminFields ), 'must' );
 query.score( views.admin_multi_match_first( adminFields ), 'must');
 query.score( views.admin_multi_match_last( adminFields ), 'must');
 
-// address components
-query.score( peliasQuery.view.address('housenumber') );
-query.score( peliasQuery.view.address('street') );
-query.score( peliasQuery.view.address('cross_street') );
-query.score( peliasQuery.view.address('postcode') );
-
 // scoring boost
 query.score( peliasQuery.view.focus( views.ngrams_strict ) );
 query.score( peliasQuery.view.popularity( peliasQuery.view.leaf.match_all ) );
--- a/test/unit/fixture/autocomplete_single_character_street.js
+++ b/test/unit/fixture/autocomplete_single_character_street.js
@@ -34,16 +34,6 @@ module.exports = {
       }],
       'should':[
         {
-          'match': {
-            'address_parts.street': {
-              'query': 'k road',
-              'cutoff_frequency': 0.01,
-              'boost': 1,
-              'analyzer': 'peliasQuery'
-            }
-          }
-        },
-        {
         'function_score': {
           'query': {
             'match_all': {}


orangejulius:
I did some more detailed testing of this last week, and the results were surprisingly good. I think we should _strongly_ consider merging this.

### Changes to address/street queries

There were actually minimal changes to address queries with this PR. One test suite we have, covering 3000 different addresses, shows the following change:
Screenshot_2020-07-28_16-11-12

That's a 0.04% difference, which by itself is very close to noise. One of the tests showed some new failures in the part of a street query with ordinals, so we might want to investigate some tests for queries like `26th st` to see if there's an issue that is more common across the board.

It does look like this PR essentially removes a small boost that was given to all address/street matches. For example, a query for the street `Avenida Juárez, Guadalajara, México`, now returns the correct result, but consistently a few spots below the first result. A couple other street queries in the same test suite (meant to test street abbreviations) also show a few new cases of failure.
Screenshot_2020-07-28_16-49-15



### Changes to POI results

The flip side to removing a boost to address/street results is, of course, an improvement in POI results. And we see many of them:
Screenshot_2020-07-28_16-21-20

Screenshot_2020-07-28_16-51-23

Even some admin results have a bit less jitter:
Screenshot_2020-07-28_16-55-12

The highly POI specific [San Diego Transit](https://github.com/pelias/fuzzy-tests/blob/master/test_cases/san_diego_transit.json) tests show a remarkable improvement:
Screenshot_2020-07-28_16-56-49

This testing predates the Wikipedia POI tests from https://github.com/pelias/fuzzy-tests/pull/71 but I'd like to re-run it with them to see the impact.


### Performance?

The results themselves are not generally overwhelming, in either direction. However I suspect that there might be a performance improvement with this change. The test runs are consistently ever so slightly faster with this PR, and while that's a very blunt measurement (and also really reflects throughput, whereas long tail latency is the main thing we want to improve with autocomplete), it is a hint that there might be improvements. That's what I plan to test next.

blackmad:
in 500 random venue-only autocomplete queries, I see 49 diffs, +15, -2, =32

the only losses are where intersections are preferred over street segments. two examples:
{ "focus.point.lat": "37.82756", "focus.point.lon": "-122.2309", "text": "Blair Avenue" }
{ "focus.point.lat": "37.79805", "focus.point.lon": "-122.42053", "text": "Green Street" }

I think it's doing this because it's getting higher TF for Blair in "Blair Place & Blair Ave" over just "Blair Ave" :-/

blackmad:
In ~500 fully qualified CA addresses, I see 20 diffs, most of which are neutral. Two losses both of which I think come from optionalizing the postcode

{ "near": "37.36182,-122.02376", "text": "625 E El Camino Real, Sunnyvale, CA 94087, US", "size": "1", "sources": "oa,osm" }
- picks a record with postal code 94086 (which is actually the wrong postcode for this record?) instead of the prior one in 94087 (same lat/lng tho)

{ "near": "37.7716,-122.43239333333332", "text": "609 Haight St, San Francisco, CA 94117, USA", "size": "1", "sources": "oa,osm" }
- picks a record that doesn't have a postcode on it, instead of one that does (same lat/lng)

Maybe we leave in postcode boosting since pulling out a postcode is rarely wrong?

That said, this is overall positive IMO and I'm overall in favor of it.

orangejulius:
Thanks for the additional analysis @blackmad, super helpful.

The issue with the `postalcode` match is a situation like this:

imagine a query like `123 Main st, anytown, USA 12345`

If there are two addresses:

`123 Main st, anytown, usa`
`234 Main st, anytown, usa 12345`

Then we will frequently return `234 main st` instead. If we do leave the postalcode boost, it should at least be an extremely small boost, so that it can really only act as a tiebreaker.

orangejulius:
I ran this PR on a split test of live traffic, and even after a few days and plenty of data, there's no discernable performance changes from this PR.

Looks like the lesson learned here is that basic Elasticsearch scoring ("does this document contain term X", etc) is extremely fast. If we want to make queries faster, we should focus on ensuring they match fewer documents and/or removing _expensive_ scoring methods, like distance from a point or geo filters.

orangejulius:
Just finished this off and it's ready to merge. I believe it's overall a good change.

In the future, we _might_ want to experiment with adding back some or all of these query clauses with _very small_ boosts.

In particular, for postalcodes, we could solve the issue of records without a postalcode appearing first. However, for the examples given so far, it looks like deduplication that handles minor abbreviation differences would be enough (we already prefer records with postal codes in deduping, but don't consider, for example, "street" and "st" to be the same). So I'd prefer to investigate that first.

PR #1471, opened 2020-07-15
bugfix and test for setting api.autocomplete.exclude_address_length=0

missinglink:
I was investigating the effect of setting the config value `api.autocomplete.exclude_address_length=0` today and discovered a bug.

The default value for this property is `undefined` which disables the feature completely.
However.. I discovered that setting it to `0` does not! Instead the value is 'clamped' to 1 (enforcing a minimum value of 1), this is incorrect.

Hopefully no-one noticed this as I would only expect users to explicitly set `api.autocomplete.exclude_address_length` to a value greater than 0 and in fact that was also what was tested.

This PR adds an additional test to cover the case where the value is explicitly set to 0 and ensures that, in this case, the result is a no-op filter.


missinglink:
I added a second commit to update the docs https://github.com/pelias/api/pull/1471/commits/8ce505b6d840c3317e70d219339bbb8fd24db103
These docs are quite hard to write because it's pretty complicated so it ends up wordy 🤷‍♂️ 

There are two major changes to the language:
- The old docs said "below the configured length" which is actually incorrect, it should say "at, or below the configured length"
- The code tests the length of `vs.var('input:name')`, which in a lot of cases probably isn't what the average developer thinks it is, namely it's the 'subject' of the query, not the entire input (eg. 'A, London') has a length of 1

Not against changing the words to be clearer if anyone can offer something better!

orangejulius:
Huh, do we actually want this filter to operate on the `subject`? Is there any way we could change it to use the full text input, if nothing else to make the docs clearer?

PR #1474, opened 2020-07-18
fix(nearby): empty categories filter returns no result

bboure:
When passing an empty `categories` parameter to the nearby endpoint, no results are returned.
The reason is that an empty array is passed as the filter:

```json
{
  "terms": {
    "category": []
  }
}
```

Expected :
All categories should be shown, and the `category` property should be returned, like in the other endpoints.

### DIFF
--- a/query/reverse.js
+++ b/query/reverse.js
@@ -92,7 +92,7 @@ function generateQuery( clean ){
   }
 
   // categories
-  if (clean.categories) {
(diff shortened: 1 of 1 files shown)


missinglink:
Thanks!

If you get a chance to add one or two simple tests for each fix like this, they will help prevent things breaking at some point in the future.

PR #1476, opened 2020-07-21
exposeInternalDebugTools: simplify code & cover additional hosts formats

missinglink:
following on from https://github.com/pelias/api/pull/1465 this PR:
- cleans up the code a bit by moving the logic responsible for generating ES host URLs to the elasticsearch client lib
- covers more edge cases, such as where hosts is a `string` or where it's an `object` but one of the props (such as `port`) is not set.

I also made a couple of minor changes:
- use a cleaner syntax for `const apiConfig =`.
- use `=== true` to test that the flags are enabled and not just 'truthy', such as when the value is the string 'false'.

### DIFF
--- a/controller/search.js
+++ b/controller/search.js
@@ -1,6 +1,4 @@
 const _ = require('lodash');
-const querystring = require('querystring');
-
 const searchService = require('../service/search');
 const logger = require('pelias-logger').get('api');
 const logging = require( '../helper/logging' );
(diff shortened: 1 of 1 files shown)


missinglink:
After thinking about this a bit more I realised we shouldn't even need to worry about checking `pelias/config` for es host values as we can just grab 'em straight from the client:

```javascript
const host = _.first(esclient.transport.connectionPool.getConnections(null, 1)).host;
```

Fixed via rebase

Issue #1308, opened 2019-06-03
Autocomplete: stricter housenumber matching


missinglink:
We currently consider `housenumber` as a `'SHOULD'` match condition for autocomplete.

https://github.com/pelias/api/blob/master/query/autocomplete.js#L29

We could probably move this to a `'MUST'` condition to improve performance and reduce noise.

I suspect the reason we have this as a SHOULD is to handle the case where the house number doesn't exist in the index, in this case we would return 0 results for a MUST query.

It's probably possible to rewrite these subqueries to be MUST and match `term OR NULL` using something like https://www.elastic.co/guide/en/elasticsearch/reference/current/null-value.html, or https://www.elastic.co/guide/en/elasticsearch/reference/2.3/query-dsl-missing-query.html

missinglink:
Looking at what we currently have in production there doesn't seem to be any significant noise issues (showing other house numbers on the same street).

The reason for this seems to be that the house number is being matched against the `name.default` field using a MUST condition, so if the token '180' isn't in the index with 'grolmanstrasse' then 0 results are returned:

```
/v1/autocomplete?debug=true&text=180 grolmanstrasse
/v1/autocomplete?debug=true&text=grolmanstrasse 180
```

...so I'm not sure why the subquery isn't already a MUST
any ideas @orangejulius ?

orangejulius:
Just came across this old issue that I don't recall from last year. I think the code is effectively meant to link to this line, which doesn't appear to have changed:

https://github.com/pelias/api/blob/e379a5f18db1bafad1b3eb4fa643f09994d048af/query/autocomplete.js#L51

I think that the housenumber and street specific query clauses are not having much of an effect, since as mentioned the full address is also included in a `must` condition. In fact we are looking to make that _less_ strict with https://github.com/pelias/api/pull/1432.

The only effect I imagine these two query clauses are having is to boost address records. As we are looking at in https://github.com/pelias/api/pull/1430 and elsewhere, we may or may not want this scoring boost.

I also bet that we could improve performance by skipping these `should` query clauses all together. Each one requires a lookup in the inverted index for another field, and additional scoring logic. Might be worth experimenting with.

missinglink:
We need to ensure that `5 E 10 st`, `5 E 5 st`, `10 E 10 st` and `10 E 5 st` score appropriately depending on the input.

Ideally we wouldn't show the incorrect versions at all, but the extra noise may be a fair compromise, so long as the ranking is preserved.

[...]

PR #1479, opened 2020-07-31
don't boost street if parse is ambiguous

blackmad:
This is a redo of https://github.com/pelias/api/pull/1430 that's on the pelias repo rather than on mine, and fixed for place->venue rename

It fixes `mccarren park` and `wrigley field` as well as `mccarren park brooklyn` and `wrigley field chicago`

this was the original approach suggested by @missinglink for solving "wrigley field" - I don't see a great way to solve it in pelias/parser at the moment, though I'm open to suggestions.

doesn't change anything in a 1k query set.

### DIFF
--- a/sanitizer/_text_pelias_parser.js
+++ b/sanitizer/_text_pelias_parser.js
@@ -53,26 +53,31 @@ function _sanitize (raw, clean) {
 }
 
 function isAmbiguousStreetOrVenueParse(solutions){
-  if (!solutions) {
+  if (!solutions || solutions.length < 2) {
     return false;
   }
 
   const topTwoSolutions = _.take(solutions, 2);
   const streetOnlySolution = topTwoSolutions.find(solution => {
     return solution.pair &&
-      solution.pair.length === 1 &&
       solution.pair[0].classification.label === 'street';
   });
   const venueOnlySolution = topTwoSolutions.find(solution => {
     return solution.pair &&
-      solution.pair.length === 1 &&
       solution.pair[0].classification.label === 'venue';
   });
+
+  const restOfSolutionIsTheSame = 
+    topTwoSolutions[0].pair.length === topTwoSolutions[1].pair.length &&
+    _.every(_.zip(topTwoSolutions[0].pair.slice(1), topTwoSolutions[1].pair.slice(1)), ([p1, p2]) => 
+       p1.span.body === p2.span.body && p1.classification.label === p2.classification.label
+    );
 
   return (
     streetOnlySolution &&
     venueOnlySolution &&
-    streetOnlySolution.pair[0].span.body === venueOnlySolution.pair[0].span.body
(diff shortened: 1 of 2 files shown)


orangejulius:
Cool, this makes a lot of sense. I'll test it out next week!

blackmad:
something else this should probably do is that if the parse is ambiguous, consume the *whole* query

it would fix queries like
Kinetic Arts Center
    "parsed_text": {
        "subject": "Arts Center",
        "street": "Arts Center"
      }

and

KT Auto Repair
    "parsed_text": {
        "subject": "Auto Repair",
        "street": "Auto Repair"
      }

orangejulius:
I just gave this PR a full run through all our tests, and it looks good. There are some improvements to POI related tests, such as the following:
Screenshot_2020-08-10_16-08-02
It looks similar, but not quite as good as #1469 in that regard.

There were _very_ minimal changes to address queries that are not worth worrying about. Here's an example of the magnitude of the changes
Screenshot_2020-08-10_16-11-19

As far as I'm concerned from a behavior perspective this PR is good to merge.

@missinglink do you have any thoughts on the code? It does add even more logic to our already complex `sanitizer/_text_pelias_parser.js` file, but that's probably unavoidable.

sanitizer/_text_pelias_parser.js
```diff
@@ -52,6 +52,36 @@ function _sanitize (raw, clean) {
...
+  const restOfSolutionIsTheSame = 
+    topTwoSolutions[0].pair.length === topTwoSolutions[1].pair.length &&
+    _.every(_.zip(topTwoSolutions[0].pair.slice(1), topTwoSolutions[1].pair.slice(1)), ([p1, p2]) => 
+       p1.span.body === p2.span.body && p1.classification.label === p2.classification.label
+    );
+
+  return (
+    streetOnlySolution &&
+    venueOnlySolution &&
+    streetOnlySolution.pair[0].span.body === venueOnlySolution.pair[0].span.body &&
```
Joxit, line 79: I think you can put this before `const restOfSolutionIsTheSame =....` ? something like
```js
if (!streetOnlySolution || !venueOnlySolution || streetOnlySolution.pair[0].span.body !== venueOnlySolution.pair[0].span.body) {
  return false
}
```

orangejulius:
We just merged #1308, so I'd be curious to see if/when this PR provides additional improvements on top of this. @blackmad if you get a chance to rebase and check it out, let me know, otherwise I'll add it to my queue for testing.

blackmad:
I think this is basically a no-op on top of 1308, I'm happy to just close it

orangejulius:
@blackmad Its a no-op for autocomplete, but not necessarily search, so it might still be useful.

blackmad:
okay true you're right. This is still important for "wrigley field" in search

PR #1480, opened 2020-08-03
Increase slop to 4 on `address_parts.street` field

stevelacy:
This is a pr for setting slop to 4.

https://github.com/pelias/pelias/issues/880#issuecomment-668149299


orangejulius:
Thanks for the quick PR turnaround :)

We'll try to run this through all our extra address tests quickly just to make sure it doesn't break anything unexpected.

orangejulius:
There were no differences at all in our test suites, so I'm pretty sure this is good to go :)

Thanks again for the PR

PR #1483, opened 2020-08-10
Update Pelias Parser to 1.56.1

orangejulius:


orangejulius:
I did some testing of this upgrade, just to stay on top of parser changes. There are actually some minor _regressions_ in address parsing, I suspect due to https://github.com/pelias/parser/pull/115? More investigation is needed. It might not matter as much after work like https://github.com/pelias/api/pull/1484

missinglink (approved):
Diff looks good to me.

Only concern against merging are the reported address regressions.
Do we have one or two examples of these?

orangejulius:
I just rebased this to keep it up to date with other PRs that have been merged. The only really notable regression is regarding an address with the `square` road type. It looks like a few characters in between `sq` and `square` are now failing.

Joxit (approved):
LGTM

orangejulius:
Replaced by #1565

PR #1485, opened 2020-08-18
Use match_all for focus point queries

orangejulius:
We handle the `focus.point` parameter by using the [function_score](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-function-score-query.html) Elasticsearch query. `function_score` requires a function, and a query.

For historical reasons, we use a `match_phrase` query based on some of the input text for this query. However, there's really no reason to do so, we want the function score for distance from the focus point to apply to _all_ results, not just the subset of results that match the `match_phrase` query.

Usually, the subset of results that has the focus point scoring applied is almost 100% of all results, but not always, which can be very confusing.

### DIFF
--- a/query/autocomplete.js
+++ b/query/autocomplete.js
@@ -48,7 +48,7 @@ query.score( views.admin_multi_match_first( adminFields ), 'must');
 query.score( views.admin_multi_match_last( adminFields ), 'must');
 
 // scoring boost
-query.score( peliasQuery.view.focus( views.ngrams_strict ) );
+query.score( peliasQuery.view.focus( peliasQuery.view.leaf.match_all ) );
 query.score( peliasQuery.view.popularity( peliasQuery.view.leaf.match_all ) );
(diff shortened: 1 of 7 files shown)


Joxit (approved):
:+1:

orangejulius:
Worth noting we already started using `match_all` quite a while ago for the other `function_score` query components (population and popularity) in https://github.com/pelias/api/pull/1356

orangejulius:
Okay, because it's pretty easy for now, I ran this through our full test suite just because I wanted to make sure we weren't missing anything.

There are in fact some extremely small changes from this PR:
Screenshot_2020-08-19_16-15-44

In both cases, a nearby result is now boosted higher than the "expected" result.

Thinking about it more, the logic before this PR was essentially "if a result is not a very good text match (all tokens in the input text present), don't give it a scoring boost for being close to the focus point".

There actually could be some validity to that sort of approach. But the way we were doing it is not the way, we would want to apply a lot more nuance and control. So this PR is good to go :)

PR #1300, opened 2019-05-20
Use default index + user-lang index in autocomplete

Joxit:
**EDITED**
The lang boos is the same as the default index.
The default lang is `en`.

I ran an acceptance test on our production server, there are no regressions :tada: 

related: pelias/api#1296
superseed: pelias/api#1298
Fixes https://github.com/pelias/pelias/issues/767

PR #1493, opened 2020-10-06
Add support to lang in `ngrams_last_token_only_multi`

Joxit:

---
@bboure found an issue in autocomplete multi-lang field, [Edo Tokyo Museum](https://pelias.github.io/compare/#/v1/autocomplete?layers=venue&debug=1&text=Edo+Tokyo+Museum) is not returned see https://github.com/pelias/api/issues/1296#issuecomment-703679827 

---

Since #1300 we are using two field for autocomplete search, `name.default` and `name.{lang}`.  `name.{lang}` was missing in the `ngrams_last_token_only_multi` query

---

[Edo Tokyo Museum](https://pelias.github.io/compare/#/v1/autocomplete?layers=venue&debug=1&text=Edo+Tokyo+Museum) should now work

### DIFF
--- a/query/autocomplete.js
+++ b/query/autocomplete.js
@@ -4,6 +4,7 @@ const defaults = require('./autocomplete_defaults');
 const textParser = require('./text_parser_pelias');
 const config = require('pelias-config').generate();
 const placeTypes = require('../helper/placeTypes');
+const toSingleField = require('./view/helper').toSingleField;
 
 // additional views (these may be merged in to pelias/query at a later date)
 var views = {
@@ -32,7 +33,7 @@ var adminFields = placeTypes.concat(['locality_a', 'region_a', 'country_a']);
 // the name of the field to use.
 // this functionality is not enabled unless the 'input:add_name_to_multimatch'
 // variable is set to a non-empty value at query-time.
-adminFields = adminFields.concat(['add_name_to_multimatch']);
+adminFields = adminFields.concat(['add_name_to_multimatch', 'add_name_lang_to_multimatch']);
 
 //------------------------------
 // autocomplete query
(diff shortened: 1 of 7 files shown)


bboure:
Awesome! Thank you @Joxit

missinglink (approved):
Looks good to me.

It would be nice to have some tests for this in https://github.com/pelias/acceptance-tests to prevent regression.

orangejulius:
Thanks for fixing this @Joxit!

I was reading along with the comments yesterday and happened to write an acceptance test to examine this issue more closely, so I've opened https://github.com/pelias/acceptance-tests/pull/535 to bring it in. Thanks @bboure for the report and test case as well!

Joxit:
Yeah, thx @orangejulius for the acceptance test :smile:

missinglink:
Heeeeey @Joxit 

I was investigating a bug report today and I noticed something on our dev server which looks incorrect:
https://pelias.github.io/compare/#/v1/autocomplete?lang=de&text=Am+Gro%C3%9Fhausberg%2C+Furtwangen+im+Schwarzwald+deut&debug=1

Dev server is running a newer `master` branch than the production server.

Screenshot 2020-10-16 at 11 51 25

The screenshot shows it's using `phrase.default` but `name.de`, is that possibly due to this PR?
I would expect to see `phrase.default` and `phrase.de`.

Joxit:
Hi @missinglink Yes you're right, I think this is due to this PR :thinking:

Maybe I should use `admin:add_name_to_multimatch:field` instead of `admin:add_name_lang_to_multimatch:field`...

missinglink:
It's not super urgent but we should fix it, could you please open an issue/pr so we don't forget?

missinglink:
Some of those query views are a hot mess, it's becoming increasingly difficult to figure out which does what, where.
I'm probably guilty of the worst of it, we need to clean all that up at some point 🤦

Joxit:
Ok, I will open an issue and work on it later (I tried something but it will be harder than I expected :cry:)
Here is where we update the `name.default` field to `phrase.default`: 

https://github.com/pelias/api/blob/5d202c48c0753f34f3401bd0c451449d53cf1e6b/query/view/admin_multi_match_first.js#L32-L37

:laughing: I agree :sweat_smile:

PR #1494, opened 2020-10-12
feat(override): allow addendum data to override imported place data

blackmad:
This is part of a workaround for the issues we've seen in pelias with scoring on records that have multiple names. The issue is that elasticsearch calculates the term frequency over all the names, so a place like ["Sutter St Cafe", "The Sutter St Cafe"] is more likely to match "Sutter St" than a record with just the name ["Sutter St"]
    
This change allows us to create multiple records, one per name, but to specify in the addendum data what the canonical name (or any other data) is, so that we can generate lots of aliases for records and not worry too much about the term frequency scoring issues, such as adding "Restaurant" to every restaurant venue in our data, or aliasing "subway" and "sub way".

It's hard to say what the impact of this is - in our pelias set up we are generating aliases outside of pelias trunk, on commercial data. Without this change, creating aliases (like turning all "AA" tokens into "A A", generating compound/decompound mappings for words like coffeeshop<->coffee shop) we were seeing some regressions when only creating one record with multiple names, due to repeated tokens. With this change in approach, we are not seeing any regressions.

I would like to get this change into pelias master (rather than applying this hack at our application server level) so that I can continue performing side-by-side evaluations at the pelias api level on new indexes and accurately seeing what we are trying to show to users.
    
I am open to other approaches to this.

### DIFF
--- a/middleware/applyOverrides.js
+++ b/middleware/applyOverrides.js
@@ -21,7 +21,7 @@ function applyOverrides(req, res, next) {
  * Rename the fields in one record
  */
 function overrideOneRecord(place) {
-  if (place.addendum && place.addendum.override) {
+  if (_.has(place, 'addendum.override')) {
     try {
       const overrideData = codec.decode(place.addendum.override);
       place = _.merge(place, overrideData);
--- a/middleware/applyOverrides.js
+++ b/middleware/applyOverrides.js
@@ -1,0 +1,28 @@
+const _ = require('lodash');
+
+function setup() {
+  return applyOverrides;
+}
+
+function applyOverrides(req, res, next) {
+  // do nothing if no result data set
+  if (!res || !res.data) {
+    return next();
+  }
+
+  res.data = res.data.map(overrideOneRecord);
+
+  next();
+}
+
+/*
+ * Rename the fields in one record
+ */
+function overrideOneRecord(place) {
+  if (place.addendum && place.addendum.override) {
+    place = _.merge(place, place.addendum.override);
+  }
+  return place;
+}
+
+module.exports = setup;
--- a/middleware/applyOverrides.js
+++ b/middleware/applyOverrides.js
@@ -1,4 +1,6 @@
 const _ = require('lodash');
+const logger = require('pelias-logger').get('api');
+const codec = require('pelias-model').codec;
 
 function setup() {
   return applyOverrides;
@@ -20,7 +22,18 @@ function applyOverrides(req, res, next) {
  */
 function overrideOneRecord(place) {
   if (place.addendum && place.addendum.override) {
-    place = _.merge(place, place.addendum.override);
+    try {
+      const overrideData = codec.decode(place.addendum.override);
+      place = _.merge(place, overrideData);
+    } catch (err) {
+      logger.error('Invalid addendum override json string:', place.addendum);
+    }
+
+    delete place.addendum.override;
+
+    if (_.isEmpty(place.addendum)) {
+      delete place.addendum;
+    }
   }
   return place;
 }
--- a/test/unit/middleware/parseBBox.js
+++ b/test/unit/middleware/parseBBox.js
@@ -4,7 +4,7 @@ const proxyquire = require('proxyquire').noCallThru();
 
 module.exports.tests = {};
 
-module.exports.tests.computeDistance = function(test, common) {
+module.exports.tests.parseBBox = function(test, common) {
   test('valid bounding_box json', function(t) {
     var res = {
       data: [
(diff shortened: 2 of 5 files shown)


middleware/applyOverrides.js
```diff
@@ -0,0 +1,41 @@
...
+      place = _.merge(place, overrideData);
+    } catch (err) {
+      logger.error('Invalid addendum override json string:', place.addendum);
+    }
+
+    delete place.addendum.override;
+
+    if (_.isEmpty(place.addendum)) {
+      delete place.addendum;
+    }
```
missinglink, line 36: is it possible to use `_.unset()` for this or does it not remove the empty parent object?

middleware/applyOverrides.js
```diff
@@ -0,0 +1,41 @@
...
+  res.data = res.data.map(overrideOneRecord);
+
+  next();
+}
+
+/*
+ * Rename the fields in one record
+ */
+function overrideOneRecord(place) {
+  if (place.addendum && place.addendum.override) {
```
missinglink, line 24: this can be simplified with `_.has()`

middleware/applyOverrides.js
```diff
@@ -0,0 +1,41 @@
...
+    try {
+      const overrideData = codec.decode(place.addendum.override);
+      place = _.merge(place, overrideData);
```
missinglink, line 27: This command is very powerful because it bypasses all the setter logic baked into `pelias/model` which ensures validity and type correctness.

It would be fairly easy to do something silly like put a string where an array was expected for vice versa which would cause a runtime error and return a HTTP 500

test/unit/middleware/parseBBox.js
```diff
@@ -4,7 +4,7 @@ const proxyquire = require('proxyquire').noCallThru();
 
 module.exports.tests = {};
 
-module.exports.tests.computeDistance = function(test, common) {
+module.exports.tests.parseBBox = function(test, common) {
```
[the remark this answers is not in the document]
blackmad: it is, but I was using this as a template for my tests and noticed it was wrong and thought it was worth fixing?

middleware/applyOverrides.js
blackmad: wouldn't remove empty parent

middleware/applyOverrides.js
blackmad: changed!

middleware/applyOverrides.js
blackmad: That's true. I'm not sure how I could easily reuse the pelias-model logic it seems like it doesn't have a concept of loading/validating an object, so I don't know how I could use it at import or serving time to guarantee correctness.

Another approach would be to add a "displayName" property to pelias-model which seems almost more fragile since there would be so many places to decide if I should use name or displayName.

Do you have any better ideas on how this could be implemented?

middleware/applyOverrides.js
```diff
@@ -0,0 +1,41 @@
+const _ = require('lodash');
+const logger = require('pelias-logger').get('api');
+const codec = require('pelias-model').codec;
+
+function setup() {
```
missinglink, line 5: can we remove this function and `module.exports = applyOverrides` instead?

test/unit/middleware/parseBBox.js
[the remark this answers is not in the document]
missinglink: agh right, it's only a test :+1:

middleware/applyOverrides.js
blackmad: we could but that doesn't fit with the style of all the other middleware that's called as a function when being added to the call chain in v1.js

ex:
      middleware.applyOverrides(),
      middleware.dedupe(),
      middleware.accuracy(),
      middleware.localNamingConventions(),
      middleware.renamePlacenames(),
      middleware.parseBBox(),
      middleware.normalizeParentIds(),

missinglink: Agh ok, fair enough.
Seems like the right thing to do to have an outer function which can optionally accept args, kinda like a constructor.
In this case it's not required but you're right, it makes them more consistent.

middleware/applyOverrides.js
missinglink: @orangejulius and I just discussed this in our call today.

It's a very powerful tool, so we're happy to merge this as an undocumented feature with the disclaimer that it can break things and errors made in the addendum will require a re-index to fix.

With that in mind I prefer the flexibility of `addendum.override` than `displayName ` as it potentially allows prototyping other features in the future without changing code.

missinglink (approved):
Looks good to me :+1:
It's super useful when required but also a no-op when `addendum.override` isn't specified.

PR #1495, opened 2020-10-15
add github actions workflow to replace travis

missinglink:
This PR is a first attempt at replacing Travis-CI with Github Actions.
We've had loads of issues over the years with Travis so we'd like to migrate over to using Actions.

The major benefit is that the configs can be edited by anyone and the reports viewed by anyone.
It's also free for open-source projects which is nice.

I've attempted to do a 'no-op' refactor of the `.travis.yml` file although I'm open to changing anything.
The long-term plan is to live with this for a while on an active repo and once we're happy with it we can roll it out across the other repos.

cc/ @pelias/contributors 

[edit] ermahgerd it's so fast!

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -1,0 +1,41 @@
+name: Continuous Integration
+on: push
+jobs:
+  unit-tests:
+    runs-on: '${{ matrix.os }}'
+    strategy:
+      matrix:
+        os:
+          - ubuntu-20.04
+          - macos-10.15
+        node-version:
+          - 10.x
+          - 12.x
+    steps:
+      - uses: actions/checkout@v2
+      - name: 'Install node.js ${{ matrix.node-version }}'
+        uses: actions/setup-node@v2-beta
+        with:
+          node-version: '${{ matrix.node-version }}'
+      - name: Run unit tests
+        run: |
+          npm install
+          npm run travis
+  npm-publish:
+    needs: unit-tests
+    if: github.ref == 'refs/heads/master' && github.event_name == 'push'
+    runs-on: ubuntu-16.04
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install node.js 12.x
+        uses: actions/setup-node@v2-beta
+        with:
+          node-version: 12.x
+      - name: Run semantic-release
+        env:
+          GH_TOKEN: ${{ secrets.GH_SEMANTIC_RELEASE_TOKEN }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: >
+          curl
+          "https://raw.githubusercontent.com/pelias/ci-tools/master/semantic-release.sh"
+          | bash -


.github/workflows/push.yml
```diff
@@ -0,0 +1,37 @@
...
+    - name: Install node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v2-beta
+      with:
+        node-version: ${{ matrix.node-version }}
+    - name: 'npm install, build, and test'
+      run: |
+        npm install -g npm
+        npm install
+        npm run travis
+  release:
```
missinglink, line 23: I haven't tested this step yet since it's only set to run on pushes to `master.
the syntax of the `curl` command might need some tweaking

.github/workflows/push.yml
```diff
@@ -0,0 +1,37 @@
...
+        node-version: [10.x, 12.x]
+    steps:
+    - uses: actions/checkout@v2
+    - name: Install node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v2-beta
+      with:
+        node-version: ${{ matrix.node-version }}
+    - name: 'npm install, build, and test'
+      run: |
+        npm install -g npm
```
missinglink, line 20: @orangejulius do you remember why this step is nessesary?

.github/workflows/push.yml
```diff
@@ -0,0 +1,37 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Install node.js 12.x
+      uses: actions/setup-node@v2-beta
+      with:
+        node-version: 12.x
+    - name: 'npm install, build, and test'
+      run: |
+        npm install -g npm
+        npm install --production
```
missinglink, line 36: Assuming `production` is correct here?

.github/workflows/push.yml
```diff
@@ -0,0 +1,37 @@
...
+        curl "https://raw.githubusercontent.com/pelias/ci-tools/master/semantic-release.sh" | bash -
```
missinglink, line 37: the syntax of the `curl` command might need some tweaking, I'm not sure about the mixing of YML, cURL, bash and a pipe 🤷‍♂️

.github/workflows/push.yml
```diff
@@ -0,0 +1,37 @@
+name: 'Continuous Integration'
+on:
+  push:
+    branches-ignore:
+      - 'v[0-9]+.[0-9]+.[0-9]+'
```
missinglink, line 5: I copied this from `.travis-yml`, do we still need this?

.github/workflows/push.yml
```diff
@@ -0,0 +1,37 @@
...
+    - name: Install node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v2-beta
+      with:
+        node-version: ${{ matrix.node-version }}
+    - name: 'unit tests'
+      run: |
+        npm install -g npm
+        npm install
+        npm run travis
+  npm-publish:
```
missinglink, line 23: I haven't tested this step yet since it's only set to run on pushes to `master`.

missinglink:
hmm... I think we might consider [doing this](https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/sharing-workflows-with-your-organization).

I think that will allow us to keep this config in a single repo and reference it from others, it also allows us to "centrally manage your secrets within an organization, and then make them available to selected repositories".

Joxit:
If you want the list of available software/cmd in github actions, here is the list : https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md

.github/workflows/push.yml
Joxit: Compatibility fix between node v0.12 and v4+ ?

orangejulius: I don't think it matters today, but at times we have wanted to make sure we are on the latest version of NPM to get features like `npx`, etc.

Node.js releases are much better at staying up to date with NPM these days, so it might be safe to remove if we really wanted to.

missinglink:
I just added an OS matrix included MacOS for laughs and it worked great, the default limits are 20 concurrent builds per account, so it actually takes 25 seconds to run all of them 🤯

.github/workflows/push.yml
orangejulius: we don't need an `npm install` step at all actually, the `semantic-release.sh` script takes care of all of it.

missinglink:
note to self and others, there is flag `npm run xxx --if-present` which only runs `npm` scripts when they are defined in `package.json`, this means we can define generic `npm run functional` or whatever and they will run only if those scripts are defined per-repo.

.github/workflows/push.yml
orangejulius: This just cuts down time running extra builds when we publish a new release. Due to the way Semantic Release works, tags will always be running the exact same code as `master` at the time they are created, so running all the tests again is a bit of a waste.

Does GitHub Actions have a parallelism limitation? If not then we could remove this if we want. The whole reason we needed it was that we were often waiting on the fairly limited parallelism of TravisCI.

.github/workflows/push.yml
missinglink: I've removed it but open to adding it back if required.

.github/workflows/push.yml
missinglink: removed

.github/workflows/push.yml
missinglink: on the free plan it's 20 concurrent builds per account

missinglink: I could go either way, leave it in, take it out... ⚖️

missinglink: There's also a `tags-ignore` option:
https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#example-ignoring-branches-and-tags

missinglink:
```diff
diff --git a/.github/workflows/push.yml b/.github/workflows/push.yml
index 6a65adf5..84a7afd4 100644
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -1,8 +1,7 @@
 name: 'Continuous Integration'
 on:
   push:
-    branches-ignore:
-      - 'v[0-9]+.[0-9]+.[0-9]+'
+    tags-ignore: ['*']
 jobs:
   unit-tests:
     runs-on: ${{ matrix.os }}
```

missinglink: agh ok, so we can just use `on: push` 😆 
by default this will only be [Triggered when code is pushed to any branch in a repository](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#example-using-a-single-event).

it seems that you have to explicitly opt-in to running workflows on tags, which is nice.

.github/workflows/push.yml
missinglink: We might want to replace this whole section with https://github.com/marketplace/actions/action-for-semantic-release

PR #1508, opened 2021-02-05
Prefer records with postal code regardless of source

orangejulius:
While testing changes related to improved street address normalization (https://github.com/pelias/openaddresses/pull/477),
I noticed a failure where a record from OA without a postal code was replacing a record from OSM with a postal codes.

It turns out our check for postal codes when deduping was not strong enough:
- in addition to preferring a candidate when only it has a postal code
- we must also explicitly prefer the existing hit when only _it_ has a postal code

Otherwise, a record from OA without a postal code that comes _after_ an OA record _with_ a postalcode (as returned from Elasticsearch), can end up replacing the record with the postalcode.

This happens because the postal code preference check did not explicitly return false (saying the existing hit is better), allowing the data source check (which prefers OA over OSM) to replace the OSM record.

This PR makes the postal code check stronger by adding that explicit `return false`.

In the future, perhaps we could consider removing the OA address over OSM address preference?

Here's the relevant example query from the acceptance tests:

https://pelias.github.io/compare/#/v1/search?text=301+Commons+Park+S%2C+Stamford%2C+CT+06902

### DIFF
--- a/middleware/dedupe.js
+++ b/middleware/dedupe.js
@@ -113,6 +113,9 @@ function isPreferred(existingHit, candidateHit) {
   // https://github.com/pelias/api/issues/872
   if( !_.has(existingHit, 'address_parts.zip') &&
        _.has(candidateHit, 'address_parts.zip') ){ return true; }
+  // if the existing hit HAS a postcode, and this candidate does NOT, keep the existing hit
+  if( _.has(existingHit, 'address_parts.zip') &&
+       !_.has(candidateHit, 'address_parts.zip') ){ return false; }
 
   // prefer non-canonical sources over canonical ones
   if( !_.includes(canonical_sources, candidateHit.source) &&
--- a/test/unit/middleware/dedupe.js
+++ b/test/unit/middleware/dedupe.js
@@ -396,6 +396,42 @@ module.exports.tests.priority = function(test, common) {
     });
   });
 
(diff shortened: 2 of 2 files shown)


missinglink (approved):

missinglink (approved):
simple change, looks good.

msschl:
This is great 🚀  
We were still having issues where no postal code was returned. See issues [pelias#887](https://github.com/pelias/pelias/issues/887) or [pelias#889](https://github.com/pelias/pelias/issues/889). Together with this fix, nearly all queries now return postal codes. 
Thanks folks ❤️👍🏽

orangejulius:
@msschl that's great to hear! In our testing this was a fairly rare situation, but we definitely don't want to be serving fewer postal codes than is possible.

Looking back at https://github.com/pelias/pelias/issues/887 maybe it was a common thing for you? I could see how its possible to have, for example, two openaddresses points for the same address, one with postalcode, and one without. The code might have previously sometimes preferred the record _without_ the postalcode, right?

msschl:
Yes, most of the results previously returned were from OA. Now mostly OSM.

orangejulius:
Awesome, that's great!

To be honest, better ability to figure out what's going on with the deduplication code is fairly high on our list of internal improvements to be made.

I honestly had no idea what was going on in your situation, and didn't have it in mind when writing this PR. But, I'm glad it worked out :)

PR #1510, opened 2021-02-10
Parameterised fuzzy autocomplete

andreibalaban:

Working with `pelias-api` I discovered that a significant source of "no results" for the `autocomplete` endpoint are typos.
A "fuzzy" autocomplete would addresses this, is supported by current elasticsearch queries and I also saw that this idea was previously explored in this PR : (https://github.com/pelias/api/pull/1268) which was my initial source of inspiration.
The changes presented here support fuzziness on the current form of the autocomplete query, which includes multi-language support and a slightly different body than at the time of cited PR.
Also, `fuzziness` and `max_expansions` are exposed as parameters at request time - when they are not given, the autocomplete queries are the same as before.

- the query parts for complete and incomplete non-admin tokens are changed from "phrase"-based "multi_match" to a "best_fields" with configurable "fuzziness" and "max_expansions" 
- an extra boost is given for the "exact matches" for more relevant results being promoted
- "fuzziness" and "max_expansions" are given at request time; a "prefix_length" with the value of 1 is also used.

 - separate unit tests (`test/unit/query/autocomplete_fuzzy.js`) and fixtures are written for the fuzzy form of the queries

### DIFF
--- a/query/autocomplete.js
+++ b/query/autocomplete.js
@@ -10,12 +10,14 @@ const toSingleField = require('./view/helper').toSingleField;
 var views = {
   custom_boosts:              require('./view/boost_sources_and_layers'),
   ngrams_strict:              require('./view/ngrams_strict'),
+  ngrams_fuzzy:               require('./view/ngrams_fuzzy'),
   ngrams_last_token_only:     require('./view/ngrams_last_token_only'),
   ngrams_last_token_only_multi: require('./view/ngrams_last_token_only_multi'),
   admin_multi_match_first: require('./view/admin_multi_match_first'),
   admin_multi_match_last: require('./view/admin_multi_match_last'),
   phrase_first_tokens_only:   require('./view/phrase_first_tokens_only'),
   boost_exact_matches:        require('./view/boost_exact_matches'),
+  boost_exact_matches_fuzzy:        require('./view/boost_exact_matches_fuzzy'),
   max_character_count_layer_filter:   require('./view/max_character_count_layer_filter'),
   focus_point_filter:         require('./view/focus_point_distance_filter')
 };
@@ -49,6 +51,7 @@ query.score( views.admin_multi_match_first( adminFields ), 'must');
 query.score( views.admin_multi_match_last( adminFields ), 'must');
 
 // scoring boost
+query.score( views.boost_exact_matches_fuzzy );
 query.score( peliasQuery.view.focus( peliasQuery.view.leaf.match_all ) );
 query.score( peliasQuery.view.popularity( peliasQuery.view.leaf.match_all ) );
 query.score( peliasQuery.view.population( peliasQuery.view.leaf.match_all ) );
@@ -173,6 +176,14 @@ function generateQuery( clean ){
   // run the address parser
   if( clean.parsed_text ){
     textParser( clean, vs );
+  }
+
+  if (clean.fuzziness) {
+    vs.var('fuzzy:fuzziness', clean.fuzziness);
+
+    if (clean.max_expansions) {
+      vs.var('fuzzy:max_expansions', clean.max_expansions);
+    }
   }
 
   // set the 'add_name_to_multimatch' variable only in the case where one
--- a/query/autocomplete_defaults.js
+++ b/query/autocomplete_defaults.js
@@ -60,8 +60,10 @@ module.exports = _.merge({}, peliasQuery.defaults, {
   // generic multi_match config
   'multi_match:type': 'cross_fields',
   'multi_match:ngrams_strict:type': 'phrase',
+  'multi_match:ngrams_fuzzy:operator': 'and',
   'multi_match:first_tokens_only:type': 'phrase',
   'multi_match:boost_exact_matches:type': 'phrase',
+  'multi_match:first_tokens_only_fuzzy:operator': 'and',
 
   // setting 'cutoff_frequency' will result in very common
   // terms such as country not scoring at all
@@ -150,5 +152,9 @@ module.exports = _.merge({}, peliasQuery.defaults, {
   'custom:boosting:boost': 5,               // multiply score by this number to increase the strength of the boost
   'custom:boosting:max_boost': 50,          // maximum boosting which can be applied (max_boost/boost = max_score)
   'custom:boosting:score_mode': 'sum',      // sum all function scores before multiplying the boost
-  'custom:boosting:boost_mode': 'multiply'  // this mode is not relevant because there is no query section
+  'custom:boosting:boost_mode': 'multiply',  // this mode is not relevant because there is no query section
+
+  'fuzzy:fuzziness': 0,
+  'fuzzy:max_expansions': 10,
+  'fuzzy:prefix_length': 1
 });
--- a/query/view/boost_exact_matches.js
+++ b/query/view/boost_exact_matches.js
@@ -28,11 +28,11 @@ module.exports = function( vs ){
 
   // set 'input' to be only the fully completed characters
   vs.var(`multi_match:${view_name}:input`).set( tokens.join(' ') );
-  vs.var(`multi_match:${view_name}:fields`).set(toMultiFields(searchDefaults['phrase:field'], vs.var('lang').get()));
+  vs.var(`multi_match:${view_name}:fields`).set(toMultiFields(vs.var('phrase:field').get(), vs.var('lang').get()));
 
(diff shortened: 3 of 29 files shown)


orangejulius:
Hi @andreibalaban,
This is really awesome, thanks for all the work you put in to get this going! I believe this is not the first nice contribution from the team at Bolt, so it's great to hear you're continuing to use Pelias.

I tested this out real quick and it does work pretty well overall, right?

However, as you might have discovered, adding `fuzziness` and `max_expansions` parameters to all the Elasticsearch queries, and giving them sensible values is the EASY part of fuzzy searching for a geocoder. There's plenty more work to do after that! Here at Geocode Earth we've explored it quite a bit already, and we'd love to do more of that work together with you, but I just wanted to be clear that this isn't one of those PRs that can get a quick test and be merged right away. You probably knew that 😄 

I'm not going to go into a ton of detail here, because we honestly don't know all the answers ourselves, but here are some considerations for merging fuzzy matching:

## API Parameters

We're generally pretty hesitant to add new parameters to the Pelias API. Once they're in, they can't be taken out, and parameters that affect performance negatively require special scrutiny. Lots of Pelias users are using it with minimal to no intermediate layer between them and end users, so to some degree the Pelias API itself needs to be "user-proof". Some things we might want to do here
- Allow configuring max/min and default values for `fuzziness` in `pelias.json`. This would give people running Pelias a lot of control, and we can imagine a lot of people would certainly want fuzziness enabled by default.
- Ideally, we wouldn't expose every internal parameter in the API. I see you've already chosen to pick a default value for `prefix_length`, which is good. We might want to do the same for `max_expansions`. `pelias.json` is a good middle ground here too

## Extra non-matching results
The _cool_ part of fuzzy matching is that it lets you return the intended result with typos. The bad part is it potentially makes you return lots of _unintended_ results even if there are no typos. Consider a query for `18 glasgow street, kelburn, wellington`. Normally it returns exactly one result, a match for that address:
image

Now it returns a bunch of other addresses on the same street
image

This is tricky to even measure: our current acceptance-test framework would still mark this as passing, since the correct result is indeed returned first. However, it's a bit of a regression from a user experience standpoint. At least a few brain cells have to be used to make sure the first result is indeed the right one 🧠.

We've talked about adding a "post-processing" step that somehow looks at all the results and prunes out options that are clearly bad, but that's a difficult problem. Another potential solution is to run an Elasticsearch query _without_ fuzziness first, and then only run another with fuzziness if there were no results returned. That has its own downsides.

## Performance and response time

Stating another obvious thing, autocomplete has to return results really fast 💨. While adding fuzzy matching capability is definitely worth some response time, we'd want to test this out quite a bit to make sure its worth it. In our experience synthetic loadtesting can help, but the real answer comes from testing live traffic at scale. That's an extra challenge for something like this that changes the responses and parameters, but it's not insurmountable. You probably have some capabilities to do this, as we certainly do.

# Next steps

So, I'm really happy to see this and push it forward over the next few months. We've been dragging our feet on it a bit, since it's going to be a challenge, but working with others who are motivated helps there 🧑‍🤝‍🧑. Feel free to follow up here with more thoughts and comments. Also we should probably set up some sort of call to figure out a path forward. hello@geocode.earth is always open :)

Thanks again!

andreibalaban:
Hello @orangejulius,

I'm really glad that our initiative for integrating `fuzziness` into the current version of `autocomplete` is well received and you are open to working with us on this.
Also, thank you for taking the time to present here valuable insight regarding this - some things are easier to tackle like you also said (your suggestions in the `API Parameters` section for example), others are more tricky. It is a challenge to do this right but we are confident that working together is the key to accomplish this and we are happy to contribute.

I wrote on 'hello@geocode.earth' to schedule a call with you guys.

Cheers !

andreibalaban:
Hello @orangejulius,

It was great meeting you last week and we really appreciated the exchange of ideas.
From what I remember, the plan was to enhance this PR with some config-related changes.
Can we help you with anything regarding this?

Thank you!

mihneadb:
@orangejulius any way we can move forward with this? Would be nice to merge the logic upstream.

orangejulius:
Hi @andreibalaban and @mihneadb,

Sorry for the delay here! I finally got around to really taking a look at the impact of this PR by setting up a [branch with some reasonable default values for the fuzzy search parameters](https://github.com/pelias/api/commit/4618ae889d71b468d32c08c1382ea65e5ea8d0cf) and running it through our acceptance test suite.

While there are a lot of cases where fuzzy autocomplete helps (obviously), I think there will be more tuning work to do before we can merge this into Pelias.

### Impact on non-typo queries

The main issue I saw is that there are a decent number of cases where the "correct" results for a query with no typos no longer comes up first. For example, take this query for Paris:
image

Pakistan coming up first is a bit of a dealbreaker here. A rough estimate from all our acceptance tests is that this happens around 3-5% of the time. By no means do we need to wait until there are _no_ regressions before merging fuzzy autocomplete, but from my non-scientific testing we aren't there yet.

### Postal codes

Postal codes are particularly impacted. Since most numeric substitutions for any valid postalcode will _also_ result in a valid postalcode, and all postal codes will generally have the same score from Elasticsearch, the odds of the "right" one coming up first as it stands now is low. In this case the expected postalcode comes up 10th, so it barely even makes it into the results at all.
image

### Next steps

I think the path forward here probably entails a bit of work: we might want to add logic to detect cases (like all numeric inputs for postalcodes), where we don't do any fuzzy matching.

We also might have to add new query clauses that give a slight boost to exact matches, to solve issues like the Paris query above. If you have any other ideas or have techniques that have worked for you, let me know.

This is an important feature so we do plan to keep working on it. It's not the main thing we're working on _today_, but we are definitely not going to stop working on it just because it's hard :)

orangejulius:
Oops, quick follow up here.

I just realized that this PR is supposed to include logic for boosting exact matches. As outlined in my examples above, that's definitely important.

I'm not 100% sure my branch to test fuzzy autocomplete is correctly enabling that new logic though. Let me know if it looks broken to you. It might also be that the new exact match logic isn't _quite_ working right, in which case by fixing it we can improve result quality quite a bit.

Also let me know if you need any help running our acceptance tests to do your own testing. Both the [regular acceptance tests](https://github.com/pelias/acceptance-tests) and some of the test suites from our [fuzzy tests](https://github.com/pelias/fuzzy-tests), in particular the [city search suite](https://github.com/pelias/fuzzy-tests/blob/master/test_cases/citySearchWof.json) are useful here.

It can be a bit of an art to interpret the acceptance test results though, so let us know if we need to lend a hand.

andreibalaban:
Hello @orangejulius,

Thank you for continuing work on this and sharing insight about the problems that you discovered.

I ran a quick test using your branch (but using our data) and I managed to reproduce the postal code issue but not the one where Pakistan appears first while looking for Paris.
The way we use Pelias kind of prevents us from experiencing the fuzzy autocomplete failures that you experienced, some differences would be that we have our own popularity model, we search only within a radius of a focus point and we only query venues and addresses. Also, we don't build data on top of a full-planet OSM base. Having the mentioned popularity model for example means that we are willing to tolerate an un-exact match appearing first if it is a more popular one.

That being said we are willing to make an effort to make this work for a more "general-purpose" use-case, hopefully in a few weeks. Maybe what needs to be done is to invest more into reformulating the "exact-match" clause (which right now I think takes only complete tokens into account, and could be improved) and parametrise somehow the balance between fuzzy+popularity driven matches and exact match oriented results (maybe with a default on exact match oriented results to suit a general purpose use).
Also, do you have an easy, shareable way that we can test the code over a "full-planet" build ? Maybe a setup that you can share ? That would be of real help when we get back to this and would ensure that we are on the same page.

Thank you!

andreibalaban:
Hello @orangejulius,

I reformulated the "exact match boost" clauses in my latest [commit](https://github.com/pelias/api/pull/1510/commits/031d9baa04ef57ba604288229f148925e966dca6)
The parts of the query from the "must" clause that were "fuzzified" are now placed in original, exact-match form in the "should" clause - so "Paris vs Pakistan" and "90210" cases should not happen on this version.
Please run a round of checks on the full-planet build and let us know how this new version works; we can also iterate on it.
I also run the `acceptance-tests` and the `fuzzy-tests` suite but on our dataset the suites had a high error rate, as we do not have access to a full planet build. Like I previously mentioned, if there would be an easy way for us to have access to the same data that you use it would speed up things.
Nevertheless, hope this iteration is an improvement - let us know :D

Thank you!

Issue #1419, opened 2020-04-07
Tests require Elasticsearch to be running if `api.targets.auto_discover` is set to true


orangejulius:
With this value set in `pelias.json`, running `npm test` will hang and print errors:

```
$ npm test

> pelias-api@0.0.0-development test /home/julian/repos/pelias/api
> npm run unit


> pelias-api@0.0.0-development unit /home/julian/repos/pelias/api
> ./bin/units


  ..................Elasticsearch ERROR: 2020-04-07T15:49:03Z
  Error: Request error, retrying
  POST http://localhost:9200/_search?request_cache=true&timeout=10s => connect ECONNREFUSED 127.0.0.1:9200
      at Log.error (/home/julian/repos/pelias/api/node_modules/elasticsearch/src/lib/log.js:239:56)
      at checkRespForFailure (/home/julian/repos/pelias/api/node_modules/elasticsearch/src/lib/transport.js:298:18)
      at HttpConnector.<anonymous> (/home/julian/repos/pelias/api/node_modules/elasticsearch/src/lib/connectors/http.js:171:7)
      at ClientRequest.wrapper (/home/julian/repos/pelias/api/node_modules/lodash/lodash.js:4929:19)
      at ClientRequest.emit (events.js:311:20)
      at Socket.socketErrorListener (_http_client.js:426:9)
      at Socket.emit (events.js:311:20)
      at emitErrorNT (internal/streams/destroy.js:92:8)
      at emitErrorAndCloseNT (internal/streams/destroy.js:60:3)
      at processTicksAndRejections (internal/process/task_queues.js:84:21)

Elasticsearch WARNING: 2020-04-07T15:49:03Z
  Unable to revive connection: http://localhost:9200/

Elasticsearch WARNING: 2020-04-07T15:49:03Z
  No living connections

{"message":"No Living connections","stack":"Error: No Living connections\n    at sendReqWithConnection (/home/julian/repos/pelias/api/node_modules/elasticsearch/src/lib/transport.js:266:15)\n    at next (/home/julian/repos/pelias/api/node_modules/elasticsearch/src/lib/connection_pool.js:243:7)\n    at processTicksAndRejections (internal/process/task_queues.js:79:11)","level":"error","label":"api:type_mapping_discovery","timestamp":"2020-04-07T15:49:03.280Z"}
```

orangejulius:
[...]

PR #1521, opened 2021-04-01
Set `auto_discover` false in all unit tests

orangejulius:
This change resolves the issue described in https://github.com/pelias/api/issues/1419, where the unit tests would try and fail to contact Elasticsearch if `api.targets.auto_discover` is set to true in `pelias.json`.

There were two things required to achieve this:

- First, expand upon the work in https://github.com/pelias/api/pull/1431 that added an empty `pelias.json` file for use in the unit tests. Its now no longer empty and includes setting `auto_discover` to false.

- Second, one unit test still ended up calling Elasticsearch. It was a test for the `TypeMapping` module itself, and it was overriding `pelias-config` with its own. Now with a test-specific `pelias.json` file, that code can simply be removed, and the unit tests can use the unmodified `pelias-config` library.

I tested these changes against https://github.com/pelias/config/pull/127 to make sure that PR won't break anything, and I believe everything is good to go.

Closes https://github.com/pelias/api/issues/1419


missinglink (approved):
LGTM :+1:

PR #1522, opened 2021-04-10
Upgrade to pelias-config-5.0.0 (set `api.targets.auto_discover` to true by default)

orangejulius:
This PR brings in version 5.0.0 of [pelias/config](https://github.com/pelias/config), which brings in a [new default value](https://github.com/pelias/config/pull/127) for the `api.targets.auto_discover` configuration property.

Setting this value to true by default (the new setting) will mean that most Pelias users have to do less work to correctly configure Pelias if they're using custom data that includes their own sources and layers.

As part of this change I've also updated the documentation in the API's README.md around this option and related configuration options. Now that the default behavior is also the recommended behavior, we can simplify things quite a bit and remove a few strongly worded warnings. :)

I took a look at most of our `pelias.json` configurations out there, such as those from [pelias/docker](https://github.com/pelias/docker/) and many, but not all of them already set `api.targets.auto_discover` to true, so this configuration change is well tested and hopefully won't cause any issues.


orangejulius:
Hi @gb-kpatel,
Thanks for bringing this to our attention!

I imagine what's happened is you don't have any records in the `postalcode` layer in your index, and the query previously used to return no results (or results from other layers, if you specified them), right?

PR #1525, opened 2021-04-27
Negative sources and layers

orangejulius:
This PR adds support for "negative sources and layers": the ability to specify that certain sources or layers should be _omitted_, rather than only specifying which sources or layers should be _included_.

**Current status:** feature complete, ready for more serious testing

## How the parameters work

At its core, this PR now allows elements in the `sources` or `layers` API parameters to be prefixed by the `-` sign. This signifies that the given source/layer will be _omitted_. This is considered a "negative" source/layer, whereas the existing type of filter is "positive".

Both negative and positive filters can be specified, leading to some interesting combinations.

For example:

`/v1/search?text=berlin&layers=-venue` will search all layers except the `venue` layer

`/v1/search?text=berlin&layers=-country,coarse` will search all coarse layers _except_ `country`. `coarse` currently expands to `"continent", "empire", "country", "dependency", "macroregion", "region", "locality", "localadmin", "macrocounty", "county", "macrohood", "borough", "neighbourhood", "microhood", "disputed", "postalcode", "ocean", "marinearea"`, so it saves quite a bit of typing. It will also continue to work if we add new coarse layers

`/v1/search?text=berlin&sources=-geonames` searches all sources except geonames

## How the code works

Under the hood, the sources and layers code now checks for elements starting with `-`, and uses that to apply two filters: positive and negative, to the list of all sources/layers.

The positive list is applied _first_, and then the negative list is applied second. Because of this, queries like `/v1/search?text=berlin&layers=locality,localadmin,-locality` work fine. This one would query only for `localadmin`s.

 It _is_ possible to specify combinations that would result in nothing such as `sources=osm,-osm` or `layers=-coarse,locality`. In this case an error message is emitted and no query is performed.

### Keeping track of positive layers

There's some additional complexity that is required now that there is the internal concept of both positive and negative layers.

In particular, we have some performance optimizations that do things like remove results from the `address` layer under certain circumstances. In general, these bits of code abort if any `layers` were specified. However, we want them to function if only negative layers are specified. The internal `clean` object now stores  `positive_layers`/`positive_sources` properties that are passed around internally. These properties allow performance optimizing filters to do what they need to do.

## Work to be done
- [ ] Documentation
- [x] Additional unit tests (especially for the `sources` property)
- [ ] Additional testing for edge cases
- [x] Validation against acceptance tests
- [ ] New acceptance tests for some negative source and layer cases.
- [x] Ensure all `address` layer performance optimizations are compatible with this change
- [x] Test with various combinations of custom sources/layers
- [x] Performance testing
- [ ] ...

Connects https://github.com/pelias/pelias/issues/844

### DIFF
--- a/sanitizer/_targets.js
+++ b/sanitizer/_targets.js
@@ -1,7 +1,17 @@
 const _ = require('lodash');
 
+const INVALID_NEGATIVE_TARGETS_ERROR = 'contains positive and negative options in a combination ' +
+ 'that results in an empty list. Please chose a different combination';
+
 function getValidKeys(mapping) {
   return _.uniq(Object.keys(mapping)).join(',');
+}
+
+function expandAliases(targets, targetMap) {
+  const expanded = targets.reduce((acc, target) => acc.concat(targetMap[target]), []);
+
+  // return deduplicated list using `Set`
+  return [...new Set(expanded)];
 }
 
 function _setup( paramName, targetMap ) {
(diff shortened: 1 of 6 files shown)


orangejulius:
I've done another round of editing and testing on this PR, and I think it's ready for wider review and testing at this point.

There are no changes in the acceptance tests with this PR, and I now think the code is in a pretty good state.

I've included a bunch of cleanup in the code as well, bringing in some nice new ES6 features for a breath of fresh air. It does make the diff a bit larger though.

Looking for help from others to find edge cases or things that don't work well. Hopefully this actually isn't too far from being merged.

sanitizer/_targets.js
```diff
@@ -1,9 +1,19 @@
 const _ = require('lodash');
 
+const INVALID_NEGATIVE_TARGETS_ERROR = 'contains positive and negative options in a combination ' +
+ 'that results in an empty list. Please chose a different combination';
```
missinglink, line 4: minor typo "chose"

sanitizer/_targets.js
```diff
@@ -15,52 +25,68 @@ function _setup( paramName, targetMap ) {
...
+      // return error if parameter ends up being all whitespace
+      if( _.isEmpty( targetsString ) ){
         messages.errors.push(
           opts.paramName + ' parameter cannot be an empty string. Valid options: ' + getValidKeys(opts.targetMap)
         );
+        return messages;
+      }
+
+      // split string in to array and lowercase each target string
+      var targets = targetsString.split(',').map( function( target ){
```
missinglink, line 48: style: use fat arrow function?

missinglink (approved):
Looks good.

It's a hard one to review, I went over it twice but I'm not confident I was able to consider all the permutations.

I think the key thing here is we have as many unit tests and acceptance-tests as we can possibly think of right now and continue to add them as we find new interesting combinations.

sanitizer/_targets.js
orangejulius: Oh yeah, definitely. Makes this all fit on one line :)

sanitizer/_targets.js
orangejulius: Fixed, thanks!

missinglink:
Testing Notes:

- The `query.positive_layers`, `query.negative_layers`, `query.positive_sources`, `query.negative_sources` arrays are quite verbose in the response geojson, we should consider hiding them from the user ([much like we do with the internal bookkeeping for tokenizer logic](https://github.com/pelias/api/blob/master/middleware/geocodeJSON.js#L63-L66)).

That's it 🤷‍♂️ works great :shipit:

orangejulius:
Oh, great idea. I forgot we can easily hide properties like that. We should definitely remove them.

Joxit (approved):
LGTM

missinglink:
We still need to add some user documentation, maybe we can split that to another issue?

PR #1528, opened 2021-05-04
Use Github Actions to build Docker Image

orangejulius:
This uses the work from https://github.com/pelias/ci-tools/pull/5 to allow building the Pelias API Docker images in Github Actions.

It might be nice for us to consolidate on a single CI provider, and this would allow a smooth transition to doing so.

A nice side effect of how Docker generates immutable images is that we don't _have_ to disable the CircleCI build right away, so for now I've chosen not to remove the `.circleci` directory in this PR. However we might want want to do that soon after we're confident in using GitHub Actions.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -21,6 +21,18 @@ jobs:
         run: |
           npm install
           npm run travis
+  build-docker-images:
+    needs: unit-tests
+    if: github.event_name == 'push'
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build Docker images
+        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/push.yml
```diff
@@ -21,6 +21,18 @@ jobs:
         run: |
           npm install
           npm run travis
+  build-docker-images:
+    needs: unit-tests
+    if: github.event_name == 'push'
```
missinglink, line 26: this line might not be required as I think it does the same thing as `on: push` on line 2, although it wouldn't hurt 😄

missinglink (approved):
nice!

PR #1529, opened 2021-05-16
Improve performance of `max_character_count_layer_filter`

orangejulius:
This is a test of the performance impact of rewriting `max_character_count_layer_filter` so that it shares a single `terms`  query with the one from the `layers` parameter, rather than using its own.

After some A/B testing, it seems we can get 7% lower `p99` latency for 1 character autocomplete requests with this change. This code is not ready to be merged as is, but it was sufficient for testing out the premise before investing in the proper code changes to make it happen.

Background
==========

The `max_character_count_layer_filter` is used to exclude results from the address layer when an autocomplete query consists of less than a configured number of characters (usually set to 1, 2, or 3). This helps performance for queries that otherwise would match many millions of documents since the input is so short.

It achieves this by adding an additional `[terms](https://www.elastic.co/guide/en/elasticsearch/reference/6.8//query-dsl-terms-query.html)` query to the generated Elasticsearch query. This query enumerates _all layers_ except the address layer and tells Elasticsearch to require documents to match those layers. On its own, this is the most performant way to exclude a single layer (Elasticsearch does not support fast `not`-style queries).

However, because there is often _another_ `terms` filter that limits results to certain layers (either from the `layers` API parameter or the `address_layer_filter` sanitizer), its often the case that some of these `terms` filters are redundant.

Here's an example of the two `terms` queries created for [/v1/autocomplete?layers=locality&text=be](https://pelias.github.io/compare/#/v1/autocomplete?layers=locality&text=be&debug=1):

```json
"filter": [
  {
    "terms": {
      "layer": [
        "venue",
        "street",
        "country",
        "macroregion",
        "region",
        "county",
        "localadmin",
        "locality",
        "borough",
        "neighbourhood",
        "continent",
        "empire",
        "dependency",
        "macrocounty",
        "macrohood",
        "microhood",
        "disputed",
        "postalcode",
        "ocean",
        "marinearea"
      ]   
    }   
  },  
  {
    "terms": {
      "layer": [
        "locality"
      ]   
    }   
  }
]
```

The first `terms` query is generated by the `max_character_count_layer_filter` and  includes _all_ layers except address, while the second contains only `locality`. It should be clear that in this case the first `terms` query isn't needed at all.

Impact
================

With the changes in this PR, where only a single `terms` query is ever generated, I've seen the following reduction in p99 latencies for autocomplete queries of the following lengths:

|query length| p99 latency reduction|
| --- | ---|
| 1 | 7% |
| 2 | 3.5% |
| 3 | 3.5% |

It's also nice to shorten our Elasticsearch queries a bit, and reduce how difficult it is to read them.

How this change works
=====================

This change is not ready to merge as is: since the code previously generated a new query clause, it's written as a query view and lives in the `query` directory with our other views. It now only modifies the query variables, so it feels out of place given the new functionality.

This is probably not how we want to manage things at the end of the day, but it was a quick way to test things out.

In the long run, we probably want to move the logic in `query/view/max_character_count_layer_filter.js` to a sanitizer, so that it behaves similarly to the address layer filter in `sanitizer/_address_layer_filter.js`.

If everyone likes the idea of this change, I'll go ahead and implement it properly, since I've been deep in all of this code lately for [negative sources and layers](https://github.com/pelias/api/pull/1525).

### DIFF
--- a/query/view/max_character_count_layer_filter.js
+++ b/query/view/max_character_count_layer_filter.js
@@ -3,8 +3,12 @@ const peliasQuery = require('pelias-query');
 const allLayers = require('../../helper/type_mapping').layers;
 
 /**
-  Layer terms filter view which counts the length of 'input:name' and only
-  applies the filter condition if the text is shorter than or equal to $maxCharCount.
+  This 'filter' does not actually create a new filter query, but it modifies the 'layers' query variable
+  so that the default 'layers' filter will be changed.
+
+  It does this if the length of'input:name'is shorter than or equal to $maxCharCount.
+
+  In that case, the provided list of excluded layers is removed from the `layers` query variable.
 
   You must provide a list of $excludedLayers, all layers listed in the type mapping
   will be targeted, minus any listed in $excludedLayers.
@@ -33,8 +37,7 @@ module.exports = function( excludedLayers, maxCharCount ) {
     return () => null;
   }
 
-  // create a new VariableStore with only the layers property
-  var vsWithOnlyIncludedLayers = new peliasQuery.Vars({ 'layers': includedLayers });
+  // remove the excludedLayers from the `layers` variable
 
   // ensure char count is within a reasonable range
   maxCharCount = _.clamp(maxCharCount, MIN_CHAR_COUNT, MAX_CHAR_COUNT);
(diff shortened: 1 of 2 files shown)


orangejulius:
I've updated this PR so that it's ready to be merged.

Ideally, I would have rewritten it so that it's no longer a file in the `query/view` directory, since it no longer adds a new view to the query. However, that seems like a lot of work rearranging code for a change that won't make any difference. I've updated some info on this PR, the commit message, and most importantly the code comments, to reflect that we might want to rewrite it someday.

In the meantime, this is a nice performance improvement! 🎉

PR #1532, opened 2021-06-03
Update CI Node.js versions and operating systems

orangejulius:
This makes 3 changes to our GitHub actions CI runs.

## Remove Mac OS from build matrix

When we started using GitHub Actions it seemed like the ability to test on a Mac was nice, but there are a couple downsides:

First, they often seem to take a lot longer to run than the Linux builds. For example, see [this run](https://github.com/pelias/api/actions/runs/900305928) that took 1.5 mins on Mac, but 20 seconds on Linux.
image

Second, builds on Mac OS cost [_10x_ per minute](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions) what a Linux build does. While this isn't directly a problem for the Pelias project, since open-source CI build minutes are _currently_ free, this could bite anyone who wants to run Pelias code in their own private repo.

Personally I don't think we get enough value out of explicitly testing on a Mac to slow things down like this, but I'm open to discussion.

## Drop Node.js 10 from build matrix

Node.js 10 reached end of life on [April 30th, 2021](https://twitter.com/nodejs/status/1388116425361874945), so we can stop testing with it.

## Add Node.js 14 and 16

These new Node.js versions seem to work fine with all Pelias code, but we've been a bit slow to add them to CI. We might as well start here :)

Eventually we can/should roll out a change like this to more Pelias repositories, but the API is a logical place to start.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -9,6 +9,8 @@ jobs:
           - ubuntu-20.04
         node-version:
           - 12.x
+          - 14.x
+          - 16.x
     steps:
       - uses: actions/checkout@v2
       - name: 'Install node.js ${{ matrix.node-version }}'
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -8,7 +8,6 @@ jobs:
         os:
           - ubuntu-20.04
         node-version:
-          - 10.x
           - 12.x
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


missinglink (approved):
Fine with me, it would be nice if the Mac builds were faster, it might make sense to leave them enabled for some other repos (such as those which are compiled) but it doesn't really need to be enabled here.

missinglink:
An argument could be made that neither the potential cost or the additional seconds of build time are much of an impediment, but I'm fine to drop them if people find those things annoying 👍

orangejulius:
Yeah, I think if we reduce the number of Node.js versions, it's also not as big of a deal to run Mac builds. They also might not be as burdensome on other repositories that we change much less frequently.

Issue #1535, opened 2021-06-18
Emoji plus admin area cause 500 error


orangejulius:
Here's a "fun one":

Queries that combine an emoji and an admin area name result in a 500 error.

Here's an example:
[/v1/autocomplete?text=new york, ❤️usa](https://pelias.github.io/compare/#/v1/autocomplete?text=new+york%2C+%E2%9D%A4%EF%B8%8Fusa)

Here's the error log:

```
Error: invalid value, value must be valid js Variable
    at Variable.set (/Users/julian/repos/pelias/api/node_modules/pelias-query/lib/Variable.js:25:11)
    at /Users/julian/repos/pelias/api/query/view/admin_multi_match_first.js:41:36
    at Array.forEach (<anonymous>)
    at /Users/julian/repos/pelias/api/query/view/admin_multi_match_first.js:39:17
    at /Users/julian/repos/pelias/api/node_modules/pelias-query/layout/FilteredBooleanQuery.js:30:22
    at Array.forEach (<anonymous>)
    at Layout.render (/Users/julian/repos/pelias/api/node_modules/pelias-query/layout/FilteredBooleanQuery.js:28:17)
    at generateQuery (/Users/julian/repos/pelias/api/query/autocomplete.js:196:17)
    at controller (/Users/julian/repos/pelias/api/controller/search.js:30:27)
    at Layer.handle [as handle_request] (/Users/julian/repos/pelias/api/node_modules/express/lib/router/layer.js:95:5)
```

While the error itself is generated in [one of our query views](https://github.com/pelias/api/blob/33ab9d0310f9426612b66e03d025f3b15d40b76b/query/view/admin_multi_match_first.js#L41), I believe the ultimate error is in [sanitizer/_text_pelias_parser.js](https://github.com/pelias/api/blob/master/sanitizer/_text_pelias_parser.js) or somewhere earlier in processing the request. 

The line linked above ends up operating on a list of tokens that looks like this: `[ '', '️usa' ]`
Ideally we'd ensure that these empty tokens are never generated.

While these inputs result in a 500 error, they don't crash the Pelias API. So this isn't the worst bug, but we prefer to fix any case where the Pelias API can generate a 500 error.

missinglink:
So weird, are we missing a Unicode range here?
https://github.com/pelias/api/blob/master/helper/unicode.js#L73

PR #1536, opened 2021-06-19
Ignore empty tokens in query view

orangejulius:
Empty tokens should not make it this far into the query code, but it can happen.

In that case, we should ignore empty tokens and avoid aggravating the `VariableStore` checks by attempting to store empty strings.

This fixes the 500 error described in https://github.com/pelias/api/issues/1535, but doesn't really fix the underlying problem.

From my investigation it looks like Emoji that include a [variation selector](https://emojipedia.org/variation-selector-16/) will not be completely removed by the Pelias Parser (the emoji codepoint will be removed, but not the variation selector). If possible we should address that in some follow up work.


orangejulius:
This seems to help, although we should also look at other ways to fix the underlying issue, like #1537

PR #1538, opened 2021-06-23
Run Docker image build _after_ release

orangejulius:
As discussed in https://github.com/pelias/ci-tools/pull/7, we need to build the Docker images _after_ we publish any new versions, so that the git tag for the version is present.

This lets us publish a Docker image corresponding to the tag.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -21,18 +21,6 @@ jobs:
         run: |
           npm install
           npm run travis
-  build-docker-images:
-    needs: unit-tests
-    if: github.event_name == 'push'
-    runs-on: ubuntu-20.04
-    steps:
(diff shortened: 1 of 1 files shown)


orangejulius:
Okay, as I suspected, simply making the docker images job depend on the `npm-publish` job means that it just doesn't run except on the master branch. Time to dive into the Github Action docs :)

orangejulius:
Finally got this working, there's some fairly cryptic behavior from GitHub actions that has to be worked around: the `if` check is not the only thing that determines if a job will run, unless it includes one of the [job status check functions](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#job-status-check-functions)

Issue #1544, opened 2021-07-20
autocomplete: avoid trimming trailing spaces in sanitizer


missinglink:
for the autocomplete endpoint specifically, the presence of a trailing whitespace character in the input text has semantic value.

it indicates a word-boundary and that the final word which has been typed was completed and does not potentially represent a prefix.

looking [at the sanitiser code](https://github.com/pelias/api/blame/master/sanitizer/_text.js#L18), we trim whitespace from both sizes of the input before passing down to the `tokenizer` and `parser` functions which could use that information to make better decisions about what sort of queries to generate.

for example, the following queries should return different results, but currently return the same thing:

```
/v1/autocomplete?layers=country&text=uk

... should return

Ukraine
United Kingdom
```

```
/v1/autocomplete?layers=country&text=uk%20

... should *only* return

United Kingdom
```

PR #1545, opened 2021-07-20
preserve trailing word boundaries

missinglink:
this PR resolves the bug described in https://github.com/pelias/api/issues/1544

### DIFF
--- a/sanitizer/_text.js
+++ b/sanitizer/_text.js
@@ -17,6 +17,13 @@ function _sanitize( raw, clean ){
   // remove superfluous whitespace and quotes
   text = _.trim(_.trim(raw.text), QUOTES);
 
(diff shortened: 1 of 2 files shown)


missinglink:
this issue appears to be more complex than I had hoped, it would take some more time to investigate:
- whether `clean.text` is producing the correct value
- whether `_sanitizer/_tokinizer.js` is marking all tokens as 'complete' when the text has extra chars at the end
- whether the parser is receiving what it needs and returning the correct thing

missinglink:
note: there is already some code which is fairly similar looking but subtly different from the changes in this PR:
https://github.com/pelias/api/blob/e6179c5734d76fedc130a3905da521303c675b6f/sanitizer/_tokenizer.js#L41-L45

Issue #1520, opened 2021-03-30
Flip street name and number for Ukrainian addresses


ShoSashko:
Similar to #1343 

[Here](https://github.com/pelias/api/blob/master/middleware/localNamingConventions.js#L5) I can't see Ukraine in list
Please add UKR to a list of countries with supporting following street format: {{address}} {{number}}

missinglink:
Hi @ShoSashko
Can you please provide 3 famous/popular example addresses for our test suite?

ShoSashko:
Hello @missinglink  
Sure:
1. Svobody Ave, 28, L’viv, Lviv Oblast, Ukraine, 
2. Rynok Square, 1, L’viv, Lviv Oblast, Ukraine,
3. Stavropihiiska Street, 3, L’viv, Lviv Oblast, Ukraine

Let me know if that works for you

ShoSashko:
@missinglink Hi, any updates here?

loadit1:
Hi @missinglink ! It should be the same for  Russia, Uzbekistan, Tajikistan, Belarus, Azerbaijan, Georgia, Kyrgyzstan and Kazakhstan too.

missinglink:
The original issue was resolved by https://github.com/pelias/api/pull/1523

PR #1549, opened 2021-08-10
local-naming-conventions: add additional country codes

missinglink:
as per https://github.com/pelias/api/issues/1520#issuecomment-825126774 this PR adds the country codes from the comment on that issue.


orangejulius (approved):
Looks good!

I wonder how far away we are from really having the full list of countries with street-first addresses 🤔

PR #1552, opened 2021-08-12
autocomplete: add macrocounty to list of potential admin matching fields

missinglink:
The query "2 Macquarie Street, Sydney" does not currently return [either of these two results](https://pelias.github.io/compare/#/v1/autocomplete?text=2+Macquarie+Street+Parramatta&debug=1) despite [Parramatta](https://en.wikipedia.org/wiki/Parramatta) being a suburb within the greater Metro Sydney Area.

The reason for this is that the [Metro Area is rather large](https://en.wikipedia.org/wiki/Regions_of_Sydney) and so WOF has designated it as a 'macrocounty'.

We're currently not targeting `macrocounty` in the list of fields which are queried for the 'admin portion' of an autocomplete query:

Screenshot 2021-08-12 at 11 04 39

With this PR we add `macrocounty` to the existing list:

Screenshot 2021-08-12 at 10 59 36

The [macrocounty placetype](https://spelunker.whosonfirst.org/placetypes/macrocounty/) is seldom used, it only has 477 places at time of writing, so this likely won't have any adverse effects.

@Joxit I noticed on that link above that there's a bunch in France, would this change be beneficial/detrimental for you?

- https://pelias.github.io/compare/#/v1/autocomplete?text=2+Macquarie+Street+Sydney&debug=1
- https://pelias.github.io/compare/#/v1/autocomplete?text=2+Macquarie+Street+Parramatta&debug=1

### DIFF
--- a/query/autocomplete_defaults.js
+++ b/query/autocomplete_defaults.js
@@ -106,6 +106,11 @@ module.exports = _.merge({}, peliasQuery.defaults, {
   'admin:county:boost': 1,
   'admin:county:cutoff_frequency': 0.01,
 
+  'admin:macrocounty:analyzer': 'peliasAdmin',
+  'admin:macrocounty:field': 'parent.macrocounty.ngram',
+  'admin:macrocounty:boost': 1,
+  'admin:macrocounty:cutoff_frequency': 0.01,
+
   'admin:localadmin:analyzer': 'peliasAdmin',
   'admin:localadmin:field': 'parent.localadmin.ngram',
   'admin:localadmin:boost': 1,
--- a/test/unit/fixture/autocomplete_linguistic_multiple_tokens.js
+++ b/test/unit/fixture/autocomplete_linguistic_multiple_tokens.js
@@ -20,6 +20,7 @@ module.exports = {
                 'parent.dependency.ngram^1',
                 'parent.macroregion.ngram^1',
                 'parent.region.ngram^1',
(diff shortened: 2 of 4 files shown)


Joxit (approved):

missinglink:
Hey @Joxit I'm feeling less confident with merging this now, I like the improvements in Australia but it feels like it might come with a regression in France/Germany and any other countries which use `macrocounty`.

I see you approved the PR, does that mean you think it's still worth merging despite that?

Joxit:
Hi @missinglink, results for France reminded me this draft pelias/wof-admin-lookup#300

IMO, if France were the only country affected and if it improves the results in Australia, I think it may be worth it. 
Since we are talking about autocomplete, I think that returning results from nearby cities might not be a bad idea either :thinking: (even if in my examples there are too many results :sweat_smile:)

missinglink:
🤔

missinglink:
This PR stalled due to it increasing noise in France while also fixing issues in Australia.

It seems that the issues in Australia have worsened now since some WOF refactoring and many addresses are no longer retrievable in `Melbourne` (for instance) because it only appears as a `macroregion` in the document parents.

I feel that the omission of `macroregion` from the query list was due to oversight (or it being introduced later) rather than intent, we should be listing all the parent fields in the query.

For that reason I'd like to imminently merge this PR.

orangejulius (approved):

PR #1553, opened 2021-08-16
debug: clone debug params to ensure they are not mutated before being rendered

missinglink:
When debugging a query today I noticed that the output of `geocoding.debug[x]["controller:search"]["ES_response"].docs` doesn't accurately reflect what is returned from elasticsearch.

It turns out that the objects are being mutated in subsequent steps and the output is deferred until after that, so what we're seeing is not the state of the object at ES response time so much as the state of the objects at the end of the request lifecycle.

This PR simply makes a deep copy of the objects so we render the debug as it was at that moment in time.

The code is wrapped in `if( req.clean.enableElasticDebug )` so I'm not concerned about the perf impact.

### DIFF
--- a/controller/search.js
+++ b/controller/search.js
@@ -152,7 +152,7 @@ function setup( peliasConfig, esclient, query, should_execute ){
         }
         logger.debug('[ES response]', docs);
         if (req.clean.enableElasticDebug) {
-          debugLog.push(req, {ES_response: {docs, meta, data}});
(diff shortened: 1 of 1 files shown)


missinglink:
This does in fact return very different `ES_response` data:

```bash
.rw-r--r-- 51k peter 16 Aug 13:23 -N  test/unit/fixture/berlin_response.json
.rw-r--r-- 39k peter 16 Aug 16:06 -N  test/unit/fixture/berlin_response_new.json
```

PR #1554, opened 2021-08-17
use parent field values from same layer when deduplicating

missinglink:
In some cases the parent hierarchy contains tokens which are relevant for deduplication.

For example if we have a `geonames` record on `layer=locality` with the `name.default="Land Berlin"` and also `parent.locality=["Berlin"]`.

In this case we can use the parent field provided by the PIP service to assist in the deduplication process.

### DIFF
--- a/helper/diffPlaces.js
+++ b/helper/diffPlaces.js
@@ -124,15 +124,35 @@ function isNameDifferent(item1, item2, requestLanguage){
   // iterate over all the languages in item2, comparing them to the
   // 'default' name of item1 and also against the language requested by the user.
   for( let lang in names2 ){
+    if (!names2.hasOwnProperty(lang)) { continue; } // do not iterate over inherited properties
     if( !isPropertyDifferent({[lang]: names1.default}, names2, lang) ){ return false; }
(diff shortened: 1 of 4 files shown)


missinglink:
Coming back to this now after thinking it over...

I think it's good to merge, the concept that the parent names at the same layer as the feature can be used for deduplication is logical and unlikely to cause error.

The issue description was missing the concrete example which was [included in the tests](https://github.com/pelias/api/pull/1554/files#diff-46fb55122266003261b8890fd13bc9313f58b5bcb6097e2f4138a52723841141):

> Given the feature `geonames:region:2950157` ("Land Berlin", en: "State of Berlin") along with another feature at the same level `whosonfirst:region:85682499` ("Berlin") in the results.

> We can use the  `parent.region = ["Berlin"]` property of `geonames:region:2950157` to establish that it is equivalent to `whosonfirst:region:85682499`

The `parent.region_id = [85682499]` property of `geonames:region:2950157` further indicates that the two are duplicate, so we *could* consider only using the IDs, I'm open to both methods.

PR #1555, opened 2021-08-17
only translate name.default when translation is same source and source_id

missinglink:
The `middleware/changeLanguage.js` code is a little too liberal when replacing the `name.default` with a translation.
Prior to this PR the code replaces with `name.default` without checking the `source` and `source_id`.

note: I struggled a bit to read and write this code, requires a second set of 👀

### DIFF
--- a/middleware/changeLanguage.js
+++ b/middleware/changeLanguage.js
@@ -80,10 +80,10 @@ function updateDocs( req, res, translations ){
       var match = attr.match(/^(.*)_id$/);
       if( !match ){ continue; }
 
-      // adminKey is the property name without the '_id'
-      // eg. for 'country_id', adminKey would be 'country'.
-      var adminKey = match[1];
-      var adminValues = doc.parent[adminKey];
+      // adminLayer is the property name without the '_id'
+      // eg. for 'country_id', adminLayer would be 'country'.
+      var adminLayer = match[1];
+      var adminValues = _.get(doc.parent, adminLayer);
 
       // skip invalid/empty arrays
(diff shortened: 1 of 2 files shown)


orangejulius:
I think we talked about this the other day, but can you provide some more details and background here regarding the issue at hand, maybe some test cases, etc? It will be good to have for reference in the future.

middleware/changeLanguage.js
```diff
@@ -92,29 +92,39 @@ function updateDocs( req, res, translations ){
...
-          doc.name.default = translations[id].names[ requestLanguage ][0];
-        }
+        adminValues[i] = translations[adminID].names[ requestLanguage ][0];
+
+        // if the record is an admin record we also translate the 'name.default' property.
+        // note: ensure the records have the same 'layer' 'source' and 'source_id'.
+        if (adminLayer !== doc.layer){ continue; }
+        if (adminSource !== doc.source){ continue; }
+        if (_.toString(adminID) !== _.toString(doc.source_id)){ continue; }
+        doc.name.default = translations[adminID].names[requestLanguage][0];
```
orangejulius, line 127: I wonder if we actually need this line now. Looking through the history of this file, the [`translateNameDefault`](https://github.com/pelias/api/blob/ef59ef4da971b078ec1291e92b88f9a1281dc3fc/middleware/changeLanguage.js#L70) method was added in 2019 in https://github.com/pelias/api/pull/1301, whereas the rest of the file is mostly much older.

It might be that the name translation is already handled well now? I'm not 100% sure, the intention behind this code is definitely hard to read.

PR #1565, opened 2021-10-05
feat(parser): upgrade pelias-parser dependency

missinglink:
As discussed in https://github.com/pelias/api/issues/1550, this PR takes us off a branch of `pelias-parser` and gets us back on the latest published version.

BREAKING CHANGE: upgrades to a new major version of pelias/parser which reverts unit number parsing functionality due to OOM errors.

Additional changes in pelias/parser being brought in:
- https://github.com/pelias/parser/pull/138
- https://github.com/pelias/parser/pull/141
- https://github.com/pelias/parser/pull/142
- https://github.com/pelias/parser/pull/143
- https://github.com/pelias/parser/pull/126

Original commit to pin pelias-parser to a branch:
https://github.com/pelias/api/commit/4061a2fb3b90d1949b5fbc74ac2ab3344a529e2d

PR #1566, opened 2021-10-06
Add support to dependencies

Joxit:
I noticed that dependencies were often seen as countries, but it is not possible to use the `boundary.country` parameter to filter searches for documents that are located in dependencies.

Now the view `boundary_country` uses a `multi_match` query to match both `country_a` and `dependency_a`.

`npm install` is failing because the branch `pelias/parser#remove-unit-type` does not exists. So I sync my branch with `origin/parser-remove-unit-type-updated`
 
needs pelias/query#129

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "pelias-logger": "^1.2.0",
     "pelias-microservice-wrapper": "^1.10.0",
     "pelias-model": "^9.0.0",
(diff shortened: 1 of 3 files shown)


missinglink:
FYI: the pelias/parser branching issue should be resolved shortly in https://github.com/pelias/api/pull/1565

missinglink (approved):
See approval comments in https://github.com/pelias/query/pull/129

PR #1573, opened 2021-10-28
Return 502 response code for service timeouts instead of 400

orangejulius:
Pelias has always had a bit of trouble selecting the right HTTP response code in the face of various error states.

Up until #1231 in 2018, we reported almost all timeouts from slow Elasticsearch queries as HTTP 400 errors, not something in the more appropriate 5XX range. This suggests to consumers of the Pelias API that they made a mistake in calling Pelias, instead of the reality that Pelias was just being slow.

Even after that change, it turns out we were _still_ classifying timeouts to other Pelias services (like Placeholder or Interpolation) as 400 errors instead of 5XX. All the Pelias services are generally very fast, so this was not nearly as much of an issue, but timeouts do happen.

This PR adds additional handling to detect timeout errors and give them their own subclass of `Error` that can be treated appropriately everywhere. Timeouts waiting for any Pelias service will now return HTTP 502 errors just like a timeout waiting for Elasticsearch.

_There's a bit of code cleanup included in this PR. It's all a no-op and the main functional change can be found in 95bda7b8_

### DIFF
--- a/middleware/sendJSON.js
+++ b/middleware/sendJSON.js
@@ -2,13 +2,15 @@ const _ = require('lodash');
 const es = require('elasticsearch');
 const logger = require( 'pelias-logger' ).get( 'api' );
 const PeliasParameterError = require('../sanitizer/PeliasParameterError');
+const PeliasTimeoutError = require('../sanitizer/PeliasTimeoutError');
 
 function isParameterError(error) {
   return error instanceof PeliasParameterError;
 }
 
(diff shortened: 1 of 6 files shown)


orangejulius:
This is working pretty well for us so far. Again there's not a huge number of requests associated with this. We're running timeout values near the pelias default (1000ms) and see only a couple requests a week that take longer. But when they do happen we don't want to return an HTTP 400.

Issue #1574, opened 2021-10-29
Long query with special characters causes 500 error


orangejulius:
We recently saw some 500 errors in Geocode Earth production that initially gave us a bit of a scare! It turned out to be some fairly wild input from one of our users that triggered a 500 error in the Pelias API.

We've solved a few problems with special characters, emoji, long inputs, etc recently but this one is especially confusing to me.

Here's a URL that demonstrates the problem:

[...]

PR #1575, opened 2021-11-01
unicode: strip additional rune ranges

missinglink:
this PR removes two additional unicode ranges mentioned in https://github.com/pelias/api/issues/1574

- https://en.wikipedia.org/wiki/Regional_indicator_symbol (flags)
- https://en.wikipedia.org/wiki/Geometric_Shapes_Extended (colourful circles/squares)


missinglink:
I feel like we should probably put in a space between two valid text blocks delimited by stripped elements?

PR #1579, opened 2021-11-09
refactor: use ESLint & Prettier instead of `jshint`

vinayakkulkarni:

---
Instead of jshint, I added support for [`prettier`](https://prettier.io/) & [`ESLint`](https://eslint.org/)

---
Nothing as such, the source code is same but formatting is standardized

---
- `npm run lint:eslint` # Checks with ESLint
- `npm run lint:eslint:fix` # Fixes all issues with ESLint
- `npm run lint:prettier` # Checks with Prettier
- `npm run lint:prettier:fix` #Fixes all issues with Prettier


vinayakkulkarni:
PS. I'm also planning on adding ESLint & Prettier in other repos as well :)

missinglink:
hi @vinayakkulkarni for such a large change I think we would want to use a bot account and consider the git history here.

My personal preference would be for https://standardjs.com, other contributors may have different preferences.

I suggest you first open an issue on pelias/pelias to discuss with the contributors of the project what they might want before opening more of these large code refactoring PRs.

orangejulius:
Going to close this PR, and in the future I think that any wide-reaching style changing pull request would have to be made by a core maintainer.

Reasons being:
- We'd want to make the change from a bot account to avoid massively upsetting contributor stats (we'd also use the amazing [.git-blame-ignore-revs](https://rich.grundy.io/blog/ignore-large-refactor-commits-with-git-blame/) file to avoid `git blame` being made much less useful, though that's a bit of an aside)
- It's impossible to review a 17k line change, so instead we would have to programmatically generate the change on a trusted machine.

Of course feedback/discussion on what style to use is welcome from anyone, but we probably can't accept an outside PR to actually make the change.

PR #1583, opened 2021-11-15
pelias-parser: print parser solutions in API debug output

missinglink:
The `pelias-parser` CLI has a nice debug output of all the solutions generated, however this isn't available via the API debug view, this is annoying since it requires copy->pasting the input across to the terminal to inspect the parser behaviour.

In the past I thought we couldn't really take advantage of the `Debug` functionality in the `sanitizer/*` modules because we don't have access to `req` like we do in the `middleware/*` modules.

The minor change I made to `sanitizer/sanitizeAll.js` in this PR exposes the `req` param to all sanitisers in a backward compatible way.

The rest of this PR is an implementation of that workflow where we use the same logic from the parser CLI to print a concise summary of each solution.

Screenshot 2021-11-15 at 14 00 44

### DIFF
--- a/sanitizer/_text_pelias_parser.js
+++ b/sanitizer/_text_pelias_parser.js
@@ -1,4 +1,6 @@
 const logger = require('pelias-logger').get('api');
+const Debug = require('../helper/debug');
+const debugLog = new Debug('santizer:text:pelias_parser');
 const unicode = require('../helper/unicode');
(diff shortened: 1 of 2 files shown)


missinglink:
played around with a few versions of this and quite happy with this:

Screenshot 2021-11-15 at 15 02 11

Joxit (approved):
LGTM

Issue #1405, opened 2019-12-02
discuss: enabling categories on all endpoints


missinglink:
Issue https://github.com/pelias/api/pull/1401 brought up the topic of whether we should enable the `categories` feature by default across all endpoints.

Currently, we have support for categories which [goes back to 2015](https://github.com/pelias/pelias/wiki/Taxonomy-v1).
The feature consists of a single elasticsearch field with strings analyzed as `keyword`, so it's very simple but can also be very powerful for industry and interest-specific queries.

When we went to release the feature there was some concern internally at Mapzen about the default taxonomy labels I had selected and the WOF team was preparing an improved list, this work unfortunately never got finished and so the feature was left in an awkward state of half-release, available, yet also hidden from view.

Over the years we've had [a bunch of issues and PRs](https://github.com/pelias/api/issues?utf8=%E2%9C%93&q=categories) opened regarding `categories` and the theme seems to be that organizations find them useful.

What would be required to completely release categories as a first-class feature?
- improved documentation including how to use custom category labels
- return categories by default in all geojson feature responses

Are there any concerns regarding the release of categories?
- once released any category labels become part of the public API, making them difficult to change or remove
- some organizations might prefer to opt-out of displaying categories, so it should really be optional?
- not all endpoints support categories, for instance: placeholder or PIP service results might not filter or display as expected

This ticket is open to discuss the pros/cons of enabling the categories feature by default and think through the work which would be required if we decided to go down that route.

cc/ @pelias/contributors

Joxit:
Hi there and Happy New Year :smile: 

I think that for those who know this feature, it's already too late, we cannot do breaking changes or delete categories.

The use of categories bypass placeholder, this gives some strange responses when we compare a query with and without the query parameter. 
[/v1/search?text=paris](https://pelias.github.io/compare/#/v1/search%3Ftext=paris)
```
 1) Paris, France
 2) Paris, TX, USA
 3) Paris, ON, Canada
 4) Paris, TN, USA
 5) Paris Township, OH, USA
 6) Paris Township, IL, USA
 7) Paris Township, OH, USA
 8) Paris, ME, USA
 9) Paris, IA, USA
10) Paris, Indonésie
```
[/v1/search?categories=&text=paris](https://pelias.github.io/compare/#/v1/search%3Fcategories=&text=paris)
```
 1) Métropole du Grand Paris, France
 2) Paris, France
 3) Paris, France
 4) Paris, TX, USA
 5) Paris Township, OH, USA
[...]

PR #1588, opened 2021-11-30
Fix category sanitizer for new call format

orangejulius:
It turns out the changes to the parameters for sanitizer functions in #1583 wasn't _quite_ backwards compatible.

## What broke?

The `categories` parameter sanitizer supports taking a 3rd parameter that defines how to determine valid categories. If there's only two parameters, it uses a default function that always returns true.

We intentionally haven't documented the categories parameter because it's not really done, and the categories mapping is not finalized. As a result almost no one uses it.

However, if they did use it, and passed in a category, they could cause the API to generate a 500 error.

This is no problem to fix, just expect the 3rd parameter to be the `req` object, and the 4th parameter, if we ever even use it, is now that category validation function.

## Aside: commentary on the categories sanitizer code and parameter

Realistically, the design here probably isn't useful, I'm not sure how we'd want to pass in a validation function in practice. I thought a lot about removing this code.

There's also some nearby code from https://github.com/pelias/api/pull/1401 that is effectively a second categories sanitizer _just_ for the place endpoint. If we ever take action on https://github.com/pelias/api/issues/1405 we'd probably end up rewriting all of this 🤷

## Thoughts on testing

Finally, I checked through all the other sanitizers to see if any might be affected similarly to this one. I don't think so, but I did clean up one unused parameter to another little used sanitizer.

I'm trying to think of an easy way for us to have caught this error. It showed up in the acceptance tests, but only for the little-used place endpoint tests, which we basically ignore.

It didn't cause any unit tests to fail, and the `ciao` tests (if we ran them regularly), wouldn't have currently caught it.

What should we add to our testing to catch stuff like this?

### DIFF
--- a/sanitizer/_categories.js
+++ b/sanitizer/_categories.js
@@ -7,7 +7,7 @@ const WARNINGS = {
 };
 
 // validate inputs, convert types and apply defaults
-function _sanitize (raw, clean, validator) {
+function _sanitize (raw, clean, req, validator) {
   validator = validator || allValidValidator;
 
   // error & warning messages
--- a/sanitizer/_geonames_deprecation.js
+++ b/sanitizer/_geonames_deprecation.js
@@ -9,7 +9,7 @@ const _ = require('lodash');
 
 const coarse_reverse_message ='coarse /reverse does not support geonames. See https://github.com/pelias/pelias/issues/675 for more info';
 
-function _sanitize( raw, clean, opts ) {
+function _sanitize( raw, clean ) {
   // error & warning messages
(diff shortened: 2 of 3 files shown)


missinglink:
Oh nice catch, I grepped and grepped again looking for code that touched that arguments change.

At the time I remember being worried mostly about usages of `Function.bind()` since they are harder to detect.

I would love to bring back the idea of spinning up the API server and running tests against it (as per ciao).

But doing that is a little annoying to configure (especially if data and all microservices are required), plus I haven't found a simple replacement for ciao yet.

Yeah dunno, good catch

PR #1595, opened 2022-01-07
debug-log: print placeholder response verbatim

missinglink:
While attempting to debug a query, I looked at the `controller:placeholder` debug log line and noticed that it doesn't actually show the response from placeholder 🤔 

What's being rendered there instead is `$res.data`, which is the modified state of the response *after* the placeholder results have been merged with it.

I'm not sure of the history of this, which is several years old, but I would prefer to show the verbatim response from the microservice response in the logs so I can inspect the response data *before* it is merged.

cc/ @orangejulius thoughts?

### DIFF
--- a/controller/placeholder.js
+++ b/controller/placeholder.js
@@ -302,8 +302,7 @@ function setup(placeholderService, do_geometric_filters_apply, should_execute) {
         ];
 
         logger.debug(messageParts.join(' '));
-        debugLog.push(req, messageParts[1].slice(1,-1));
-        debugLog.push(req, res.data);
+        debugLog.push(req, (results || []));
       }
 
       debugLog.stopTimer(req, initialTime);
--- a/test/unit/controller/placeholder.js
+++ b/test/unit/controller/placeholder.js
@@ -66,139 +66,138 @@ module.exports.tests.should_execute = (test, common) => {
 };
 
 module.exports.tests.success = (test, common) => {
+  const response = [
+    {
+      id: '123',
+      name: 'name 1',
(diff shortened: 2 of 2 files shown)


controller/placeholder.js
```diff
@@ -302,8 +302,7 @@ function setup(placeholderService, do_geometric_filters_apply, should_execute) {
         ];
 
         logger.debug(messageParts.join(' '));
-        debugLog.push(req, messageParts[1].slice(1,-1));
```
missinglink, line 305: this `.slice()` command is odd, the purpose appears to be to trim the square brackets off the start and end.
it was somewhat convenient to print the `count` (ie. `results.length`) but that's probably not helpful in practise.

test/unit/controller/placeholder.js
```diff
@@ -66,147 +66,146 @@ module.exports.tests.should_execute = (test, common) => {
...
 
 module.exports.tests.success = (test, common) => {
-  test('response from service should be converted', (t) => {
-    const logger = mock_logger();
-
-    const placeholder_service = (req, callback) => {
-      t.deepEqual(req, { param1: 'param1 value' });
-
-      const response = [
+  const response = [
```
missinglink, line 69: this `$response` object was simply moved up one scope without modification so it could be referenced below.

orangejulius:
Makes sense, the raw Placeholder result is more "unfamiliar" (looks like the API basically converts it to look like a result from Elasticsearch), but it's more truthful.

orangejulius (approved):

PR #1596, opened 2022-01-28
config: enable/disable prefix matching numerals feature via config flag

missinglink:
numeral prefix matching was introduced in https://github.com/pelias/api/pull/1514 for the purpose of prefix-matching numeric postalcodes.

we've had a request to optionally disable this functionality, in some installations this behaviour has undesirable effects on custom layers such as a transit `stops` layer which has numerical stop IDs.

this PR introduces a 'feature flag' which disables that feature via config, but leaves it on by default.
I'm somewhat hesitant to introduce many feature flags as they make testing and debugging more difficult.

until such a time as we're able to better handle disabling numeral prefix matching on custom layers this is the really only viable solution.

the following configuration will disable https://github.com/pelias/api/pull/1514 on a per-installation basis:

```bash
{
  "api": {
    "feature": {
      "prefix_match_numerals": false
    }
  }
}
```

### DIFF
--- a/sanitizer/_tokenizer.js
+++ b/sanitizer/_tokenizer.js
@@ -1,4 +1,33 @@
 const _ = require('lodash');
+const config = require('pelias-config').generate();
+
+// optionally disable prefix matching numerals
+const feature = {
+  prefix_match_numerals: config.get('api.feature.prefix_match_numerals', true)
+};
+
+// only in the event that the final token is a numeral, this function
+// decides whether that numeric token should be considered complete or not.
+const shouldMarkNumericFinalTokenAsComplete = (clean) => {
+
+  // this feature may optionally be disabled via config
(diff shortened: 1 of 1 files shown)


sanitizer/_tokenizer.js
```diff
@@ -1,4 +1,33 @@
...
+  // this will include addresses, so disable for the same reasons as below.
+  if (!_.isArray(layers) || _.isEmpty(layers)) { return true; }
+
+  // user has explicitely requested the address layer.
+  // avoid prefix matching house number numerals.
+  if (layers.includes('address')) { return true; }
+
+  // default behaviour is to allow prefix matching numerals
+  return false;
+};
```
orangejulius, line 30: Moving all this out to a function where we can clearly document it is huge!

orangejulius (approved):

PR #1598, opened 2022-02-01
Added Missing HTML / BODY tag

### LINKED ISSUE #1597: markdownToHtml controller doesn't render valid HTML
the [controller/markdownToHtml.js](https://github.com/pelias/api/blob/master/controller/markdownToHtml.js) module doesn't render valid HTML, it's missing the `<HTML>`, `<BODY>` tags.

this issue is to modify that behaviour so that requests to `/` and `/attribution` render valid HTML

Screenshot 2022-02-01 at 09 27 50

Dhananjay-JSR:

---

  - This PR Closes #1597 
  - The reason PR was required because due to an error in variable , the api couldn't return a static html for markdown 

---
#### Here's what actually got changed 
Added HTML and BODY tag in [file](https://github.com/pelias/api/blob/master/controller/markdownToHtml.js)
changes

---

### DIFF
--- a/controller/markdownToHtml.js
+++ b/controller/markdownToHtml.js
@@ -7,7 +7,7 @@ function setup(peliasConfig, markdownFile){
   var text = '# Pelias API\n';
   text += '### Version: [' + peliasConfig.version + '](https://github.com/pelias/api/releases)\n';
   text += fs.readFileSync( markdownFile, 'utf8');
-  var html = styleString + markdown.toHTML(text);
+  var html = '<html><body>'+styleString + markdown.toHTML(text)+'</body></html>';
 
   function controller( req, res ) {
     if (req.accepts('html')) {
--- a/controller/markdownToHtml.js
+++ b/controller/markdownToHtml.js
@@ -3,22 +3,27 @@ var fs = require('fs');
 
 function setup(peliasConfig, markdownFile){
 
-  var styleString = '<style>html{font-family:monospace}</style>';
-  var text = '# Pelias API\n';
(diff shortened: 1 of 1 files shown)


missinglink:
hey, thanks for looking into this, a couple minor changes I'd like to see before merging:

- use [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) to make the markdown and html code more legible.
- use a standard HTML5 boilerplate with a DOCTYPE, TITLE etc.
- remove the `styleString` variable and instead just write the style info inline in the `<head><style>`

```js
const text = `
# Pelias API
### Version: [${peliasConfig.version}](https://github.com/pelias/api/releases)
${fs.readFileSync( markdownFile, 'utf8'}`

const html = `
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="utf-8">
  <title>Pelias Geocoder</title>
  <style>html { font-family:monospace; }</style>
</head>
<body>
  ${markdown.toHTML(text)}
</body>
</html>`
```

missinglink:
can you also remove the content negotiation code, we don't need that and AFAIK it doesn't even work 🤷‍♂️ 

the controller can be simplified to:

```js
function controller( req, res ) {
  res.send(html);
}
```

Dhananjay-JSR:
Made the changes as stated can you please check the PR once again

missinglink:
I've pulled your commits into https://github.com/pelias/api/pull/1599 and merged, you may now close this PR/branch.

Thank you 🙇

PR #1600, opened 2022-02-02
default-parameters: extend functionality to be more general purpose

missinglink:
in https://github.com/pelias/api/pull/928 we introduced the ability to apply default values to the `focus.point.lat` and `focus.point.lon` query string params if/when they were unset.

this PR refactors that code in such a way that it allows us to extend the list of properties which can be overridden by defaults.

note: there is a concept of 'parameter groups' which I've maintained, the idea being that it makes sense for some parameters to be treated as a cohesive group (such as lat/lon pairs) and if one or the other is missing then we don't make modifications.


missinglink:
should we emit a warning when a default parameter is used?

sanitizer/search.js
```diff
@@ -13,7 +30,7 @@ module.exports.middleware = (_api_pelias_config) => {
         // depends on the layers and sources sanitizers, must be run after them
         sources_and_layers: require('../sanitizer/_sources_and_layers')(),
         private: require('../sanitizer/_flag_bool')('private', false),
-        location_bias: require('../sanitizer/_location_bias')(_api_pelias_config.defaultParameters),
+        // location_bias: require('../sanitizer/_location_bias')(_api_pelias_config.defaultParameters),
```
orangejulius, line 33: You can remove this, right? :)

missinglink: resolved via rebase

missinglink:
Screenshot 2022-02-02 at 18 49 44
Screenshot 2022-02-02 at 18 49 58
Screenshot 2022-02-02 at 18 50 06
Screenshot 2022-02-02 at 18 50 15

orangejulius:
This looks good. Since the PR as it stands is a pure refactoring no-op, how about the following:

Let's set up another PR on top of this that adds a couple parameters to the list that can have a default. Maybe `sources`, `layers`, `boundary.country`? Then we can test that PR and make sure it all works as we expect. There's not really much of a reason to merge anything until it impacts actual functionality.

I think we can skip the warning for default parameters. Whoever is configuring Pelias will be aware of what's going on.

missinglink:
Yeah, so before we merge this we should add *all* params for each endpoint to the 'group lists' and add the functionality to any endpoints which don't currently have it (such as reverse).

Since it still requires adding new properties to the `pelias.json` config, this would still be a no-op unless a developer opted in to it.

orangejulius (approved):

PR #1606, opened 2022-02-28
Dedupe Geonames records with WOF concordances

orangejulius:
This PR implements deduplication between WOF and Geonames when there's a Geonames concordance ID on the WOF record.

These concordances mean we can be fairly certain that the two records are the same, and that we don't even have to look at the name, or any other properties.

Still in draft state: no tests yet, but the general logic should be solid.

### DIFF
--- a/helper/diffPlaces.js
+++ b/helper/diffPlaces.js
@@ -169,11 +169,47 @@ function isAddressDifferent(item1, item2){
   return false;
 }
 
+function isGeonamesConcordanceSame(item1, item2) {
+  let wof_record;
+  let gn_record;
+
+  if (item1.source === 'geonames' && item2.source === 'whosonfirst') {
+    gn_record = item1;
+    wof_record = item2;
+  } else if (item2.source === 'geonames' && item1.source === 'whosonfirst') {
+    gn_record = item2;
+    wof_record = item1;
+  } else {
+    // could not match to one geonames and one wof concordance, so this check does not apply
+    return false;
+  }
+
+  const concordances = _.get(wof_record, 'addendum.concordances');
+
+  if (concordances) {
+    const json = JSON.parse(concordances);
+
+    //console.log(`comparing gn ${gn_record.source_id} to concordance ${json['gn:id']} (wof ${wof_record.source_id})`);
+
+    const concordance_id = json['gn:id'];
+
+    if (concordance_id && typeof concordance_id === 'number' && json['gn:id'].toString() === gn_record.source_id) {
+      console.log(`gn ${gn_record.source_id} matches concordance from wof ${wof_record.source_id}`);
+      return true;
+    }
+  }
+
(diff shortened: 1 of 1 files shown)


missinglink (changes requested):
Conceptually this is great :+1:

I'm not in love with the code style for some reason, maybe we can use `lodash` to clean this up a bit? [Something like this](https://github.com/pelias/api/blob/master/helper/diffPlaces.js#L141-L170) might be 'nicer'?

I'll add a couple in-line comments but yeah, we should add some tests and merge this at some point.

helper/diffPlaces.js
```diff
@@ -169,11 +169,47 @@ function isAddressDifferent(item1, item2){
...
+    wof_record = item1;
+  } else {
+    // could not match to one geonames and one wof concordance, so this check does not apply
+    return false;
+  }
+
+  const concordances = _.get(wof_record, 'addendum.concordances');
+
+  if (concordances) {
+    const json = JSON.parse(concordances);
```
missinglink, line 190: there is a 'codec' in `pelias/model` which is actually just using `JSON.parse` under-the-hood but serves as a nice abstraction for if we later want to change the encoding or apply compression or whatever.

there are some examples in this repo already, have a look for `const codec = require('pelias-model').codec;`

missinglink (commented):
Also worth mentioning that it would be great it we could make this check more generalized.

For instance, if an OSM record, a WOF record and a GEONAMES record all shared a Wikidata concordance then we should be able to conflate all three using a third-party concordance.

missinglink:
Had a shower thought about this, it might be easier & more generic to do it like this:
- decode the concordances, or create an empty `{}` for each record
- append self-references, such as setting `gn:id` for `geonames` records within it's own concordances.
- intersect all the concordance maps with each other, anything with an intersection size > 0 is a match

At a later stage we could add some logic to virally merge concordances to make them work across 3+ degree connections but that's probably out-of-scope for now.

orangejulius:
I've finished this off with a test, use of the `pelias-model` `codec`, and a different code style. It now avoids any use of `let`, or `typeof`. The structure is also simpler and not as deeply nested as all the checks now bail out early.

I like the strategy you've described for how we might generalize concordance checks in the future. It's something to keep in mind for sure. I don't _think_ it really solves a problem we see today: the only possible deduplication between GN/OSM would be venues, and so few of them have matching Wikidata concordances it might not be worth it.

It also looks like Geonames records do not themselves have Wikidata concordances. So we would have to look up the data from Wikidata. For example, see La Sagrada Familia in [OSM](https://www.openstreetmap.org/relation/9194723), [Geonames](https://www.geonames.org/6452887/temple-expiatori-de-la-sagrada-familia.html), and [Wikidata](https://www.wikidata.org/wiki/Q48435).

PR #1371, opened 2019-10-09
dedupe placetype in name

missinglink:
~~note: branched off https://github.com/pelias/api/pull/1370, [view diff](https://github.com/pelias/api/compare/dedupe-normalization...dedupe-placetype-in-name)~~

this PR adds some logic which allows strings such as `City of Philadelphia` and `Philadelphia` to match for the sake of deduplication.

the normalization function is only applied for the applicable layers, so in the case above the `City of Philadelphia` must be on the `locality` layer in order to be considered for normalization.

I've added a bunch of tests and also ensured that the original data is not mutated so the display labels will remain unaffected by the deduplication normalization.

Closes https://github.com/pelias/geonames/pull/372
Resolves https://github.com/pelias/geonames/issues/395

PR #1607, opened 2022-02-28
simplify deduplication logic

missinglink:
simplify deduplication logic


missinglink:
This PR improves the deduplication of https://pelias.github.io/compare/#/v1/autocomplete?text=berlin

missinglink:
Good-to-go! I think this is a positive change which should help reduce some of the noisier deduplication cases that have been plaguing us for some time now 🎉

missinglink:
housekeeping: the [dedupe-without-skiplist](https://github.com/pelias/api/tree/dedupe-without-skiplist) branch can probably be deleted after merging this.

orangejulius:
Cool, I really love the simplicity of this logic. Even if it's technically more computationally expensive (though I would imagine the effect is not measurable), it's _much much_ easier to understand.

 I do think the case you mentioned with transitive deduplication has prevented a lot of our deduplication checks from performing as we expect.

My guess is this is good to merge but let me try it out with some of the other recent changes I've been experimenting with, like #1609 and https://github.com/pelias/api/pull/1606.

orangejulius:
Okay, after some fairly complex testing I can confirm this PR _improves_ deduplication, or more precisely it allows some of the newer deduplication strategies we are experimenting with to be more effective.

Here's a summary of some results from testing with https://github.com/pelias/acceptance-tests/pull/557

branch | Passing tests
-- | --
master | 2
A: simplified dedupe (this PR) | 2
B: dedupe-placetype-in-name (https://github.com/pelias/api/pull/1371) | 3
C: concordance-dedupe (https://github.com/pelias/api/pull/1606) | 3
A + B | 3
A + C | 3
B+ C | 4
A+ B + C | 5

The interesting thing to note is that while this PR by itself does not change the test results, nor does it change the results when combined with either of our two other deduplication PRs, when all three PRs are combined, there is an improvement over _not_ using this PR.

So that seems to confirm to me the idea that transitive deduplication is important.

One last thing to test is that this PR (as well as a combination of this and the other PRs) does not introduce regressions. That info will be coming soon, then I think we are good to go.

PR #1610, opened 2022-03-01
do not dedupe disparate unit numbers at same address

missinglink:
as mentioned in https://github.com/pelias/openaddresses/pull/504 the current deduplication behaviour considers all units at the same address to be duplicate, therefore 'squashing them' into a single response.

this PR modifies this behaviour so that each individual unit number at an address returns a row in the results.

I'm opening this as DRAFT for now since @orangejulius mentioned in https://github.com/pelias/openaddresses/pull/504#issuecomment-1054461416 that we might consider alternatives, I wanted to have a PR open to remind us.

### DIFF
--- a/helper/diffPlaces.js
+++ b/helper/diffPlaces.js
@@ -157,6 +157,7 @@ function isAddressDifferent(item1, item2){
   if( !isPojo1 || !isPojo2 ){ return false; }
 
   // else both have address info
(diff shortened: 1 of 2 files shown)


missinglink:
duplicate of https://github.com/pelias/api/pull/1472

PR #1612, opened 2022-03-03
simplified dedupe + preserve order

missinglink:
an extension to https://github.com/pelias/api/pull/1607 which preserves ordering of synonymous records (ie. the lower ordinal position is inherited by the superior record).


orangejulius:
All our acceptance testing shows this is either a no-op or a slight improvement. I think this is good to go, and then we can start evaluating the other deduplication related PRs: https://github.com/pelias/api/pull/1606 and https://github.com/pelias/api/pull/1371, since their full benefits are only unlocked with this change to the deduplication logic.

PR #1614, opened 2022-03-07
Do not deduplicate US States

orangejulius:
It's common for US states to have either a county or city within them that shares a name (with minor possible differences).

For example:
- Arkansas City in Arkansas
- Hawaii County in Hawaii
- Iowa City in Iowa
- California City in California
- Utah County in Utah
- Nebraska City in Nebraska
- Idaho City _and_ Idaho County in Idaho (Idaho City is _not_ in Idaho County)
- Maryland City, in Maryland
- Minnesota City in Minnesota
- New York City in New York (of course!)

Our general deduplication logic considers admin records that are parented by a record sharing its name to be the same. This works well for places like Singapore, Berlin, and Tokyo, which all have a city (locality) that is conceptually the same as a region or country.

US states, however, are _not_ conceptually the same as any of these cities, and they should pretty much always show up in results.

This PR adds a check that the record is _not_ a US state before performing the general hierarchy checks. There's one exception: US states can dedupe against other US states, so that Geonames and WOF records can deduplicate themselves.

This PR allows us to merge https://github.com/pelias/api/pull/1371

### DIFF
--- a/helper/diffPlaces.js
+++ b/helper/diffPlaces.js
@@ -40,6 +40,11 @@ function isLayerDifferent(item1, item2){
   return false;
 }
 
+function isUsState(item) {
+  if (!_.isArray(item.parent.country_a)) { return false; }
+  return item.parent.country_a[0] === 'USA' && item.layer === 'region';
+}
+
 /**
  * Compare the parent properties if they exist.
  * Returns false if the objects are the same, else true.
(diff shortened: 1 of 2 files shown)


helper/diffPlaces.js
```diff
@@ -40,6 +40,11 @@ function isLayerDifferent(item1, item2){
   return false;
 }
 
+function isUsState(item) {
```
missinglink, line 43: nit: I would do the later check first since it's cheap and quickly discards most undesirable cases

missinglink: it might be possible to use lodash `_.head` to make it even simpler

helper/diffPlaces.js
```diff
@@ -59,6 +64,13 @@ function isParentHierarchyDifferent(item1, item2){
   // note: this really shouldn't happen as at least one parent should exist
   if( !isPojo1 || !isPojo2 ){ return false; }
 
+  // US states should not be deduplicated, except against other US states
```
missinglink, line 67: trying to wrap my head around this, is it an XOR?

missinglink: there is another function called `isLayerDifferent` which handles figuring out which layers can be considered synonymous, maybe it makes more sense to have this functionality in there? as like a subset of the includes check?

PR #1616, opened 2022-03-10
enable autocomplete of country-codes

missinglink:
enable autocomplete of country-codes


query/autocomplete.js
```diff
@@ -9,7 +9,6 @@ const toSingleField = require('./view/helper').toSingleField;
 // additional views (these may be merged in to pelias/query at a later date)
 var views = {
   custom_boosts:              require('./view/boost_sources_and_layers'),
-  ngrams_strict:              require('./view/ngrams_strict'),
```
missinglink, line 12: this wasn't used by anything 🤷

PR #1618, opened 2022-03-22
clamp boundary.circle.radius for reverse queries

missinglink:
I noticed that the `boundary.circle.radius` param is currently unbounded for `/v1/reverse`.
If a users sets this value very high it can result in a lot of disk I/O as ES pages in the docvalues to score all the results.

This PR adds some minimum and maximum values for this radius value, I've selected `5km` as the max and an arbitrarily low value of `0.00001km` for the min; although I'm open to changing either of these.


sanitizer/_geo_reverse.js
```diff
@@ -1,10 +1,9 @@
-var geo_common = require ('./_geo_common');
-var _ = require('lodash');
-var defaults = require('../query/reverse_defaults');
```
missinglink, line 3: this was unused

sanitizer/_geo_reverse.js
```diff
@@ -1,10 +1,9 @@
...
-var LAT_LON_IS_REQUIRED = true,
-    CIRCLE_IS_REQUIRED = false;
-
-const non_coarse_layers = ['venue', 'address', 'street'];
```
missinglink, line 7: this was unused

test/unit/sanitizer/_geo_reverse.js
```diff
@@ -61,27 +61,88 @@ module.exports.tests.warning_situations = (test, common) => {
...
 
 module.exports.tests.success_conditions = (test, common) => {
+  // note: this behaviour is historic and should probably be removed.
+  // it's possible to add non-numeric tokens to the boundary.circle.radius
+  // value and they are simply stripped out by parseFloat() with warning.
+  test('boundary.circle.radius must be a positive number', (t) => {
+    const raw = {
+      'point.lat': '12.121212',
+      'point.lon': '21.212121',
+      'boundary.circle.radius': '1km' // note the 'km'
```
missinglink, line 71: this is weird, I'd like to change it, but in another PR.
right now you can specify `5km` or `5m` and in both cases the unit designation is stripped and results in `5`, which is `km`.

test/unit/sanitizer/_geo_reverse.js
```diff
@@ -61,27 +61,88 @@ module.exports.tests.warning_situations = (test, common) => {
...
+      'point.lat': '12.121212',
+      'point.lon': '21.212121',
+      'boundary.circle.radius': '0.000000000000001'
+    };
+    const clean = {};
+    const errorsAndWarnings = sanitizer.sanitize(raw, clean);
+
+    t.equals(raw['boundary.circle.lat'], 12.121212);
+    t.equals(raw['boundary.circle.lon'], 21.212121);
+    t.equals(raw['boundary.circle.radius'], '0.000000000000001');
```
orangejulius, line 138: I think you can skip testing the `raw` values, since you specified them above.

IMO it just makes the test longer and it took me a second to realize it was checking the raw value, not clean. At first glance it makes this test look like it doesn't work!

sanitizer/_geo_reverse.js
orangejulius: Huh, I wonder when we last used it... We have long needed to improve reverse geocoding support for custom layers, as mentioned in https://github.com/pelias/api/issues/1161. Not something to solve in this PR though.

orangejulius (approved):
Looks good. I have a suggestion on the tests but it's very much optional. :)

test/unit/sanitizer/_geo_reverse.js
missinglink: Hmm actually no, that's not right, because if so, then the raw values would all be strings 🤷‍♂️ 
Anyway, I just cleaned up the tests by removing the assertions for `$raw`

Issue #1591, opened 2021-12-18
Add some retries to type mapping discovery?


orangejulius:
A helpful but sometimes confusing part of the APIs behavior is that it queries Elasticsearch on startup to get the list of [custom sources and layers](https://github.com/pelias/api#custom-sources-and-layers).

We had a report from a Pelias user with custom layers that they were no longer seeing them after running `pelias compose down && pelias compose up`. It turned out that telling both the API and Elasticsearch to start up at the same time doesn't work. Tthe API's initial query to Elasticsearch will fail, so it won't detect any custom sources and layers.

While we have the `pelias elastic wait` command to help with this, the "safe" way to stop and start all Pelias components is rather verbose: `pelias compose kill && pelias elastic start && pelias elastic wait && pelias compose up`.

What if instead we added a couple retries to the API's Elasticsearch request. Elasticsearch usually starts up within a few seconds, and then the call to detect custom sources and layers would work fine.

PR #1624, opened 2022-05-09
Add option to quit on type mapping discovery errors

orangejulius:
This adds a new config option, `api.targets.auto_discover_required`. If set to true, the API will quit with a fatal error if there is any problem performing the type mapping discovery.

In practice what this means is a working Elasticsearch cluster with a valid Pelias index must exist and be properly configured in `pelias.json`, or the API won't start.

This can actually be quite helpful in production environments as it can protect against misconfiguration issues such as using the wrong URL to the Elasticsearch cluster.

Because the check is only performed once at startup, it won't cause the API to quit if there's an intermittent network issue or something similar later on.

This PR includes the changes from #1625 so the Express webserver does not even start until the TypeMapping has been loaded. That means there's zero chance of any requests being served without correct config information.

Connects https://github.com/pelias/api/issues/1591
Connects https://github.com/pelias/pelias/issues/925


missinglink (approved):
LGTM, I tried to come up with a better name for the variable but failed 🤷‍♂️ 

note: PR could do with a git squash/rebase

PR #1625, opened 2022-05-10
Load type mapping before starting webserver

orangejulius:
This is an extension to #1624 that ensures the type mapping is fully loaded from Elasticsearch before the Express webserver is started.

In my testing of #1624 I noticed that it's possible for a few requests to sneak by before the type mapping has loaded (either with success or failure).

This means in the case where valid type mapping data can't be loaded from Elasticsearch (because there's no data or no Elasticsearch cluster correctly configured at all), the API can successfully start up, attempt to serve requests (though they would probably result in errors), and then quit.

This does _not_ play well with health checks like those that come with Kubernetes, ELBs, etc. Instead what we want is for the webserver to not come up at all until it's proven that the type mapping has loaded.

The change to do this wasn't too complicated but it ends up messy: a ton of code depends on the type mapping having been loaded beforehand, including tests. While having the type mapping `.load()` call happen before the Express webserver is started is desired, I'd love to find a way to avoid having to do something similar before running (or even `require`-ing) the unit test code. @missinglink if you have any ideas let me know.

### DIFF
--- a/helper/TypeMapping.js
+++ b/helper/TypeMapping.js
@@ -126,15 +126,13 @@ TypeMapping.prototype.loadTargets = function( targetsBlock ){
 
 // load values from either pelias config file or from elasticsearch
 TypeMapping.prototype.load = function( done ){
+  // load targets from config file
+  this.loadFromConfig();
 
-  // load pelias config
   const peliasConfigTargets = _.get(
     require('pelias-config').generate(require('../schema')),
     'api.targets', {}
   );
-
-  // load targets from config file
-  this.loadTargets( peliasConfigTargets );
 
   // do not load values from elasticsearch
   if( true !== peliasConfigTargets.auto_discover ){
@@ -144,6 +142,19 @@ TypeMapping.prototype.load = function( done ){
 
   // load values from elasticsearch
   loadFromElasticsearch(this, done);
+};
+
+// load values from either pelias config file or from elasticsearch
+TypeMapping.prototype.loadFromConfig = function( done ){
+
+  // load pelias config
+  const peliasConfigTargets = _.get(
+    require('pelias-config').generate(require('../schema')),
+    'api.targets', {}
+  );
+
+  // load targets from config file
+  this.loadTargets( peliasConfigTargets );
 };
 
 // replace the contents of an object or array
--- a/helper/type_mapping.js
+++ b/helper/type_mapping.js
@@ -1,8 +1,11 @@
 const TypeMapping = require('./TypeMapping');
 
-// instantiate a new type mapping
+// instantiate a singleton type mapping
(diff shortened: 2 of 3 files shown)


missinglink:
Hmm yeah I feel like the unit tests requiring a wrapper is an indication of the control flow bring amiss.

Could we just add an [EventEmitter](https://nodejs.dev/learn/the-nodejs-event-emitter) to the type mapping which emits a `load` event?

We could then subscribe to events with a `.once()` function which wraps `server.bind()` but not the unit tests?

missinglink:
Or event simpler the hralthcheck endpoint has a bool which is false and becomes true until receiving the event, then we return a 400 for false and a 200 for true

missinglink:
I guess the issue with this approach would be that some scripts may assume that once the server is bound that it is available.

So we'd either need to check both that the port is bound AND the hralthcheck is successful before sending requests.

We can certainly add that to the `pelias` binary but would it be inflexible in other environments? 🤔

orangejulius:
I like the `EventEmitter` idea. It's a bit of a shame since the TypeMapping class already has a callback parameter all set up, but it does make things complicated.

Using a `bool` like your code sample would probably work in most cases (it would be just fine for Kubernetes health checks which _do_ look at the HTTP code). But you're right that it's not a 100% clean solution either.

EventEmitters aren't too hard to work with, so lets try that out!

orangejulius:
Ah, I realized the EventEmitter won't buy us anything. The issue being worked around is that the `TypeMapping.load` call _must_ complete before _any_ of the unit tests are even `require`-d. I think instead we can and should fix that issue.

The issue I saw in the tests come down to the list of `canonical_layers` being empty at startup. So perhaps we can move the code that initializes that from the `TypeMapping.load` method to the constructor. That way, lines of code like [this one](https://github.com/pelias/api/blob/418bf0489d168b9824206c1f5d06a5afb457e29e/helper/diffPlaces.js#L5) will continue to work properly.

missinglink:
Agh I didn't realize what the issue was with the tests, but yeah the initial load from the config can be sync and then the http req can be async.

orangejulius:
Okay, that was relatively simple. The commit in this PR now splits the `TypeMapping.load()` function into two parts. The defaults are loaded from `pelias-config` as the `type_mapping` singleton file is `require`-d. Changes from Elasticsearch are optionally loaded later, and the callback for that async method is used in `index.js` to wait on starting the Express webserver.

This should be pretty foolproof and ensure no HTTP requests are served (and indeed the port isn't even open) until the type mapping is loaded and we are sure there's a valid Elasticsearch cluster configured.

orangejulius:
This work has been brought into #1624 for final release

PR #1626, opened 2022-06-20
update pelias/sorting module to fix numeric sorting bug

missinglink:
this PR fixes a subtle sorting bug: https://github.com/pelias/sorting/pull/26

in particular, a user noted that the query `Acton, ME, USA` returned the result for `Meeker County` above that of `Maine`, which is now resolved.


missinglink:
I've added a second commit which removes the `stable` module since stable sort has been available natively since node@12.

> Note: the comparator is expected to return a finite number, I hunted for any cases where it might return a boolean and couldn't find any in the code, one in the tests, worth keeping a 👁️ out for.

PR #1627, opened 2022-06-21
libpostal: discard parse containing single "suburb" label

missinglink:
It seems that `libpostal` gets confused with airports, returning the label `suburb`:

```bash
curl --get http://localhost:4400/parse \
  --data-urlencode 'address=john f kennedy international airport'

[{"label":"suburb","value":"john f kennedy international airport"}]
```

However it returns the correct label for a `neighbourhood`:

```bash
curl --get http://localhost:4400/parse \
  --data-urlencode 'address=soho'

[{"label":"suburb","value":"soho"}]
```

And also in the case where the text refers to either an airport or a neighbourhood:

```bash
curl --get http://localhost:4400/parse \
  --data-urlencode 'address=tegel'

[{"label":"suburb","value":"tegel"}]
```

I think the best course of action here is to check for any parse containing a single label of type "suburb" and then discarding it.
While this certainly will discard some valid neighbourhood parses, they should be adequately handled by the fallback parsing.

### DIFF
--- a/controller/libpostal.js
+++ b/controller/libpostal.js
@@ -76,11 +76,18 @@ function setup(libpostalService, should_execute) {
         // push err.message or err onto req.errors
         req.errors.push( _.get(err, 'message', err) );
 
+      } else if (!Array.isArray(response) || !response.length) {
+        return next();
+
       } else if (_.some(_.countBy(response, o => o.label), count => count > 1)) {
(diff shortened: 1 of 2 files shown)


missinglink:
falling back to the `pelias/parser` in these cases seems preferable:

Screenshot 2022-06-21 at 17 00 29
Screenshot 2022-06-21 at 17 00 50

PR #1631, opened 2022-08-29
Allow custom layers by removing the restriction for non-coarse layers

mansoor-sajjad:
:wave: 
We are working the custom sources and layers which are now supported by the Pelias, which is very good and we are excited about that.
We have data from different sources and we have defined the layers which suits our needs and fit nicely with our data.

---
We are testing the custom sources and layers. It works with the `/autocomplete`, but it doesn't works with the `/reverse` endpoint. 
We have tried to search on this topic and found other issues and discussion regarding this issue.
Specially this one https://github.com/pelias/api/issues/1161, which pointed us towards the `reverse.js`
We see that the restriction has been added to support only the non-coarse layers in the `/reverse` endpoint in the following commit https://github.com/pelias/api/commit/10b1d282010053e4ddece83e6351aa0f9a148a3e. We have not been able to figure out why it was necessary. @trescube If you remember anything about that change from 2017 please help us.

---
We have removed that restriction and make the `/reverse` endpoint to support all the layers like other endpoints.
We have tested the API after that change and it works fine.

### DIFF
--- a/query/reverse.js
+++ b/query/reverse.js
@@ -39,8 +39,7 @@ function generateQuery( clean ){
 
   // layers
   if( _.isArray(clean.layers) && !_.isEmpty(clean.layers) ) {
-    // only include non-coarse layers
-    vs.var( 'layers', _.intersection(clean.layers, ['address', 'street', 'venue']));
+    vs.var( 'layers', clean.layers);
   }
 
   // focus point to score by distance
--- a/test/unit/query/reverse.js
+++ b/test/unit/query/reverse.js
@@ -209,27 +209,6 @@ module.exports.tests.layers = (test, common) => {
 
   });
 
-  test('non-empty array clean.layers should only set non-coarse layers in vs', t => {
-    const clean = {
-      layers: all_layers
-    };
-
-    const query = proxyquire('../../../query/reverse', {
-      'pelias-query': {
-        layout: {
-          FilteredBooleanQuery: MockQuery
-        },
-        view: views,
-        Vars: require('pelias-query').Vars
-      },
-      './reverse_defaults': {}
-    })(clean);
-
-    t.deepEquals(query.body.vs.var('layers').toString(), ['address', 'venue', 'street']);
-    t.end();
-
-  });
-
 };
 
 module.exports.tests.focus_point = (test, common) => {
--- a/query/reverse_defaults.js
+++ b/query/reverse_defaults.js
@@ -5,7 +5,6 @@ module.exports = _.merge({}, peliasQuery.defaults, {
 
   'size': 1,
   'track_scores': true,
-  'layers': ['venue', 'address', 'street'],
 
   'centroid:field': 'center_point',
 


mansoor-sajjad:
@missinglink Can you please review this?

mansoor-sajjad:
@missinglink Can you please review this?

missinglink:
Hi @mansoor-sajjad, I believe the original intent was to avoid showing a mix of administrative data and non-administrative data in the results, unfortunately there are no screenshots of the behaviour at that time but I can imagine a worst-case scenario of something like this:

```bash
# Search for 'Torstraße 10, Berlin, Germany'

1. Torstraße 10, Berlin, Germany
2. Berlin, Germany
3. Germany
```

This isn't the default behaviour we'd like to see in Pelias, as results `2.` and `3.` are included in the `properties` of the first.

I haven't tested your PR but I'm guessing that it would revert to the previous behaviour?

missinglink:
Just a thought, but instead of doing `_.intersection(clean.layers, ['address', 'street', 'venue']))` can't we do `_.difference(clean.layers, config.api.targets.layer_aliases.coarse)` to establish which layers are non-coarse.

https://github.com/pelias/config/blob/master/config/defaults.json#L63

test/unit/query/reverse.js
```diff
@@ -209,27 +209,6 @@ module.exports.tests.layers = (test, common) => {
...
-        Vars: require('pelias-query').Vars
-      },
-      './reverse_defaults': {}
-    })(clean);
-
-    t.deepEquals(query.body.vs.var('layers').toString(), ['address', 'venue', 'street']);
-    t.end();
-
-  });
-
```
orangejulius, line 232: Make sure to restore this test and have it check for the new values after the change to the layer calculation :)

orangejulius (changes requested):

mansoor-sajjad:
Hi, 

Sorry for the late response. 
Thanks for the review and excellent solution suggestion. @missinglink @orangejulius 
I have implemented the suggested solution and fixed / restored the unit tests.

query/reverse.js
```diff
@@ -39,7 +40,7 @@ function generateQuery( clean ){
 
   // layers
   if( _.isArray(clean.layers) && !_.isEmpty(clean.layers) ) {
-    vs.var( 'layers', clean.layers);
+    vs.var( 'layers', _.difference(clean.layers, peliasConfig.api.targets.layer_aliases.coarse));
```
missinglink, line 43: Sorry I didn't mention this before, but there is a `get()` method on the config which handles the case where this path doesn't exist for some reason, which would otherwise result in an error.

ie. if `api.targets` didn't exist for some reason.

If I recall correctly, it would look like this, the second param being a default value if the path is not resolvable:
```
peliasConfig.get('api.targets.layer_aliases.coarse', [])
```

ref: https://github.com/pelias/config/blob/master/index.js#L80

mansoor-sajjad: Fixed 👍

missinglink (approved):
Nice and simple, happy to merge this.

@orangejulius you'll need to remove your block before we can proceed.

orangejulius:
Hi @mansoor-sajjad,
This all looks good to me. Can you please check the "allow edits from maintainers" checkbox on this PR so that I can do a little quick cleanup before we merge?

Please see the docs here: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

mansoor-sajjad:
Hei @orangejulius,

"Allow edits by maintainers" is already checked. ✅

Issue #1502, opened 2020-11-09
pelias-api used as lib


stefanocudini:
# Use-cases

useful for creating a wrappers for pelias api

### Attempted Solutions

I want use this module as npm lib like this way:
```javascript
const app = require('pelias-api/app');
```
but noew recive this error, I think that could be solved easily
```
node index.js 
2020-11-09T14:17:54.309Z - warn: [pip] pip service disabled
2020-11-09T14:17:54.310Z - warn: [placeholder] placeholder service disabled
2020-11-09T14:17:54.310Z - warn: [language] language service disabled
2020-11-09T14:17:54.311Z - warn: [interpolation] interpolation service disabled
2020-11-09T14:17:54.311Z - warn: [libpostal] libpostal service disabled
2020-11-09T14:17:54.311Z - warn: [libpostal] libpostal service disabled
fs.js:114
    throw err;
    ^

Error: ENOENT: no such file or directory, open './public/apiDoc.md'
```

stefanocudini:
hi,
this strange behavior been changed recently?

PR #1640, opened 2022-10-05
fix issue #1502

stefanocudini:
this allows you to use pelias-api in a slightly different environment as a library.
I also think it is a bit more elegant solution, considering that including an md file as a result of an api rest endpoint is already quite a whimsical thing :)

I wish you good work

### DIFF
--- a/routes/v1.js
+++ b/routes/v1.js
@@ -1,4 +1,5 @@
 const _ = require('lodash');
+const fs = require('fs');
 const path = require('path');
 const requireAll = require('require-all');
 const express = require('express');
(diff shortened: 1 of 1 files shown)


stefanocudini:
@missinglink @orangejulius could you please accept this small contribution? it is useful for my prod systems and I have made it a generic feature, the code is simple and clean

missinglink:
Hi @stefanocudini, I've fixed the core issue in https://github.com/pelias/api/pull/1642, we never intended this use-case (use as a library) but it should now be possible.

I don't want to encourage users to remove attribution route, it's a legal requirement for the use of the data, and gives credit to the maintainers of the project and contributors.

That being said, the `app` variable returned by `addRoutes()` which you can [edit in your application](https://stackoverflow.com/a/14034355/339280).

PR #1641, opened 2022-10-26
update lodash to newer version

vanessayuenn:
The project current uses `lodash@^4.17.4`, which is vulnerable to Prototype Pollution (see this [advisory link](https://github.com/advisories/GHSA-jf85-cpcp-j695)). The vulnerability has been patched in `4.17.12`.

---

dependency on lodash has been updated to `^4.17.12`

---

I ran the test suite and nothing seems to have broken from this update, so nothing to see here!

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "geolib": "^3.0.0",
     "iso-639-3": "^1.0.0",
     "locale": "^0.1.0",
-    "lodash": "^4.17.4",
+    "lodash": "^4.17.12",
     "markdown": "^0.5.0",
     "morgan": "^1.8.2",
     "pelias-compare": "^0.1.16",


missinglink:
Thanks @vanessayuenn, I doesn't look like we're using `defaultsDeep` but it's always good to keep up-to-date on security patches.

Before I merge this, is there a reason why we wouldn't pin to the latest version [4.17.21](https://www.npmjs.com/package/lodash/v/4.17.21)?

vanessayuenn:
@missinglink hah just an oversight on my part. Thanks for catching that. I've updated to use the latest version instead.

Mohammed101msa:
> #### Here's the reason for this change 🚀
> The project current uses , which is vulnerable to Prototype Pollution (see this [advisory link](https://github.com/advisories/GHSA-jf85-cpcp-j695)). The vulnerability has been patched in .`lodash@^4.17.4``4.17.12`
> 
> #### Here's what actually got changed 👏
> dependency on lodash has been updated to `^4.17.12`
> 
> #### Here's how others can test the changes 👀
> I ran the test suite and nothing seems to have broken from this update, so nothing to see here!

Issue #1644, opened 2022-12-22
fix: test case fails for diffPlaces


wasi-m:
**Describe the bug**

The Test case fails for the diff Places

**Steps to Reproduce**

Clone dev branch, and run the tests

`PELIAS_CONFIG=./test/test-pelias-config.json node --trace-uncaught test/unit/run.js | npx tap-dot`

**Expected behavior**

Should check for `grolmanstrasse` 

**Environment (please complete the following information):**

- OS: Mac
- Docker version: Docker version 20.10.21, build baeda1f
- Docker Compose version: Docker Compose version v2.12.2

**Pastebin/Screenshots**

<img width="385" alt="image" src="https://user-images.githubusercontent.com/47974693/209108401-f263fc0d-07aa-44a9-bca9-af26c2619f3d.png">

michaelkirk:
This test is failing for me too. 

It seems like the underlying remove-accents package is implicated.

The test is depending on behavior from this change: https://github.com/tyxla/remove-accents/issues/12 included first in  release v0.4.2.

But that behavior was reverted in 0.4.4 released in November '22.
See https://github.com/tyxla/remove-accents/pull/31

I don't really know what behavior is better. But if we want to just delegate that decision to `remove-accents`, then as well as updating this test, it might make sense to bump the minimum version of remove-accents to 0.4.4 to keep everything coherent.

missinglink:
Hi, thanks for tracking this down, I alerted the author in the linked issue above.

orangejulius:
I just merged https://github.com/pelias/api/pull/1655 to pin to a version of `remove-accents` that operates how our unit tests expect and we believe to be correct. Sounds like we can undo that pin in the near future though, as `remove-accents` may revert the change that broke our tests. Until then tests are passing again for everyone :)

PR #1645, opened 2022-12-22
fix: diffPlaces diacritics test case 

wasi-m:
Signed-off-by: wasim <wasim.tamboli@olacabs.com>

---

:wave: Hi fixed issue https://github.com/pelias/api/issues/1644

---

Changed the expected result in the test case for diffPlaces 
---

To check, run the test cases
`PELIAS_CONFIG=./test/test-pelias-config.json node --trace-uncaught test/unit/run.js | npx tap-dot`


orangejulius:
Thanks for this PR, we've actually fixed the test by pinning to a version of the `require-accents` NPM module that behaves as we believe is correct and is expected by the unit tests. See #1655 :)

PR #1655, opened 2023-07-02
Pin `remove-accents` to 0.4.2

orangejulius:
A unit test has been failing because this dependency added normalization to the German [Eszet](https://en.wikipedia.org/wiki/%C3%9F) character.

From a issue thread it sounds like we are going to encourage the maintainers to back out that change, but until then pin to a specific known-good version so that our unit tests pass :)

Fixes https://github.com/pelias/api/issues/1644
Connects https://github.com/tyxla/remove-accents/issues/12

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
     "pelias-sorting": "^1.7.0",
     "predicates": "^2.0.0",
     "regenerate": "^1.4.0",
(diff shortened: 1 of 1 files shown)


missinglink:
Good idea, I'll follow up the discussion with the author of that library when I have time.

PR #1675, opened 2024-07-11
improved deduplication between USA ZIP vs ZIP+4 properties

missinglink:
this PR considers two American addresses with matching ZIP & ZIP+4 as equivalent.


missinglink:
Screenshot 2024-07-12 at 12 48 07

missinglink:
There is an opportunity for follow-up work here if we decide to try and favour the `ZIP+4` over the `ZIP`, it's not a priority at this stage. Currently the former duplicate takes preference over the latter.

PR #1679, opened 2024-10-21
Removed unused boost_exact_matches

gmarti:
Hello,

When looking at the code I've found that boost_exact_matches has been disabled but the code is still there.

What about removing it ?


missinglink:
hmm yeah you're right, it seems all references were removed in https://github.com/pelias/api/pull/1287 but these files/lines were missed.

The tests all pass, happy to merge this for house-keeping.

PR #1685, opened 2025-01-23
structured: make string replacements case insensitive

missinglink:
I noticed today that a `/v1/search/structured` query returned different results based on the casing of the unit number:

`/v1/search/structured?address=kinkerstraat+175f`
```
"parsed_text": [{
  "housenumber": "175f",
  "street": "kinkerstraat"
},
```

`/v1/search/structured?address=kinkerstraat+175F`
```
"parsed_text": [{
  "housenumber": "175f",
  "street": "kinkerstraat 175F"
},
```

This PR resolves the issue by converting some string replacement logic to be case-insensitive.

### DIFF
--- a/controller/structured_libpostal.js
+++ b/controller/structured_libpostal.js
@@ -49,14 +49,14 @@ function setup(libpostalService, should_execute) {
           req.clean.parsed_text.housenumber = house_number_field.value;
 
           // remove the first instance of the number and trim whitespace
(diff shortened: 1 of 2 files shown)


missinglink:
I pushed a rebase which uses `_.escapeRegExp()`, because safety.

orangejulius (approved):
Looks good and this fixes a subtle bug.

PR #1688, opened 2025-03-06
Allow custom minSize, maxSize and defaultSize values

### LINKED ISSUE #1649: Increase MAX_SIZE limit for Pelias API
Could the limit be increased to i.e. 100/150 results per request or implement pagination? I'm currently querying Pelias API but the results I want get hidden behind the limit set to 40.

arnesetzer:

---
Closes #1649 

---
Allows users to modify their max/max/defaultSize in the pelias.json. If no values are given it will defaults to the original ones (1,40,10) so this PR should be backwards compatible.

---
Spin up an instance, set some custom values in the pelias.json 
(eg: 
```json
"api": {
    "services": {
      "pip": {
        "url": "http://localhost:4200"
      },
      "libpostal": {
        "url": "http://localhost:4400"
      },
      "placeholder": {
        "url": "http://localhost:4100"
      },
      "interpolation": {
        "url": "http://localhost:4300"
      }
    },
    "minSize": 2,
    "maxSize": 5,
    "defaultSize": 5
  },
  ```
) and play around with the `size` parameter (eg. `size=6` should throw a warning).
There are also a few tests to verify that everything runs as expected :)

### DIFF
--- a/sanitizer/_size.js
+++ b/sanitizer/_size.js
@@ -18,33 +18,33 @@ const DEFAULT_SIZE = default_size_init;
 
 
 // validate inputs, convert types and apply defaults
-function _setup(size_min, size_max, size_def) {
+function _setup( size_min, size_max, size_def ){
 
   // allow caller to inject custom min/max/default values
-  if (!_.isFinite(size_min)) { size_min = MIN_SIZE; }
-  if (!_.isFinite(size_max)) { size_max = MAX_SIZE; }
-  if (!_.isFinite(size_def)) { size_def = DEFAULT_SIZE; }
+  if( !_.isFinite( size_min ) ){ size_min = MIN_SIZE; }
+  if( !_.isFinite( size_max ) ){ size_max = MAX_SIZE; }
+  if( !_.isFinite( size_def ) ){ size_def = DEFAULT_SIZE; }
 
   return {
-    sanitize: function _sanitize(raw, clean) {
+    sanitize: function _sanitize( raw, clean ){
 
       // error & warning messages
       var messages = { errors: [], warnings: [] };
 
       // coercions
-      clean.size = parseInt(raw.size, 10);
+      clean.size = parseInt( raw.size, 10 );
 
       // invalid numeric input
-      if (isNaN(clean.size)) {
+      if( isNaN( clean.size ) ){
         clean.size = size_def;
       }
       // ensure size falls within defined range
-      else if (clean.size > size_max) {
+      else if( clean.size > size_max ){
         // set the max size
         messages.warnings.push('out-of-range integer \'size\', using MAX_SIZE');
         clean.size = size_max;
       }
-      else if (clean.size < size_min) {
+      else if( clean.size < size_min ){
         // set the min size
         messages.warnings.push('out-of-range integer \'size\', using MIN_SIZE');
         clean.size = size_min;
--- a/sanitizer/_size.js
+++ b/sanitizer/_size.js
@@ -1,36 +1,40 @@
 const _ = require('lodash');
-const MIN_SIZE = 1;
-const MAX_SIZE = 40;
-const DEFAULT_SIZE = 10;
+const peliasConfig = require('pelias-config').generate().api;
+//Allow custom min, max and default sizes
+const MIN_SIZE = peliasConfig.minSize || 1;
+const MAX_SIZE = peliasConfig.maxSize || 40;
+//If max_size is smaller than default => default=max
+const DEFAULT_SIZE = peliasConfig.defaultSize || (MAX_SIZE < 10) ? MAX_SIZE : 10;
+
 
 // validate inputs, convert types and apply defaults
-function _setup( size_min, size_max, size_def ){
+function _setup(size_min, size_max, size_def) {
 
   // allow caller to inject custom min/max/default values
-  if( !_.isFinite( size_min ) ){ size_min = MIN_SIZE; }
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: schema.js [hunk 1]
```diff
@@ -26,6 +29,10 @@ module.exports = Joi.object().keys({
       layer: Joi.object(),
       source: Joi.object()
     }),
+    minSize: Joi.number().integer().min(1).max(Joi.ref('maxSize')).default(1),
+    maxSize: Joi.number().integer().min(1).default(40),
+    defaultSize: Joi.number().min(1).max(Joi.ref('maxSize')).min(Joi.ref('minSize')).default(10),
```
orangejulius, line 34: wow, fancy! I didn't know Joi could do this.

sanitizer/_size.js
```diff
@@ -1,36 +1,54 @@
...
-function _setup( size_min, size_max, size_def ){
+function _setup(size_min, size_max, size_def) {
 
   // allow caller to inject custom min/max/default values
-  if( !_.isFinite( size_min ) ){ size_min = MIN_SIZE; }
-  if( !_.isFinite( size_max ) ){ size_max = MAX_SIZE; }
-  if( !_.isFinite( size_def ) ){ size_def = DEFAULT_SIZE; }
+  if (!_.isFinite(size_min)) { size_min = MIN_SIZE; }
+  if (!_.isFinite(size_max)) { size_max = MAX_SIZE; }
+  if (!_.isFinite(size_def)) { size_def = default_size_init; }
```
orangejulius, line 30: this can remain the same and continue using the constant, right?

sanitizer/_size.js
```diff
@@ -1,36 +1,54 @@
 const _ = require('lodash');
-const MIN_SIZE = 1;
-const MAX_SIZE = 40;
-const DEFAULT_SIZE = 10;
+const peliasConfig = require('pelias-config').generate().api;
+//Allow custom min, max and default sizes
+const MIN_SIZE = peliasConfig.minSize || 1;
+const MAX_SIZE = peliasConfig.maxSize || 40;
+let default_size_init;
```
orangejulius, line 6: maybe assign this to 10 to start so the Pelias defaults are more clearly listed in one place?

sanitizer/_size.js
```diff
@@ -1,36 +1,54 @@
...
+if (peliasConfig.DEFAULT_SIZE) {
+  //Cant be smaller/bigger than min/max due to validation in schema.js
+  default_size_init = peliasConfig.DEFAULT_SIZE;
```
orangejulius, line 9: Even though Joi does some bounds checking I would still do some here, there's no telling if we'll keep using Joi in the future, if it's features will change (we are on a very old version). In either case some logic left in the schema could be lost.

I don't think we need anything fancy with throwing errors, just clamp between MIN_SIZE/MAX_SIZE and if there's an invalid condition leave it at the default of 10.

orangejulius (changes requested):
Awesome, thanks for this. We do get a lot of requests here so I'm sure it will help people out.

I left some fairly minor suggestions for changes but overall it looks good.

sanitizer/_size.js
arnesetzer: Yeah, artifact from search and replace. Gonna change it.

arnesetzer:
Thanks for the feedback. Added your requested changes. Hope that met what you had in mind.

missinglink:
I can take this over and merge it, I'll push to your branch

missinglink:
This is kinda hard to test with our current setup since calling `require('sanitizer/_size.js')` gets cached, so, short of wrapping it in a setup function purely for testing it's impossible to test *both* the defaults *and* the config vars 🤷‍♂️ 

That said, I confirmed it works as expected by making these edits (which I didn't commit) and running the tests, the failures are expected.

```diff
diff --git a/test/test-pelias-config.json b/test/test-pelias-config.json
index e730d72f..d0488295 100644
--- a/test/test-pelias-config.json
+++ b/test/test-pelias-config.json
@@ -2,6 +2,9 @@
   "api": {
     "targets": {
       "auto_discover": false
-    }
+    },
+    "minSize": 10,
+    "maxSize": 20,
+    "defaultSize": 15
   }
 }
```

```bash
  Failed Tests:   There were 11 failures

    x default to 1
    x default to 40
    x default to 10
    x should return warning
    x set to correct integer
    x should return warning
    x set to correct integer
    x should return warning
    x set to correct integer
    x should return warning
    x set to correct integer
```

missinglink:
agh crap, after I merged this I noticed a minor error, probably not a big deal?

```diff
diff --git a/schema.js b/schema.js
index 462f6090..7c3be6a7 100644
--- a/schema.js
+++ b/schema.js
@@ -31,7 +31,7 @@ module.exports = Joi.object().keys({
     }),
     minSize: Joi.number().integer().min(1).max(Joi.ref('maxSize')).default(1),
     maxSize: Joi.number().integer().min(1).default(40),
-    defaultSize: Joi.number().min(Joi.ref('minSize')).max(Joi.ref('maxSize')).default(10),
+    defaultSize: Joi.number().integer().min(Joi.ref('minSize')).max(Joi.ref('maxSize')).default(10),
     localization: Joi.object().keys({
       flipNumberAndStreetCountries: Joi.array().items(Joi.string().regex(/^[A-Z]{3}$/))
     }).unknown(false),
```

arnesetzer:
This allows the user to use floats as max-size elements which can lead to a permanent warning, when size the size parameter is near the maxSize limit (eg. maxSize=40.5 and size=40). But the api still works normally.

PR #1691, opened 2025-03-11
correct `config.schema.indexName` reference

missinglink:
The `indexName` configuration is parented by `schema`, not by `api`.

ref: https://github.com/pelias/config/blob/4a52a15a85767407591deaabd78b4e87f09f74f7/config/defaults.json#L73
related: https://github.com/pelias/schema/pull/507#discussion_r1989007224

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,6 @@ The API recognizes the following properties under the top-level `api` key in you
 |`targets.layers_by_source` <br> `targets.source_aliases` <br> `targets.layer_aliases`|no | |custom values for which `sources` and `layers` the API accepts (See more info in the [Custom sources and layers](#custom-sources-and-layers) section below). We recommend using the `targets.auto_discover:true` configuration instead of setting these manually.
 |`customBoosts` | no | `{}` | Allows configuring boosts for specific sources and layers, in order to influence result order. See [Configurable Boosts](#custom-boosts) below for details |
 |`autocomplete.exclude_address_length` | no | 0 | As a performance optimization, this optional filter excludes results from the 'address' layer for any queries where the character length of the 'subject' portion of the parsed_text is equal to or less than this many characters in length. Addresses are usually the bulk of the records in Elasticsearch, and searching across all of them for very short text inputs can be slow, with little benefit. Consider setting this to 1 or 2 if you have several million addresses in Pelias. |
-|`indexName`|*no*|*pelias*|name of the Elasticsearch index to be used when building queries|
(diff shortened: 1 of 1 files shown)


missinglink:
Hmm actually this is wrong too 
https://github.com/pelias/api/blob/master/schema.js#L22

@orangejulius sanity check plz, am I doing the right thing?

missinglink:
`api.indexName` is referenced all over this repo but it's not in the default Pelias config, I'm much confused

missinglink:
Oh I'm blind, it is duplicated in the config:
https://github.com/pelias/config/blob/4a52a15a85767407591deaabd78b4e87f09f74f7/config/defaults.json#L39

Do we need both?

missinglink:
I would advocate removing `api.indexName`.

The reason is that when you [create an index](https://github.com/pelias/schema/blob/master/scripts/create_index.js#L29) you can change `config.schema.indexName` to whatever you like, but then `pelias/api` requires additionally setting  `config.api.indexName` or it will fail to find that index, this seems unnecessarily confusing.

Is there a valid use-case where someone might want `config.schema.indexName` & `config.api.indexName` to differ?

missinglink:
OK, fair enough, I'm going to close this as a `wontfix` since having both is the expected behaviour.

missinglink:
For anyone who find this issue in the future, a blue-green deployment can be achieved more simply using [Elasticsearch Aliases](https://www.elastic.co/guide/en/elasticsearch/reference/current/aliases.html) with zero-downtime.

orangejulius:
Ah, I remember the issue we had at Mapzen now: normally, we loaded builds in to Elasticsearch with an index name of the format `pelias-$date`, and then had `pelias` be an index alias to whichever build we wanted to be live. When we were loading a new build there would be `pelias-$oldDate` and `pelias-$newDate` in various levels of readiness on the same Elasticsearch cluster.

But then at some point something was messed up in our very convoluted and painful infrastructure setup, and we loaded an index into the cluster directly as `pelias`. That means we couldn't use an alias because there was an actual index there. We were stuck without a way to change which index the API used. Hence this config option.

I think having the configurability here in the API is still good, many people might prefer making changes to Pelias config instead of using Elasticsearch index aliases (one could imagine an infrastructure setup with read-only access to the Elasticsearch cluster for safety, but easy ways to change `pelias.json` without downtime). Plus it allows anyone to get out of the situation that we once faced at Mapzen.

Issue #1653, opened 2023-05-26
Layers and sources ignored results in ignored sources only


arnesetzer:
**Describe the bug**
If you start a request and ignore some sources (eg: `sources=-gn`) and some layers (eg. `layers=-macrocounty%2C-empire%2C-continent%2C-country%2C-region%2C-ocean%2C-venue`) the negated layers are ignored every time.

**Steps to Reproduce**
Create a query where some layers and sources are ignored. (eg. https://pelias.github.io/compare/#/v1/search?sources=-gn&layers=-macrocounty%2C-empire%2C-continent%2C-country%2C-region%2C-ocean%2C-venue&text=M%C3%BCnchen+Hbf&debug=1).
As you can see in the results that venues are not ignored. If the query just ignores some layers (eg. https://pelias.github.io/compare/#/v1/search?layers=-macrocounty%2C-empire%2C-continent%2C-country%2C-region%2C-ocean%2C-venue&text=M%C3%BCnchen+Hbf&debug=1) everything looks good.

**Expected behavior**
That the two parameters are combinded an return a result which suits the parameters. Or throws an error/warning that this combination isn't allowed.

**Environment (please complete the following information):**
- OS: Debian/Ubuntu
-  Docker Version: 24.0.1
- Docker compose version: 1.29.2/2.18.1 (with https://github.com/pelias/docker/pull/316)
Since this error occurs on the offical geocode instance I don't think that it is OS dependent.

### References

Didn't found any. Sorry, if I created a duplicate 🙈

PR #1693, opened 2025-03-28
Handle sources and layers param with address filter

orangejulius:
As reported in https://github.com/pelias/api/issues/1653, there is an issue when negative sources are specified in combination with negative layers. The end result is that the negative layers are ignored.

This issue was likely introduced in https://github.com/pelias/api/pull/1604 or https://github.com/pelias/api/pull/1525, the pull requests where we added the concept of negative sources and layers and subsequently improved them.

The change is split into two commits, a pure refactoring for clarity, and then the fix, which is quite short.

Fixes https://github.com/pelias/api/pull/1653

### DIFF
--- a/test/unit/sanitizer/_address_layer_filter.js
+++ b/test/unit/sanitizer/_address_layer_filter.js
@@ -117,6 +117,26 @@ module.exports.tests.sanitize = function (test, common) {
     t.end();
   });
 
+  test('sanitize - exclude addresses when negative layers and sources are specified', (t) => {
+    // select all layers except venue to simulate value of clean.layers from targets sanitizer
+    const clean_layers = real_type_mapping.getCanonicalLayers().filter(layer => layer !== 'venue').sort();
+
+    let clean = { text: 'foo',
+      layers: clean_layers,
+      negative_layers: ['venue'],
+      positive_layers: [],
+      sources: ['openstreetmap', 'openaddresses','whosonfirst'],
+      negative_sources: ['geonames'],
+      positive_sources: []
+    };
+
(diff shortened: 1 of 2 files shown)


test/unit/sanitizer/_address_layer_filter.js
```diff
@@ -117,6 +117,26 @@ module.exports.tests.sanitize = function (test, common) {
...
+      positive_sources: []
+    };
+
+    const expected_layers = clean_layers.filter(layer => layer !== 'address').sort();
+
+    t.deepEqual(real_sanitizer.sanitize(null, clean), STD_MESSAGES);
+    t.deepEqual(clean.layers.sort(), expected_layers, 'layer list is reduced to exclude addresses');
+    t.end();
+  });
+
```
missinglink, line 139: I'm struggling to understand a few things about this unit test:
- The test is named "exclude addresses when negative layers and sources are specified", is this specific to the `address` layer?
- `clean_layers` filters out `venue` and sorts the array, then `expected_layers` does the same thing? Is this maybe to avoid sharing the same pointer as `clean_layers`?
- The selected `sources` are not being asserted.

missinglink (approved):

test/unit/sanitizer/_address_layer_filter.js
orangejulius: That's right, the entire `_address_layer_filter` file is centered around an optimization to filter out the address layer when two conditions are met:
- the input has certain parses, is below a certain number of tokens, etc
- the layers specified by the user are compatible with filtering out addresses. For example `layers=addresses` means we can't, but `layers=-venue` means we can.

PR #1697, opened 2025-06-30
upgrade `pelias/parser` to latest version

missinglink:
this PR upgrades the `pelias/parser` version to latest
https://github.com/pelias/parser/releases

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "pelias-logger": "^1.2.0",
     "pelias-microservice-wrapper": "^1.10.0",
     "pelias-model": "^9.0.0",
-    "pelias-parser": "2.2.0",
+    "pelias-parser": "3.4.0",
     "pelias-query": "^11.4.0",
     "pelias-sorting": "^1.7.0",
     "predicates": "^2.0.0",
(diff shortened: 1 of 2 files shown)


Joxit:
Hi there,
I just tried your PR, I think there are no ideal solution here. We are using `phrase.*` that contains the full address (number + street), since we need to deal with "junk in house number data" and try to match as much as data we can, stripping the alpha is for now the best solution.

Maybe we could also create some [post-processing](https://github.com/pelias/model/blob/master/post/) in pelias model to include alias or sanitised names/phrases mixed to a fallback query ?

I have in mind `1 bis`, the data may have either `1b`, `1 bis` or `1bis`, in the current state, `1 bis` will not match `1bis`.
So the model post-processing may replace all `1bis` to `1 bis` (or add an alias). `1b` is a bit tricky since it can be a unit, but as an alias of `1b` it should be fine ? The final state might be `{"name": {"default": ["1bis Avenue Amélie", "1 bis  Avenue Amélie"]}}`

Then the query we make could include a `should` with a `minimum_should_match` that will query both `1 bis Avenue Amélie` and `1 Avenue Amélie` as a fallback.

Oh I forgot the case when the user's input is `1bis  Avenue Amélie` => change to `1 bis Avenue Amélie` and if it's `1b  Avenue Amélie` => IDK if it's safe to include 3 queries, `1b  Avenue Amélie`, `1 bis Avenue Amélie` and `1 Avenue Amélie`

It is still an idea and should not be put in this PR, this might be a big change.

PR #1720, opened 2026-04-13
Remove JSONP middleware

orangejulius:
JSONP is a relic of the ancient past. If you're using it, you shouldn't be.


missinglink (approved):
😱

PR #1721, opened 2026-08-21
Unify on using pretty diff and Javascript fixtures

orangejulius:
These fixtures have been a considerable source of pain and annoyance when working with query code. This is a first step at some quality of life improvements.

Some fixtures were written as JSON, which seems like the right call (Elasticsearch queries are JSON), but in practice means we can't have comments, etc, and different fixtures have different syntax. Most are Javascript, and after this commit all of them are.

We also use difflet for better visual diffs, which has been available for a while in this repo but never really used.


missinglink (approved):
Nice

